PR #2, opened 2022-03-10
Setup multi-module project structure

sven1103:


sven1103:
> Nice Job! I was able to login and test the application. I also really like that you included links with additional information in the Readme 🚀
> 
> However what confused me was that during application start the domain call returning a andom numbers was performed twice 😕 . Since this is not relevant to the overall structure this should not block the PR.
> 
> Bildschirmfoto 2022-03-11 um 08 29 08

From the log you can see, that the devtools RestartClassLoader has been instantiated. I think that's why it gets executed twice. When you compile the application with the Maven `production` profile, this is not the case.

PR #3, opened 2022-03-11
Set up dependency management

sven1103:
This PR configures the dependency management


pom.xml
```diff
@@ -1,83 +1,110 @@
...
+        <checksumPolicy>fail</checksumPolicy>
+        <enabled>true</enabled>
+        <updatePolicy>always</updatePolicy>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <url>https://qbic-repo.qbic.uni-tuebingen.de/repository/maven-releases</url>
+    </repository>
+  </repositories>
```
jenniferboedker, line 78: Shouldn't those two repos be first in order to reduce compile time?

webapp/pom.xml
```diff
@@ -70,21 +70,13 @@
                 <artifactId>domain</artifactId>
                 <version>1.0-SNAPSHOT</version>
             </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
-                <artifactId>vaadin-bom</artifactId>
-                <version>${vaadin.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
```
jenniferboedker, line 79: why did you move the vaadin dependency? I would expect it on the app layer and not in the webapp since only here there should be UI stuff.

jenniferboedker (commented):
Thanks for reorganizing, I have added some comments.

webapp/pom.xml
sven1103: true, i have refactored it

pom.xml
sven1103: no, exactly the oposite. We dont want to search third party dependencies in the Vaadin or our repo. This will lead to increased latency due to the empty requests to these repos, and we know a-priori that most of the stuff is in Maven central :)

jenniferboedker (approved):

PR #6, opened 2022-04-19
Provide JPA implementation for user access 

sven1103:


KochTobi:
The branch does not build :)

jenniferboedker:
I updated the base branch for this PR to merge all changes into the parent story of jira (DM-306)

KochTobi (approved):
Hi @sven1103 
Very nice work!

domain/src/test/groovy/life/qbic/usermanagement/repository/UserRepositorySpec.groovy
```diff
@@ -24,6 +24,34 @@ class UserRepositorySpec extends Specification {
...
+        UserDataStorage storage = Mock(UserDataStorage.class)
+        def user = createDummyUser()
+        storage.findUsersByEmail(_ as String) >> [user]
+        UserRepository repository = new UserRepository(storage)
+
+        when:
+        def matchingUser = repository.findByEmail("my.example@example.com")
+
+        then:
+        matchingUser.get().getId().equalsIgnoreCase(user.getId())
```
KochTobi, line 38: I think in this case an additional check makes understanding test failure easier
```suggestion
        then:
        matchingUser.isPresent()
        matchingUser.get().getId().equalsIgnoreCase(user.getId())
```
Otherwise, you would have to look into a NoSuchElementException for empty Optionals.

domain/src/main/java/life/qbic/usermanagement/repository/UserRepository.java
```diff
@@ -54,10 +55,7 @@ public Optional<User> findByEmail(String email) throws RuntimeException {
     if (matchingUsers.size() > 1) {
       throw new RuntimeException("More than one user entry with the same email exists!");
     }
-    if (matchingUsers.isEmpty()) {
-      return Optional.empty();
-    }
-    return Optional.of(matchingUsers.get(0));
+    return Optional.empty();
```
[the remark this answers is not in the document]
sven1103: nice spot, fixed it

KochTobi (approved):
Works now.

domain/src/test/groovy/life/qbic/usermanagement/repository/UserRepositorySpec.groovy
sven1103: True, but the test checks explicitly for success :) So if the implementation returns an Optional.empty, this test would fail and the implementation would need to be refactored

KochTobi: Yes, I tried it and the matching.isPresent not met was easier to follow up on then the NoSuchElementException from the second check. Both do their job. So this is optional 😛

PR #9, opened 2022-05-04
Provide PBKDF2 password encryption and fix route handling

sven1103:

PR #7, opened 2022-04-20
Dm 317 implement interface register login

jenniferboedker:

PR #10, opened 2022-05-09
Create new user account

jenniferboedker:
- [x] Provide JPA implentation #6
- [x] Implement GUI #7 and #9
- [x] Implement encryption #9

**Acceptance Criteria:** 

- User can login with correct credentials
- Wrong credentials prevent login
- Wrong credentials notify user of login failure
- A logged in person can see sites that an unauthorized user can not
- After successful login the user is redirected to a welcome page


webapp/src/main/java/life/qbic/views/ErrorMessage.java
```diff
@@ -0,0 +1,68 @@
...
+public class ErrorMessage extends Composite<VerticalLayout> {
+
+  private final String descriptionText;
+  private final String titleText;
+
+  private Span titleSpan;
+  public Span titleTextSpan;
+
+  private Div descriptionDiv;
+  public Span descriptionTextSpan;
```
KochTobi, line 25: I am a bit confused about which fields are public and which are not and why there is a difference. Can someone explain? :)

webapp/src/main/resources/application.properties
```diff
@@ -7,3 +7,10 @@ vaadin.launch-browser=true
 # To improve the performance during development.
 # For more information https://vaadin.com/docs/flow/spring/tutorial-spring-configuration.html#special-configuration-parameters
 vaadin.whitelisted-packages = com.vaadin,org.vaadin,dev.hilla,life.qbic
+
+# Database setup configuration
```
KochTobi, line 11: I think those properties should be added to the README

KochTobi (approved):
LGTM. It would be nice to update the README. Other then that, I tried it and it worked.

webapp/src/main/java/life/qbic/views/ErrorMessage.java
jenniferboedker: The fields that contain the text are public so they could be accessed e.g to get the content. 
The idea was to hide the other elements since they are only containers for the text elements

KochTobi (approved):

webapp/src/main/java/life/qbic/views/ErrorMessage.java
KochTobi: I think then you should introduce setters. this mixture looks strange

PR #12, opened 2022-05-12
Change encryption method

jenniferboedker:
Change encryption method from PBKDF2 to PBE because of reported security vulnerability issues (see https://github.com/qbicsoftware/data-manager-app/security/code-scanning/1)

I also increased the iterations as well

### DIFF
--- a/domain/src/main/java/life/qbic/usermanagement/policies/PasswordEncryptionPolicy.java
+++ b/domain/src/main/java/life/qbic/usermanagement/policies/PasswordEncryptionPolicy.java
@@ -1,12 +1,12 @@
 package life.qbic.usermanagement.policies;
 
+import javax.crypto.SecretKey;
+import javax.crypto.SecretKeyFactory;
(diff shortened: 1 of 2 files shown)


domain/src/main/java/life/qbic/usermanagement/policies/PasswordEncryptionPolicy.java
```diff
@@ -25,7 +25,7 @@ public class PasswordEncryptionPolicy {
   private static final int HASH_INDEX =
       2; // the index of the hash content in the encoded password String
   private static final int ITERATIONS =
-      4242; // the iteration count used for the encryption algorithm
+      10000; // the iteration count used for the encryption algorithm
```
sven1103, line 28: Increases readability:
```suggestion
      10_000; // the iteration count used for the encryption algorithm
```

sven1103 (dismissed):
One nit suggestion, else I am happy with the improvement.

sven1103 (approved):

PR #15, opened 2022-05-17
Implement Event store

KochTobi:


domain/src/main/java/life/qbic/events/DomainEventSubscriber.java
```diff
@@ -0,0 +1,15 @@
...
+
+import life.qbic.events.DomainEvent;
+
+/**
+ * <b>short description</b>
+ *
+ * <p>detailed description</p>
+ *
+ * @since <version tag>
+ */
```
sven1103, line 11: Description and purpose is missing

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
```diff
@@ -0,0 +1,63 @@
+package life.qbic.events;
+
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import life.qbic.usermanagement.registration.UserRegistered;
+
+public class DomainEventSerializer {
```
sven1103, line 9: Description and purpose missing

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public <T extends DomainEvent> String serialize(T event) {
```
sven1103, line 11: Docs missing

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+    if (event instanceof UserRegistered) {
+      return UserRegisteredSerializer.serialize((UserRegistered) event);
+    }
+    throw
+        new UnrecognizedEventTypeException(
+        "Cannot serialize events of type " + event.getClass().getName() + ". Unknown event type.");
+  }
+
+  public DomainEvent deserialize(String serializedEvent, String desiredType) {
```
sven1103, line 20: Docs missing

domain/src/main/java/life/qbic/usermanagement/registration/UserRegistered.java
```diff
@@ -23,6 +19,10 @@ public UserRegistered(String userId) {
     this.occurredOn = Instant.now();
   }
 
+  public void setOccurredOn(Instant occurredOn) {
+    this.occurredOn = occurredOn;
+  }
+
```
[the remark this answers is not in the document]
KochTobi: We need to set it during deserialization. The event timestamp should not change. We could do it with a static creation method as well

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+      }
+      if (userId == null) {
+        throw new RuntimeException("Cannot deserialize to type " + UserRegistered.class.getName());
+      } else if (occurredOn == null) {
+        throw new RuntimeException("Cannot deserialize to type " + UserRegistered.class.getName());
+      }
+      UserRegistered deserializedEvent = new UserRegistered(userId);
+      deserializedEvent.setOccurredOn(occurredOn);
+      return deserializedEvent;
+    }
```
sven1103, line 60: Why the pain and implement serialisation and de-serialisation? We can use Java's build in Object Serialisation and Deserialisation features. 

For example create an StoredEvent value object, use Javas ObjectOutputStream (https://docs.oracle.com/javase/7/docs/api/java/io/ObjectOutputStream.html) and write it as Blob in the DB. Same works in the reverse order, with ObjectInputStream.

Another issue is, that you did not save the event type. This is important information, since the type determines all later routing, within and for remote contexts. The Domain Event class name as String is sufficient, following the example from Vaughn Vernon.

We also need an event id for de-duplication processing.

webapp/src/main/java/life/qbic/AppConfig.java
```diff
@@ -42,4 +44,13 @@ public UserRepository userRepository(UserDataStorage userDataStorage) {
     return UserRepository.getInstance(userDataStorage);
   }
 
+  /**
+   * Creates the event store instance
+   * @return 1.0.0
+   */
+  @Bean
+  public EventStore eventStore() {
+    return new EventStore(new TemporaryEventRepository());
```
sven1103, line 53: I prefere a Singleton vs constructor, to also guarantee that we only have on instance of an event store in the application instance.

```suggestion
    return EventStore.instance(new TemporaryEventRepository());
```

webapp/src/main/java/life/qbic/events/EventStore.java
```diff
@@ -0,0 +1,37 @@
...
+ * Implementation of a basic event store. It handles events and provides accessor methods to retain
+ * the events later.
+ */
+public class EventStore {
+
+  private final EventRepository eventRepository;
+
+  public EventStore(EventRepository eventRepository) {
+    this.eventRepository = eventRepository;
+  }
```
sven1103, line 16: Effective Java: Favour static factory methods over constructors for object creation.  This also works when injecting dependencies. 

```suggestion
  public EventStore(EventRepository eventRepository) {
    this.eventRepository = eventRepository;
  }
```

sven1103 (commented):
Thank you @KochTobi for your implementation! I left some comments open for discussion :)

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+        "Cannot deserialize events of type " + desiredType + ". Unknown event type.");
+  }
+
+  private static class UserRegisteredSerializer {
+
+    static String serialize(UserRegistered event) {
+      HashMap<String, String> propertiesMap = new HashMap<>();
+      propertiesMap.put("userId", event.getUserId());
+      propertiesMap.put("occurredOn", event.occurredOn().toString());
+      return propertiesMap.toString();
```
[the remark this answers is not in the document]
KochTobi: It is not clear to me what you mean here?

domain/src/main/java/life/qbic/usermanagement/registration/UserRegistered.java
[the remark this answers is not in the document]
sven1103: ah, I see. yeah, then I think we move the responsibility to the client to set the Instant, not as intrinsic behaviour.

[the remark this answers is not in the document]
sven1103: But i think once created, the event should be immutable

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
sven1103: I suggest to introduce an object serialisation interface, and go with a default implementation (like `StringObjectSerialization`) and later we can introduce JSON, or Javas native Object Serialisation

sven1103: and yes, we need the id (incremented by one) so we preserve event order information. We get that for free from the data base.

KochTobi: the event order information is perserved in the stored event :) it does however not have anything to do with the event store

sven1103: how is that?

webapp/src/main/java/life/qbic/events/EventStore.java
KochTobi: Hi @sven1103, you suggested earlier to make the EventStore a singleton. This would mean introducing a `getInstance()` method. Do you know how one can implement a singleton with different repositories or is there only one repository allowed? I implemented a static `getInstance()`. Is this what you had in mind?

webapp/src/main/java/life/qbic/events/DomainEventSerializer.java
KochTobi: The stored event is assigned a `long id` from the event store. This id does not propagate out of the context of the event store and is used by the event store to handle event order information. Furhter the event order information can be read by comparing occurredOn timestamps.

webapp/src/main/java/life/qbic/events/EventStore.java
sven1103: Hi @KochTobi i think a singleton is fine for now, no need for multiple EventStores at the moment. when the model requires it, we change it :)

sven1103 (approved):
Nice work @KochTobi ! I approve and merge it, so we can continue with the integration of events. we can still refactor smaller things later, if required.

PR #16, opened 2022-05-19
Event serialization

KochTobi:


domain/src/main/java/life/qbic/domain/events/SerializableDomainEvent.java
```diff
@@ -0,0 +1,12 @@
+package life.qbic.domain.events;
+
+import java.io.Serializable;
+
+/**
+ * A domain event that can be serialized.
+ *
+ * @since 1.0.0
+ */
+public abstract class SerializableDomainEvent implements DomainEvent, Serializable {
```
sven1103, line 10: I argue that here we sneak in technical vocab into the domain language.

I would favour an abstract domain event class then, that implements the serializable interface.

sven1103 (changes requested):
One small change request, regarding domain language

sven1103 (approved):
Nice!

PR #18, opened 2022-05-23
Send an email given a recipient and fullName

KochTobi:


webapp/src/main/java/life/qbic/usermanagement/registration/RegistrationMessageFactory.java
```diff
@@ -0,0 +1,20 @@
+package life.qbic.usermanagement.registration;
+
+/**
+ * Creates registration messages.
+ */
+public class RegistrationMessageFactory {
+
+  public static String registrationMessage(String fullName) {
```
sven1103, line 8: Since we already know, that we will provide a URL to confirm the email address, we can add it as a method parameter.

Can be a String, must not be a URL, since it is formatting only.

KochTobi: I would extend it once we have it in place. Atm, we do not have an URL so I would opt for a functioning version where a simple email is sent without any action required from the user.

webapp/src/main/java/life/qbic/usermanagement/registration/RegistrationEmailSender.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+/**
+ * Sends emails informing a user that she was registered in the system
+ *
+ * @since 1.0.0
+ */
+@Service
+public class RegistrationEmailSender {
```
sven1103, line 25: Instead of having a solution that cannot be re-used, I favour an EmailSubmissionService, that provides some public methods, without knowing anything about the email purpose. The service needs a recipient email address and a pre-formatted content. 

We can then re-use the service, for example when we confirm the registration and activate the user account. 

This implementation cannot be re-used for other emails unfortunately.

I suggest to wrap the the registration email logic in a EmailComposer factory, that creates Email objects. An Email has a recipient and a message content + subject. 

You can for example then have a call like:

`Email registrationMail = EmailComposer.createRegistrationMail(String recipientEmail, String subject, String content)`

And pass it to the EmailSubmissionService.

sven1103 (changes requested):
Dear @KochTobi thank you for your implementation.

I have a suggestion for the architecture and gave some impulses.

webapp/src/main/java/life/qbic/usermanagement/registration/RegistrationEmailSender.java
KochTobi: I think that makes the solution way too complex. Generalisation makes sense when we need the functionality in other spaces. Currently, we only need it in this function.

sven1103: But why is it complex? It fits into the current application architecture and separates the concerns. Your suggestion puts formatting and submission into one class. 

Also, it does not fit in the Service Layout we have.

sven1103 (changes requested):
I answered the comment, still would like to see a solution that fits into the overall architecture.

domain/src/main/java/life/qbic/apps/datamanager/notifications/MessageParameters.java
```diff
@@ -16,18 +16,18 @@ public final class MessageParameters {
 
   public final String messageId;
 
-  public final String occuredOn;
+  public final String occurredOn;
 
-  private MessageParameters(String messageType, String messageId, String occuredOn) {
+  private MessageParameters(String messageType, String messageId, String occurredOn) {
```
sven1103, line 21: well spotted!

webapp/src/main/java/life/qbic/email/Email.java
```diff
@@ -0,0 +1,23 @@
...
+import life.qbic.usermanagement.registration.Recipient;
+
+/**
+ * <b>short description</b>
+ *
+ * <p>detailed description</p>
+ *
+ * @since <version tag>
+ */
+public interface Email {
```
sven1103, line 12: why an interface? I dont see why we have several implementations in the future. 

I get why you did it, because of the formatting of the content. However I dont see it as the responsibility to format the content. An email just has content that should fit the MIME type. We dont need to model it as complex as the real email world is, but formatting is not an email concern. 

An email is a data carrier, but it needs some logic that prepares the content. I would outsource this logic into an own object, and make the Email a simple data structure.

webapp/src/main/java/life/qbic/email/EmailService.java
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * <b>short description</b>
+ *
+ * <p>detailed description</p>
+ *
+ * @since <version tag>
+ */
+@Service
+public interface EmailService {
```
sven1103, line 13: Not sure if we need an interface, but I will not block it atm.

sven1103 (changes requested):
Very nice, that is quite an improvement in readability and separation of concerns.

I have on final remark to the email implementation and formatting respronsibility. 

Then I give it a go!

domain/src/main/java/life/qbic/apps/datamanager/notifications/MessageParameters.java
KochTobi: The yellow underline from IntelliJ helped a lot here :)

PR #20, opened 2022-05-24
As a User i want to be notified by email after registration

KochTobi:


Steffengreiner (approved):
Thanks for updating the Readme! Now it's perfect 👍

PR #24, opened 2022-05-27
Show errors on why the registration of a user fails

Steffengreiner:
Addresses DM-361 by adding a full name policy to check for user input in the full name field. 
Additionally introduces distinct error messaging and component handling dependent on the violated domain policy.


domain/src/main/java/life/qbic/domain/usermanagement/User.java
```diff
@@ -55,7 +55,8 @@ protected User() {}
    */
   public static User create(String fullName, String email) throws UserException {
     String uuid = String.valueOf(UUID.randomUUID());
-    var user = new User(fullName);
+    var user = new User();
+    user.setFullName(fullName);
```
jenniferboedker, line 59: Why did you change that?

domain/src/main/java/life/qbic/domain/usermanagement/policies/FullNamePolicy.java
```diff
@@ -0,0 +1,38 @@
...
+ * <b>Full Name Policy</b>
+ *
+ * <p>Validates if a String is concurrent with the full name requirements
+ *
+ * @since 1.0.0
+ */
+public class FullNamePolicy {
+
+  private static final int MIN_LENGTH = 1;
+
```
jenniferboedker, line 13: Why did you add that? AFAIK we dont have a policy there other than that it needs to be provided

webapp/src/main/java/life/qbic/views/register/UserRegistrationHandler.java
```diff
@@ -57,30 +60,16 @@ private void addListener() {
...
-    userRegistrationLayout.passwordTooShortMessage.setVisible(false);
     userRegistrationLayout.alreadyUsedEmailMessage.setVisible(false);
+    userRegistrationLayout.errorMessage.setVisible(false);
+    userRegistrationLayout.passwordTooShortMessage.setVisible(false);
+  }
+
+  private void resetComponentErrors() {
+    userRegistrationLayout.fullName.setErrorMessage("A username must be at least 1 character");
+    userRegistrationLayout.email.setErrorMessage("Please provide a valid email");
+    userRegistrationLayout.password.setErrorMessage("Password too short");
```
jenniferboedker, line 72: I am confused, this reads as if you set the error rather than reset

webapp/src/main/java/life/qbic/views/register/UserRegistrationHandler.java
```diff
@@ -90,9 +79,30 @@ public void onSuccess() {
...
+    }
+    if (reason.equals("User with email address already exists.")) {
+      userRegistrationLayout.alreadyUsedEmailMessage.setVisible(true);
+      userRegistrationLayout.email.setInvalid(true);
+      userRegistrationLayout.email.setErrorMessage(reason);
+    }
+    if (reason.equals("Unexpected error occurred.")) {
+      userRegistrationLayout.errorMessage.setVisible(true);
+    }
+    if (reason.equals("Password shorter than 8 characters.")) {
```
jenniferboedker, line 102: Screams for a switch statement

jenniferboedker (changes requested):
Thanks for your contribution! I am a little surprised about the new policy that you have added. Maybe you can elaborate on how you did that. I am not sure if that is a policy in our case since we require the value only to be present. But I am happy to discuss that with you :) 
Other than that: did something change with the application properties? I cannot run the application on my laptop to test if it works as desired 😿

webapp/src/main/java/life/qbic/views/register/UserRegistrationHandler.java
Steffengreiner: Good Point i renamed the method to make it's purpose more clear 👍

domain/src/main/java/life/qbic/domain/usermanagement/policies/EmailFormatPolicy.java
```diff
@@ -36,7 +38,7 @@ public static EmailFormatPolicy create() {
    */
   public PolicyCheckReport validate(String email) {
     if (!honoursRFCSpec(email)) {
-      return new PolicyCheckReport(PolicyStatus.FAILED, "Invalid email address format.");
+      return new PolicyCheckReport(PolicyStatus.FAILED, InvalidEmailMessage);
```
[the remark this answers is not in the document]
Steffengreiner: Originally i wanted to have a link between the invalid email messages the policy could trigger and the ones recognized by the switch statement in the handler to avoid duplication here.
 But I'm unsure how to do this without violating the clean architecture principles.

domain/src/main/java/life/qbic/domain/usermanagement/policies/FullNamePolicy.java
Steffengreiner: I think a policy is necessary here since the Domain logic should decide on the validity of the input. 
Also this allows us to test our requirements easily and adapt the policy if necessary down the line. 
We could e.g. think about not allowing nonsensical names down the line as suggested e.g. here: https://support.google.com/a/answer/9193374?hl=en

domain/src/main/java/life/qbic/domain/usermanagement/User.java
jenniferboedker: I would prefer to keep it as it was since it is no real policy to check for the presence of a value.

domain/src/main/java/life/qbic/domain/usermanagement/policies/EmailFormatPolicy.java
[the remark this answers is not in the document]
jenniferboedker: ok but then you can change it back, right?

domain/src/main/java/life/qbic/domain/usermanagement/policies/FullNamePolicy.java
jenniferboedker: As I said above I dont see it as a policy. Its a check for presence, the other things are requirements you just came up with. Its nothing that we have for now. Therfore I would kindly ask you to remove the policy and check the value in the user entity (I think this was done like this before, right?). You can also do it in the use case but a policy is overkill for this

domain/src/main/java/life/qbic/domain/usermanagement/policies/PasswordPolicy.java
```diff
@@ -32,7 +34,7 @@ public static PasswordPolicy create() {
    */
   public PolicyCheckReport validate(char[] rawPassword) {
     if (rawPassword.length < MIN_LENGTH) {
-      return new PolicyCheckReport(PolicyStatus.FAILED, "Password shorter than 8 characters.");
+      return new PolicyCheckReport(PolicyStatus.FAILED, invalidPasswordMessage);
```
jenniferboedker, line 37: same here as above. no need for a variable

webapp/src/main/java/life/qbic/views/register/UserRegistrationHandler.java
```diff
@@ -90,9 +79,30 @@ public void onSuccess() {
...
+        userRegistrationLayout.alreadyUsedEmailMessage.setVisible(true);
+        userRegistrationLayout.email.setInvalid(true);
+        userRegistrationLayout.email.setErrorMessage(reason);
+      }
+      case "Unexpected error occurred." -> userRegistrationLayout.errorMessage.setVisible(true);
+      case "Password shorter than 8 characters." -> {
+        userRegistrationLayout.passwordTooShortMessage.setVisible(true);
+        userRegistrationLayout.password.setInvalid(true);
+        userRegistrationLayout.password.setErrorMessage(reason);
+      }
```
jenniferboedker, line 105: default case is missing. we could show an unexpected error or so :)

KochTobi: Short remark, feel free to ignore:
Here an enum of possible reasons could make sense, you would not need a default then.

Steffengreiner: In my dreamworld i'd define the reason set in the domain logic in an enum which can also be used in the handler for checking. But this would violate clean architecture 🤷‍♂️

domain/src/main/java/life/qbic/domain/usermanagement/policies/FullNamePolicy.java
Steffengreiner: I don't see the problem with having an easily extendable policy based solution that checks the user input. 
IMO the policy should contain ALL requirements for the associated user field.
Before the introduction of the policy we did not check the user input at all in the user entity.

domain/src/main/java/life/qbic/domain/usermanagement/policies/PasswordPolicy.java
Steffengreiner: alright 👍

domain/src/main/java/life/qbic/domain/usermanagement/policies/EmailFormatPolicy.java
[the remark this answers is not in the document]
Steffengreiner: Sure 👍

domain/src/main/java/life/qbic/domain/usermanagement/UserDomainService.java
```diff
@@ -34,15 +36,51 @@ public UserDomainService(UserRepository userRepository) {
    */
   public void createUser(String fullName, String email, char[] rawPassword) throws UserException {
     // First check, if a user with the provided email already exists
-    if (userRepository.findByEmail(email).isPresent()) {
-      throw new UserException("User with email address already exists.");
-    }
+    validateUserCredentials(fullName, email, rawPassword);
     var domainEventPublisher = DomainEventPublisher.instance();
```
KochTobi, line 40: What you do here is a different behaviour then to lookup whether a user already exists. You could replace that with `isEmailInDatabase` but I do not think that it adds readability here.

domain/src/main/java/life/qbic/domain/usermanagement/UserDomainService.java
```diff
@@ -34,15 +36,51 @@ public UserDomainService(UserRepository userRepository) {
...
+    } catch (UserException e) {
+      userExceptions.add(e);
+    }
+    try {
+      user.setPassword(rawPassword);
+    } catch (UserException e) {
+      userExceptions.add(e);
+    }
+    throwCorrectUserException(userExceptions);
+  }
```
KochTobi, line 74: This does not belong here. The user itself has the responsibility to check it's parameters upon creation. That is why we have a static method in the user for creating users. We do not intent the constructor of a User to be used outside of the User class.

domain/src/main/java/life/qbic/domain/usermanagement/UserDomainService.java
```diff
@@ -34,15 +36,51 @@ public UserDomainService(UserRepository userRepository) {
...
+  private void throwCorrectUserException(List<UserException> userExceptions) {
+    if (userExceptions.size() == 1) {
+      throw userExceptions.get(0);
+    }
+    if (userExceptions.size() > 1) {
+      UserException groupedUserException = new UserException("Multiple Invalid Credentials");
+      userExceptions.forEach(groupedUserException::addSuppressed);
+      throw groupedUserException;
+    }
+  }
```
KochTobi, line 85: From the method name I would read the following information -> we throw an exception if the user is correct?
I think this method needs another name.

KochTobi (commented):
Some comments for reference

webapp/src/main/java/life/qbic/views/register/UserRegistrationHandler.java
```diff
@@ -39,48 +39,53 @@ public void handle(UserRegistrationLayout registrationLayout) {
   }
 
   private void initFields() {
+    userRegistrationLayout.fullName.setPattern(".{1,}");
```
KochTobi, line 42: I do not think this validation is correct. We want a non-empty full name. That requires at least one non-whitespace character.
```suggestion
    userRegistrationLayout.fullName.setPattern("\\S.*");
```
With the requirement: Your name must start with a non-whitespace character.
If we want to allow whitespace in front of the name (and strip it later .. although I do not know why we would do that) the expression would be
```suggestion
    userRegistrationLayout.fullName.setPattern("\\s*\\S.*");
```

KochTobi (changes requested):

domain/pom.xml
```diff
@@ -96,8 +96,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>16</source>
-                    <target>16</target>
+                    <source>17</source>
+                    <target>17</target>
+                  <compilerArgs>--enable-preview</compilerArgs>
```
sven1103, line 101: Preview features in production code? For what feature do we require Java's preview?

Steffengreiner: I think this was added by me to try out the new [switch](https://blog.adamgamboa.dev/switch-expression-on-java-17/ ) feature of java 17

domain/pom.xml
```diff
@@ -104,8 +104,9 @@
...
```
KochTobi, line 109:
```suggestion
```

domain/src/main/java/life/qbic/domain/user/UserDomainService.java
```diff
@@ -25,22 +28,24 @@ public UserDomainService(UserRepository userRepository) {
...
-   * <p>Note: this will create a new domain event of type {@link UserRegistered}
+   * <p>Note: this will create a new domain event of type {@link UserRegistered}.
+   * This method is idempotent, if a user already exists with the given emailAddress, no new user
+   * will be created.
    *
    * @param fullName    the full name of the user
-   * @param email       a valid email address
-   * @param rawPassword the raw password desired by the user
+   * @param emailAddress       a valid email address
+   * @param password the encrypted password desired by the user
```
KochTobi, line 37:
```suggestion
   * @param password the password desired by the user
```

KochTobi (dismissed):
Looks good. However, I noticed that the error shown in case some inputs are not provided correctly does not fit the use case. The invalid credentials error is specific to the login. In the registration case, I do not provide any credentials but only some (personal) information. I would argue it suffices to show the errors on the fields and generalise the text to something like `Please check your input and try again`.
image

PR #25, opened 2022-05-30
Dm 326/dm 364/implement service creating confirmation link

KochTobi:


webapp/src/main/java/life/qbic/usermanagement/registration/EmailConfirmationLinkSupplier.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+@Service
+public class EmailConfirmationLinkSupplier {
+
+  @Value("${server.address}")
+  private String host;
+  @Value("${server.port}")
+  private String port;
+  @Value("${email-confirmation-endpoint}")
+  private String emailConfirmationRoute;
```
jenniferboedker, line 19: Why the terminology switch between route and endpoint? You also have that in the properties file where you use both terms as well

webapp/src/main/java/life/qbic/Application.java
```diff
@@ -67,6 +68,8 @@ private static MessageSubscriber whenUserRegisteredSendEmail(
       }
       try {
         UserRegistered userRegistered = deserialize(message);
+        System.out.println(appContext.getBean(EmailConfirmationLinkSupplier.class)
+            .emailConfirmationUrl(userRegistered.userId()));
```
jenniferboedker, line 72: You could add a test to see if it works rather than adding a command-line printout. What do you think in general about a test to see if the link contains all required parts and/or if no user information is shown in the link

jenniferboedker (commented):
Left some comments to your code. Looks very good :)

webapp/src/main/java/life/qbic/Application.java
KochTobi: I do not know what we want to test here. Testing whether the link has the userid at the end is very close to the implementation. So we would test our current implementation rather than the behaviour (which I do not know how to summarise with a test)

I will remove this printout once anything happens to the link (e.g. it is passed into an email)

webapp/src/main/java/life/qbic/usermanagement/registration/EmailConfirmationLinkSupplier.java
KochTobi: Good point I will switch to endpoint everywhere :+1:

webapp/src/main/java/life/qbic/Application.java
KochTobi: Added some test now to confirm the fixed part of the URL

webapp/src/main/java/life/qbic/usermanagement/registration/EmailConfirmationLinkSupplier.java
```diff
@@ -11,15 +11,24 @@
...
-  @Value("${email-confirmation-endpoint}")
-  private String emailConfirmationRoute;
+
+  private final String host;
+
+  private final String port;
+
+  private final String emailConfirmationEndpoint;
+
+  public EmailConfirmationLinkSupplier(@Value("${server.address}") String host,
```
jenniferboedker, line 21: readability was better before but this is NIT

jenniferboedker (approved):
Thanks for addressing my feedback, good to go now 🚀

webapp/src/main/java/life/qbic/usermanagement/registration/EmailConfirmationLinkSupplier.java
KochTobi: Yes, but it is better to inject dependencies using the constructor than using reflection.

PR #29, opened 2022-05-30
Dm 326/dm 365/populate send email with confirmation link

wow-such-code:
Makes small changes to the endpoint and parameter handling. Uses generated confirmation link in the email sent to the user that just registered.


KochTobi (changes requested):
Code looks good however:
Does not run atm.
```
2022-05-30 17:20:31.086  WARN 84574 --- [0.1-8080-exec-5] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 1364, SQLState: HY000
2022-05-30 17:20:31.086 ERROR 84574 --- [0.1-8080-exec-5] o.h.engine.jdbc.spi.SqlExceptionHelper   : Field 'active' doesn't have a default value
2022-05-30 17:20:31.087  INFO 84574 --- [0.1-8080-exec-5] o.h.e.j.b.internal.AbstractBatchImpl     : HHH000010: On release of batch it still contained JDBC statements
Unexpected error occurred.
```

webapp/src/main/resources/application.properties
```diff
@@ -20,4 +20,5 @@ spring.mail.default-encoding=UTF-8
 spring.mail.port=${MAIL_PORT:587}
 # route for email confirmation consumption
 server.address=${HOST:127.0.0.1}
```
KochTobi, line 22: I think we need to change this as well and use a custom property here

PR #32, opened 2022-06-01
Use Confirmation Text in EmailConfirmationMessage

Steffengreiner:
Replaces the nonsensical text with the confirmation message.


sven1103 (approved):
👍🏽

PR #33, opened 2022-06-03
Feature/dm 420 disable login for non active users

KochTobi:
Introduces the LoginUser use case.


wow-such-code (approved):
makes sense to me, looks very clean, nice work.

Steffengreiner (approved):
LGTM 🚀

PR #36, opened 2022-06-07
make login possible again

KochTobi:


Steffengreiner (approved):
LGTM. However this PR will necessiate a seperate PR handling the Handler logic right?

webapp/src/main/java/life/qbic/views/login/LoginLayout.java
```diff
@@ -33,163 +25,81 @@
...
     registerSpan = new Span(new Text("Need an account? "), routerLink);
   }
 
-  private void setRequiredIndicatorVisible(HasValueAndElement<?, ?>... components) {
-    Stream.of(components).forEach(comp -> comp.setRequiredIndicatorVisible(true));
-  }
   @Override
-  public void setParameter(BeforeEvent beforeEvent, @OptionalParameter String s) {
-    loginHandlerInterface.handle(beforeEvent);
+  public void beforeEnter(BeforeEnterEvent beforeEnterEvent) {
```
sven1103, line 102: Why the change of the interface? This part of the code was working fine already, so I would like to know the reason for the change.

sven1103: https://vaadin.com/docs/latest/flow/routing/url-parameters

sven1103: I would appreciate, if functional code done by others is not just overridden but have a discussion first. This increases a) the learning experience and b) shows appreciation.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -37,25 +41,34 @@ public void handle(LoginLayout loginView) {
...
   private void initFields() {
-    registeredLoginView.password.setHelperText("A password must be at least 8 characters");
-    registeredLoginView.password.setPattern(".{8,}");
-    registeredLoginView.password.setErrorMessage("Invalid Password");
-    registeredLoginView.email.setErrorMessage("Invalid Email");
+    resetErrorMessages();
+  }
+
+  private void resetErrorMessages() {
+    showDefaultError();
```
Steffengreiner, line 52: Why do you show the error here to just hide it immediatly afterwards? 
```suggestion
```

Steffengreiner (approved):
Heya @KochTobi The revert of the loginform looks good to me and i can confirm that it works as intended during testing. 
I'd suggest avoiding a default error method and set the specific error as necessary. 
Since the error messages will be adapted anyway this should not block this PR however.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
KochTobi: I originally had `setDefaultError()` and that did not change the visibility of the error message.
As I want to make sure that the state of the view is reset I not only hide the error but overwrite it with the default one.
The method `setError()` would not communicate if the error is shown afterwards.

Steffengreiner (changes requested):
@KochTobi Thank you for your work! I like the idea of introducing a dedicated `ConfigureableLoginForm` to handle the login and using the vaadin component invalidates the necessity of a dedicated use case 👍  

However I found some issues with the PR:

**1. Login Functionality:**
- I'm currently able to login even with inactive users. Is this intended?
- I registered a new user (Steffen Greiner) and clicking on the confirmation link didn't work (User was not set to active and no `ConfirmationMessage` was shown)

**2. Notification Generation and Style:**
- The `ErrorMessage` shown upon login failure should be the one drafted by @jenniferboedker not the default vaadin one. 
- The `InformationLayout` showing the messages should be moved to the outer `LoginLayout` IMO. It's purpose is to show the result of the login and therefore does not have to be associated with the form itself. This would also make switching between the individual messages easier

webapp/frontend/styles/views/login/login-view.css
```diff
@@ -5,3 +5,11 @@
 .min-width-300px {
     min-width: 300px;
 }
+
+.form-title {
+    margin: 0;
+    margin-top: calc(var(--lumo-font-size-xxxl) - var(--lumo-font-size-xxl));
```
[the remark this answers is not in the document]
KochTobi: I copied the calculation from the default login view component.

webapp/frontend/styles/views/login/login-view.css
```diff
@@ -5,3 +5,11 @@
...
+    padding: 0;
+    padding-top: var(--lumo-space-l);
+    padding-left: var(--lumo-space-l);
```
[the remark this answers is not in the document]
KochTobi: Yes, this is another possibility. I decided to go for the declarative way and specify the desired paddings separately.

webapp/src/main/java/life/qbic/views/login/ConfigurableLoginForm.java
```diff
@@ -0,0 +1,104 @@
...
+ * <b>A login form that remembers the text it displays.</b>
+ *
+ * @since 1.0.0
+ */
+class ConfigurableLoginForm extends VerticalLayout {
+
+  private final LoginForm loginForm;
+
+  private final LoginI18n loginI18n;
+  private final HorizontalLayout informationMessageLayout;
```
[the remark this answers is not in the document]
KochTobi: As far as I know, this does not show the result of the login attempt but rather adds some information to the form. Thus I would like to leave it in the form.
I am happy to talk about that however :)

webapp/src/main/java/life/qbic/views/login/ConfigurableLoginForm.java
```diff
@@ -0,0 +1,104 @@
...
+  public void setAction(String action) {
+    loginForm.setAction(action);
+  }
+
+  public void setError(Message errorMessage) {
+    LoginI18n.ErrorMessage error = new LoginI18n.ErrorMessage();
+    error.setMessage(errorMessage.message());
+    error.setTitle(errorMessage.title());
+    loginI18n.setErrorMessage(error);
+    updateText();
```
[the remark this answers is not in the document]
KochTobi: The vaadin login form automatically shows the error message if the login was unsuccessful. AFAIK there is no way around that.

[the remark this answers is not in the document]
KochTobi: This gives us the possibility to change the message shown to the user.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -19,15 +18,20 @@
...
   private final ConfirmEmailInput confirmEmailInput;
+  private static final Message INCORRECT_USERNAME_OR_PASSWORD = new Message(
+      "Incorrect username or password",
+      "Check that you have entered the correct username and password and try again."
+  );
+
+  private static final Message EMAIL_CONFIRMATION_SUCCESS = new Message(
+      "Email address confirmed",
+      "You can now login with your credentials."
+  );
```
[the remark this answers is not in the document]
KochTobi: The handler is responsible to populate the view with values. As the displayed error/information messages are values they belong in the handler.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -37,25 +41,44 @@ public void handle(LoginLayout loginView) {
...
+    registeredLoginView.resetInformation();
+  }
+
+  private void hideError() {
+    registeredLoginView.hideError();
+  }
+
+  private void showDefaultError() {
+    registeredLoginView.showError(INCORRECT_USERNAME_OR_PASSWORD);
+  }
```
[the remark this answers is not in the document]
KochTobi: As I mentioned I would like to keep the default text in there to prevent random errors to show up. I like your suggestion however I would add another method like you suggested when we can differentiate between the errors. The method `showDefaultError()` can exist as a fallback.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -37,25 +41,44 @@ public void handle(LoginLayout loginView) {
...
-          checkUserEmail(registeredLoginView.email.getValue());
-        });
-    //ToDo Add forgot password Logic
+    registeredLoginView.addLoginListener(it -> onLoginSucceeded());
+    registeredLoginView.addForgotPasswordListener(it -> {/*TODO*/});
+  }
+
+  private void onLoginSucceeded() {
+    resetErrorMessages();
+    resetInformationMessages();
```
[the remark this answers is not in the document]
KochTobi: not possible

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -64,59 +87,23 @@ public void handle(BeforeEvent beforeEvent) {
...
+
   }
 
   @Override
   public void onEmailConfirmationFailure(String reason) {
-    resetMessages();
-    registeredLoginView.errorMessage.titleTextSpan.setText("Email confirmation failed");
-    registeredLoginView.errorMessage.descriptionTextSpan.setText(reason);
-    registeredLoginView.errorMessage.setVisible(true);
+    registeredLoginView.showError(new Message("Email confirmation failed", reason));
```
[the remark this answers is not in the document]
KochTobi: No it can not. In this case the reason of the displayed information is dynamic and provided by the output interface.

webapp/src/main/java/life/qbic/views/login/ConfigurableLoginForm.java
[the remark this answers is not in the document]
KochTobi: I moved the message handling to the layout

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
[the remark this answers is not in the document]
KochTobi: As the errors are never shown and hidden but added and removed this is not valid anymore

webapp/src/main/java/life/qbic/views/login/ConfigurableLoginForm.java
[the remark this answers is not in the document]
KochTobi: moved to the layout

webapp/src/main/java/life/qbic/views/login/LoginLayout.java
```diff
@@ -143,53 +104,47 @@ private void styleFormLayout() {
...
+    errorContainer.add(new ErrorMessage(errorMessage.title(), errorMessage.message()));
+    errorContainer.setVisible(true);
   }
 
-  private void styleLoginButton() {
-    loginButton.setWidthFull();
-    loginButton.addThemeVariants(ButtonVariant.LUMO_PRIMARY);
+  public void showInformation(InformationMessage message) {
+    informationContainer.add(new InformationMessage(message.title(), message.message()));
+    informationContainer.setVisible(true);
```
Steffengreiner, line 126: What's the reason for recreating the message instead of using the one in the parameter?
NIT: I still think that since the Messages themselves are components to be used in the view their creation belongs in the view. 
So you could also provide the title and message as Strings from the handler

webapp/src/main/java/life/qbic/views/login/LoginLayout.java
```diff
@@ -38,44 +36,45 @@ public class LoginLayout extends VerticalLayout implements HasUrlParameter<Strin
 
   private VerticalLayout contentLayout;
 
-  public PasswordField password;
+  private VerticalLayout informationContainer;
 
-  public EmailField email;
+  private VerticalLayout errorContainer;
```
Steffengreiner, line 41: I think you can replace this with just one "notificationContainer", since it's purpose is to contain the notification. The type will be determined by the notification that is shown.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -19,15 +19,20 @@
...
   private final ConfirmEmailInput confirmEmailInput;
+  private static final ErrorMessage INCORRECT_USERNAME_OR_PASSWORD = new ErrorMessage(
+      "Incorrect username or password",
+      "Check that you have entered the correct username and password and try again."
+  );
+
+  private static final InformationMessage EMAIL_CONFIRMATION_SUCCESS = new InformationMessage(
+      "Email address confirmed",
+      "You can now login with your credentials."
+  );
```
Steffengreiner, line 33: NIT: The creation of the message component could be handled by the view.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -37,25 +42,35 @@ public void handle(LoginLayout loginView) {
...
-        event -> {
-          resetMessages();
-          resetComponentErrors();
-          checkUserEmail(registeredLoginView.email.getValue());
-        });
-    //ToDo Add forgot password Logic
+    registeredLoginView.addLoginListener(it -> onLoginSucceeded());
+    registeredLoginView.addForgotPasswordListener(it -> {/*TODO*/});
+    registeredLoginView.addErrorListener(() -> System.out.println("error occurred"));
+  }
```
Steffengreiner, line 69: NIT: I tested this with invalid credentials and the listener is never triggered. 
One solution could be that the`handle` method could fire an event that triggers the listener which then sets the invalidcredentials message. 
Down the line we should also discuss if it's possible to extract the error reason from the beforeEvent parameter.

webapp/src/main/java/life/qbic/views/login/LoginLayout.java
KochTobi: they are components of the view but the handler populates it. We could introduce a `Message` record that only delivers the values to the view but I think that adds complexity we do not need right now.

webapp/src/main/java/life/qbic/views/login/LoginLayout.java
KochTobi: We need to have the possibility to handle errors and information separate from each other. As the login view serves multiple purposes (login and email confirmation)

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
KochTobi: As the handler decides which message is shown and added/removed, it belongs in the handler imho.

webapp/src/main/java/life/qbic/views/login/LoginHandler.java
KochTobi: Yes it is bypassed and I removed it now as we do not need it :+1: good spot

KochTobi: The login should never give a reason! It only fails.

Steffengreiner: I agree that it should not specify which credential is wrong but what about other kind of Exceptions (unexpected Exceptions)?

KochTobi: the user should not be informed that there was an unexpected exception.

KochTobi: that would expose internals

Steffengreiner: There should still be a way to differentiate between showing an invalid credential error message (prompting the user to check his input) and an error message that something went wrong on our side

PR #39, opened 2022-06-13
Introduce UserId concept

sven1103:
To make user object a conceptual whole value object, the String user id is converted to a DDD object UserId, using the domain language.


domain/src/main/java/life/qbic/apps/datamanager/services/UserRegistrationService.java
```diff
@@ -278,7 +279,8 @@ private void sendNotification(UserEmailConfirmed event) {
         notificationService.send(notification);
       }
     });
-    Optional<User> optionalUser = userRepository.findById(userId);
+
+    Optional<User> optionalUser = userRepository.findById(UserId.from(userId));
```
KochTobi, line 283: Can we somehow get the userid value object directly instead of recreating it?

KochTobi (approved):
LGTM

PR #40, opened 2022-06-14
Make LoginLayout spacing consistent with RegistrationLayout

Steffengreiner:
**Aim of this PR**

**1. Removes the standard vaadin spacings of the loginform and the notification containers in loginlayout:**
Allows us to set the padding and margins once in the contentlayout and not have to specify it again in the individual components
 
**2. Add Span around the Displaymessage icon to enable it to proberly adjust to multiline text in the title of a notification.** 
Ensures that the icon stays at the same size and position independent of the title text length.

**3. Adapts the Styling of the Displaymessage to make it distinct of ErrorMessage, InformationMessage and Successmessage.** 
The Displaymessage is currently just used as a Superclass for the specific implementations. 
One could think about making it Abstract to avoid accidental usage in the future?


sven1103:
Hi @Steffengreiner , 

I can't confirm actually, the spacing of the input fields look still different to me:

image
 
image

sven1103 (approved):
Great adjustment in the UI!

PR #41, opened 2022-06-20
Reorganise modules to make them more cohesive

sven1103:
This PR reorganises substantial parts of the former domain Maven module, 
no called `core`.

It introduces two new root packages: domain and application. The domain model knowledge is now structured in the first, and everything that belongs to application logic and domain object orchestration is in the ladder.

With this we hope to improve the class organisation and reflect the actual domain and application structure.


jenniferboedker (commented):
Thanks for giving the code more structure :) 
Thoughts I had while looking through your changes:

1. I would expect the folder 'identityaccess' to be also present in the webapp to make it clear which parts connect to which part of the core
2. The notification folder (and the Service and ApplicatinException) would be independent of the identityaccess right? I would expect them at a higher level shared with future use cases. Same goes for the domain events (folder).
3. Is the idea that each new use case will have the same folder structure as identityaccess? (e.g projectoverview > domain,application)

In general I would expect to be able to associate how the webapp connects to the application part in the core (see 1.). Happy to have a short discussion with you to clarify open questions :)

jenniferboedker (dismissed):
Looks good to me now. Would be great if the infrastucture will also be adapted eg in a follow-up PR

wow-such-code (approved):

PR #42, opened 2022-06-20
Inform user about registration email after successful registration

Steffengreiner:
**Purpose of this PR**

Introduces an `InformationMessage` in the `LoginLayout` after successful registration that advises the user to check his inbox to confirm the registration. 

Bildschirmfoto 2022-06-20 um 13 29 28

**Additional Information** 

This solution uses `QueryParameters` created in the `RegistrationHandler` to trigger the correct response in the `LoginHandler` similiar to the ConfirmEmail Use Case. 
Since the `QueryParameter` appear in the URL(http://localhost:8080/login?userRegistered=true) and therefore to the user, it would make sense to check if the parameters are specified according to our wishes. 

Message content inspired by stackoverflow Notification: 
Bildschirmfoto 2022-06-20 um 09 28 42


webapp/src/main/java/life/qbic/views/login/LoginHandler.java
```diff
@@ -35,6 +35,11 @@ public class LoginHandler implements LoginHandlerInterface, ConfirmEmailOutput {
       "You can now login with your credentials."
   );
 
+  private static final InformationMessage EMAIL_CONFIRMATION_REMINDER = new InformationMessage(
+      "Registration Email sent",
```
sven1103, line 39:
```suggestion
      "Registration email sent",
```

sven1103 (changes requested):
Thank you @Steffengreiner for your contribution.

Only one minor request

PR #43, opened 2022-06-21
Add domain model that includes the password reset

sven1103:
This PR adds only the domain model for the user's password request.

### DIFF
--- a/core/src/main/java/life/qbic/identityaccess/domain/user/PasswordReset.java
+++ b/core/src/main/java/life/qbic/identityaccess/domain/user/PasswordReset.java
@@ -45,7 +45,7 @@ public class PasswordReset extends DomainEvent {
   }
 
   /**
-   * Queries the user id of the user for whom the password needs to be reset
+   * Returns the user id of the user for whom the password needs to be reset
    *
    * @return the affected user's id
    * @since 1.0.0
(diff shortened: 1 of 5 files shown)


core/src/main/java/life/qbic/identityaccess/domain/user/PasswordReset.java
```diff
@@ -0,0 +1,56 @@
...
+    this.occurredOn = occurredOn;
+  }
+
+  @Override
+  public Instant occurredOn() {
+    return occurredOn;
+  }
+
+  /**
+   * Queries the user id of the user for whom the password needs to be reset
```
jenniferboedker, line 48: NIT: does not really query, does it? Is rather just returning the value

```suggestion
   * Returns the user id of the user for whom the password needs to be reset
```

jenniferboedker (approved):
Looks good to me!
Very easy to read, only one nitty comment. The test makes clear how to use it. Great work 🚀

core/src/main/java/life/qbic/identityaccess/domain/user/PasswordReset.java
sven1103: Ah, that is CQS lingua that snuck in https://en.wikipedia.org/wiki/Command%E2%80%93query_separation

PR #47, opened 2022-06-23
Set new user password - application logic

sven1103:
This PR adds application logic and use case interfaces
to support the set of a new user password.


Steffengreiner (changes requested):
Heya @sven1103 very well done. 👍 However i found some minor nitpicks. 
I'm also wondering, do you think it would make sense to implement a dedicated test for the application logic?

Steffengreiner (approved):
Nicely Done 👍

PR #48, opened 2022-06-23
Correct the use case input parameter

sven1103:
Since user do provide an email address and not their id, the use case
needs to be adjusted.


jenniferboedker (approved):
Looks good 🚀

PR #54, opened 2022-06-27
Introduce error notification for failure of password reset 

Steffengreiner:
**Purpose of this PR**
This PR introduces an error notification shown to the user if the password reset fails

**How to Test**
In the Password Reset View, input an email address not stored in the database and click the submit button. 
The Password Reset

**Possible Improvements**
The Handler could try to distinct between a password reset failure caused by a domain error a passwort reset failure caused by an invalid email input reset. 
To do so we would need to introduce a dedicated PasswortResetException similiar to the UserRegistrationException which can then be parsed from the ApplicationResponse via the output interface to the Handler

### DIFF
--- a/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
+++ b/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
@@ -81,14 +81,14 @@ public class PasswordResetHandler implements PasswordResetHandlerInterface, Pass
         showPasswordResetFailedError(failure.getMessage(), "Please provide a valid email address.");
       }
       else if (failure instanceof UserNotFoundException) {
-        showPasswordResetFailedError(failure.getMessage(), "No user registered for the email.");
+        showPasswordResetFailedError(failure.getMessage(), "No user registered for the provided email.");
       }
       else if (failure instanceof UserRegistrationService.UserNotActivatedException) {
         showPasswordResetFailedError(
-            failure.getMessage(), "The account needs to be active to reset the password.");
+            failure.getMessage(), "Activate your account to reset the password.");
       } else {
         showPasswordResetFailedError(
-            "An error occurred", "Please contact support@qbic.zendesk.com for help.");
+            "An unexpected error occurred", "Please contact support@qbic.zendesk.com for help.");
       }
     }
   }
--- a/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
+++ b/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
@@ -87,8 +87,6 @@ public class PasswordResetHandler implements PasswordResetHandlerInterface, Pass
         showPasswordResetFailedError(
             failure.getMessage(), "The account needs to be active to reset the password.");
       } else {
-        System.out.println(failure.getMessage());
-        System.out.println(failure.getClass());
         showPasswordResetFailedError(
             "An error occurred", "Please contact support@qbic.zendesk.com for help.");
       }
--- a/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
+++ b/webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
@@ -78,14 +78,14 @@ public class PasswordResetHandler implements PasswordResetHandlerInterface, Pass
   public void onPasswordResetFailed(ApplicationResponse response) {
     for (RuntimeException failure : response.failures()) {
       if (failure instanceof EmailAddress.EmailValidationException) {
-        showPasswordResetFailedError(failure.getMessage(), "Please provide a valid email address.");
+        showPasswordResetFailedError("Invalid email address format", "Please provide a valid email address.");
       }
       else if (failure instanceof UserNotFoundException) {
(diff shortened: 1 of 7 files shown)


webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -51,7 +66,6 @@ public void onPasswordResetSucceeded() {
 
     @Override
     public void onPasswordResetFailed() {
-        //todo
-        throw new NotImplementedException();
+        showPasswordResetFailedError();
```
sven1103, line 69: If we handle use case exceptions, i think we should do it with max satisfaction as possible to the user.

"The Password Reset has failed", "Please try again."

The user does not know, what they have done wrong. So i think we should increase transparency a little bit, and introduce methods in the output interface.

For example we can invoke a method `onUnknownEmailAddress()` if the email does not belong to any registered and active user account. We can further introduce a `onUnexpectedException()`, that will redirect the user to support@qbic.zendesk.com to handle more severe scenarios.

sven1103 (changes requested):
Dear @Steffengreiner thank you for your contribution. I think if we introduce error messages, we should make the experience for the user as good as possible (given that sth failed, will already lead to increased frustration levels).

I have made some suggestions to improve the handling.

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -3,55 +3,93 @@
...
+    for (RuntimeException failure : response.failures()) {
+      if (failure instanceof EmailAddress.EmailValidationException) {
+        showPasswordResetFailedError(failure.getMessage(), "Please provide a valid email address.");
+      }
+      else if (failure instanceof UserNotFoundException) {
+        showPasswordResetFailedError(failure.getMessage(), "No user registered for the provided email.");
+      }
+      else if (failure instanceof UserRegistrationService.UserNotActivatedException) {
+        showPasswordResetFailedError(
+            failure.getMessage(), "Activate your account to reset the password.");
```
sven1103, line 88:
```suggestion
            failure.getMessage(), "Please activate your account first to reset the password.");
```

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -3,55 +3,93 @@
...
-        //todo
-        throw new NotImplementedException();
+  @Override
+  public void onPasswordResetFailed(ApplicationResponse response) {
+    for (RuntimeException failure : response.failures()) {
+      if (failure instanceof EmailAddress.EmailValidationException) {
+        showPasswordResetFailedError(failure.getMessage(), "Please provide a valid email address.");
+      }
+      else if (failure instanceof UserNotFoundException) {
+        showPasswordResetFailedError(failure.getMessage(), "No user registered for the provided email.");
```
sven1103, line 84:
```suggestion
        showPasswordResetFailedError(failure.getMessage(), "No user with the provided email address is known.");
```

sven1103 (changes requested):
Only some minor phrasing suggestions

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -3,55 +3,93 @@
...
 
-    @Override
-    public void onPasswordResetFailed() {
-        //todo
-        throw new NotImplementedException();
+  @Override
+  public void onPasswordResetFailed(ApplicationResponse response) {
+    for (RuntimeException failure : response.failures()) {
+      if (failure instanceof EmailAddress.EmailValidationException) {
+        showPasswordResetFailedError(failure.getMessage(), "Please provide a valid email address.");
```
sven1103, line 81: I think we should not use the exception message here, and remain in control what we display to the user. Based on the exception type, we can define that easily.

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -3,55 +3,93 @@
...
+      }
+      else if (failure instanceof UserNotFoundException) {
+        showPasswordResetFailedError(failure.getMessage(), "No user with the provided email address is known.");
```
sven1103, line 84: same here, i think defining the message title explicit is more robust.

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
```diff
@@ -3,55 +3,93 @@
...
+      }
+      else if (failure instanceof UserRegistrationService.UserNotActivatedException) {
+        showPasswordResetFailedError(
+            failure.getMessage(), "Please activate your account first to reset the password.");
```
sven1103, line 88: same here

webapp/src/main/java/life/qbic/views/login/passwordreset/PasswordResetHandler.java
jenniferboedker: I am ok with that. Will do

sven1103 (approved):
great !

PR #59, opened 2022-06-28
Add service port configuration parameter

sven1103:
Enables the port configuration for the URLs
that are generated when user interaction
is requested via email.

The default -1 means, that the default port for the selected protocol is used.


jenniferboedker (commented):
Looks good to me! You want to update the readme to keep track of new env variables and how to set them up?

PR #64, opened 2022-06-30
fix/dm-447-password-encryption-policy-does-not-generate-random-hash

sven1103:


core/src/test/groovy/life/qbic/domain/usermanagement/policies/PasswordEncryptionPolicySpec.groovy
```diff
@@ -82,4 +82,21 @@ class PasswordEncryptionPolicySpec extends Specification {
         password = "abcdefghihdeo"
     }
 
+    def "Equal raw user passwords must not create the same secret" () {
```
Steffengreiner, line 85: Very nice! However I'd also suggest adding a check for the entire key length(similiar to the salt length check) to ensure the key adheres to the defined bit length 👍

Steffengreiner (dismissed):
Heya @sven1103 Very nice implementation. 
I'd have a suggestion concerning a check for the entire key length which would avoid having shortened key in the future.

Steffengreiner (approved):
Perfect 👍

PR #67, opened 2022-07-14
Enable removal of all subscribers from a domain publisher instance

sven1103:


core/src/main/java/life/qbic/identityaccess/application/user/Registration.java
```diff
@@ -67,6 +67,7 @@ public void register(String fullName, String email, char[] rawPassword) {
               response -> registerUserOutput.onUnexpectedFailure(build(response)));
     } catch (Exception e) {
       log.error("User registration failed", e.getCause());
+      log.error(e.getMessage());
```
KochTobi, line 70: Why not 
```suggestion
      log.error("User registration failed", e);
```

sven1103: Tbh, I dont really understand how the exception logging works. I just observed, that passing the throwable cause does not print anything in the logs, which made it impossible for debugging.

core/src/main/java/life/qbic/identityaccess/application/user/UserRegistrationService.java
```diff
@@ -80,6 +80,14 @@ public ApplicationResponse registerUser(final String fullName, final String emai
     if (userDomainService.isEmpty()) {
       throw new RuntimeException("User registration failed.");
     }
+
+    DomainEventPublisher domainEventPublisher = DomainEventPublisher.instance();
+    while (!domainEventPublisher.clear()) {
+      try {
+        Thread.sleep(1);
+      } catch (InterruptedException ignored) {
+      }
```
KochTobi, line 89: As I recently learned, swallowing exceptions might lead to a huge overhead when trying to figure out what happened.
(https://www.youtube.com/watch?v=LRwCE7GreSM)
In this case, I would ask that we log the error here or rethrow a RuntimeException and handle the exception on an outer level.
```suggestion
      } catch (InterruptedException ignored) {
        log.error(ignored);
      }
```
```suggestion
      } catch (InterruptedException ignored) {
        throw new RuntimeException("Unexpected interrupt.", ignored);
      }
```

core/src/main/java/life/qbic/identityaccess/application/user/UserRegistrationService.java
```diff
@@ -173,7 +181,16 @@ public ApplicationResponse requestPasswordReset(String userEmailAddress) {
...
       return ApplicationResponse.failureResponse(new UserNotActivatedException("User not active"));
     }
-    DomainEventPublisher.instance().subscribe(new DomainEventSubscriber<PasswordReset>() {
+
+    DomainEventPublisher domainEventPublisher = DomainEventPublisher.instance();
+    while (!domainEventPublisher.clear()) {
+      try {
+        Thread.sleep(1);
+      } catch (InterruptedException ignored) {
+      }
```
KochTobi, line 190: As I recently learned, swallowing exceptions might lead to a huge overhead when trying to figure out what happened.
(https://www.youtube.com/watch?v=LRwCE7GreSM)
In this case, I would ask that we log the error here or rethrow a RuntimeException and handle the exception on an outer level.
```suggestion
      } catch (InterruptedException ignored) {
        log.error(ignored);
      }
```
```suggestion
      } catch (InterruptedException ignored) {
        throw new RuntimeException("Unexpected interrupt.", ignored);
      }
```

core/src/main/java/life/qbic/identityaccess/application/user/UserRegistrationService.java
```diff
@@ -256,7 +273,14 @@ public UserExistsException() {
    * @since 1.0.0
    */
   public void confirmUserEmail(String userId) throws UserNotFoundException {
-    DomainEventPublisher.instance().subscribe(new DomainEventSubscriber<UserActivated>() {
+    DomainEventPublisher domainEventPublisher = DomainEventPublisher.instance();
+    while (!domainEventPublisher.clear()) {
+      try {
+        Thread.sleep(1);
+      } catch (InterruptedException ignored) {
+      }
```
KochTobi, line 281: As I recently learned, swallowing exceptions might lead to a huge overhead when trying to figure out what happened.
(https://www.youtube.com/watch?v=LRwCE7GreSM)
In this case, I would ask that we log the error here or rethrow a RuntimeException and handle the exception on an outer level.
```suggestion
      } catch (InterruptedException ignored) {
        log.error(ignored);
      }
```
```suggestion
      } catch (InterruptedException ignored) {
        throw new RuntimeException("Unexpected interrupt.", ignored);
      }
```

core/src/main/java/life/qbic/identityaccess/application/user/UserRegistrationService.java
```diff
@@ -303,11 +327,12 @@ private void sendNotification(UserEmailConfirmed event) {
 
   /**
    * <p>
-   * An exception to be thrown if an user is not activated.
-   * This implies that the user cannot login into the application
+   * An exception to be thrown if an user is not activated. This implies that the user cannot login
+   * into the application
```
KochTobi, line 331:
```suggestion
   * An exception to be thrown if a user is not activated. This implies that the user cannot log in
   * to the application
```

core/src/main/java/life/qbic/shared/domain/events/DomainEventPublisher.java
```diff
@@ -56,4 +56,23 @@ public <T extends DomainEvent> void publish(final T domainEvent) {
...
+
+  /**
+   * Removes all subscribers from the current thread's domain event subscription list.
+   * <p>
+   * This method is not interrupting current publishing domain event processes. In this case, the
+   * method will return <code>false</code>.
+   *
+   * @return false, when there are currently domain events published to signal that the removal was
+   * not performed. If the method returns true, the publisher was not in the process of publishing
+   * and all subscribers have been removed.
```
KochTobi, line 68:
```suggestion
   * @return false, when there are currently domain events published to signal that the removal was
   * not performed; true, when the publisher was not in the process of publishing
   * and all subscribers have been removed.
```

KochTobi (changes requested):
As I understand this implementation, the DomainEventPublisher is asked to clear every second by every client that requested the instance. Is this correct?

We have to make sure not to swallow exceptions. Either we have to implement some form of logging here or we have to propagate the information outwards.

KochTobi (approved):
Hi @sven1103 
Thank you for addressing my comments. The last suggestion is NIT so feel free to ignore it.
Nice work!

PR #68, opened 2022-07-14
Avoid artifact repackaging during deployment of release workflow

Steffengreiner:
**What has changed**
Adds the deploy goal to mvn lifecycle deploy phase. 
If only a phase is defined, then all the phases before that phase will be redone repackaging our artifact without the specified profile.

**Additional information**
Information about maven lifecycles can be found [here:](https://baeldung-cn.com/maven-goals-phases)


sven1103 (approved):
Cool, looking forward seeing it in action!

PR #69, opened 2022-07-21
Re-arrange architectural composition

sven1103:
Changes the data manager architecture into components (represented by Maven modules) and makes use of Java modules to enforce stronger encapsulation of the components' implementation detail.

We draft four two modules (BLUE), that will contain domain logic: authentication and authorization. 

In addition, we introduce a module for broadcasting information through the application but decouple it from the domains as good as possible and still allow for gossip. 

The newsreader module holds the current logic of consuming messages from the broadcasting module by implementing the `MessageSubscriber` interface of the broadcasting module. When required later, this can be divided into own sub-components and realize own (micro)services if required.


sven1103:
Architectural composition:

image

newsreader/src/main/java/life/qbic/newsreader/usermanagement/email/EmailSubmissionService.java
```diff
@@ -1,4 +1,4 @@
-package life.qbic.usermanagement;
+package life.qbic.newsreader.usermanagement.email;
```
jenniferboedker, line 1: Why is the service for sending emails located in the news reader? I would expect it somewhere close to the domain but not the component that is supposed to consume the message and trigger an event. 
trigger an event != containing the service that is triggered

authentication/src/main/java/life/qbic/authentication/application/notification/Notification.java
```diff
@@ -1,7 +1,7 @@
-package life.qbic.shared.application.notification;
+package life.qbic.authentication.application.notification;
```
jenniferboedker, line 1: Isnt the notification and the notification service something broader and should be part of the messaging of the software (newsreader + broadcasting)?

jenniferboedker (commented):
Thanks for tackling the clean up of our current architecture!
I left some comments where I did not expect some classes to appear. Another question I have is: your current structure implies to me that for each new topic we introduce e.g. project or datasets or samples,... we need to create a new module?

sven1103:
As discussed @jenniferboedker 

image

broadcasting/src/main/java/life/qbic/broadcasting/MessageBusSubmission.java
```diff
@@ -0,0 +1,6 @@
+package life.qbic.broadcasting;
+
```
jenniferboedker, line 2: Missing java doc

broadcasting/src/main/java/life/qbic/broadcasting/MessageSubscription.java
```diff
@@ -0,0 +1,7 @@
+package life.qbic.broadcasting;
+
```
jenniferboedker, line 2: missing java doc

authentication/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
...
+  requires life.qbic.broadcasting;
+
+  exports life.qbic.authentication.domain.user.concept;
+  exports life.qbic.authentication.application;
+  exports life.qbic.authentication.domain.user.repository;
+  exports life.qbic.authentication.domain.user.event;
+  exports life.qbic.authentication.domain.user.policy;
+  exports life.qbic.authentication.application.user.registration;
+  exports life.qbic.authentication.application.user.password;
+  exports life.qbic.authentication.application.notification;
```
jenniferboedker, line 13: Do you need to specify those if you already export the parent package of all of those?

jenniferboedker: see line 7

authentication/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
+module life.qbic.authentication {
+  requires org.slf4j;
+  requires java.persistence;
+  requires life.qbic.broadcasting;
+
+  exports life.qbic.authentication.domain.user.concept;
```
jenniferboedker, line 6: for the domain you could also just export the complete package since you list all of the subpackages here

newsreader/src/main/java/module-info.java
```diff
@@ -0,0 +1,8 @@
+module newsreader {
+  requires spring.beans;
+  requires spring.context;
+  requires jakarta.mail;
+  exports life.qbic.newsreader.usermanagement.registration;
+  exports life.qbic.newsreader.usermanagement.email;
+  exports life.qbic.newsreader.usermanagement.passwordreset;
```
jenniferboedker, line 7: why not exporting the complete usermanagement package?

authentication/src/main/java/module-info.java
sven1103: Sub-packages are not included in the export - All exports are specific to the package level that is defined.

authentication/src/main/java/module-info.java
sven1103: wild card exports are not possible and would be also bad practise

jenniferboedker (approved):

jenniferboedker (approved):
🚀

jenniferboedker (changes requested):
Naming of the news reader needs to be adapted

jenniferboedker (approved):

PR #70, opened 2022-07-26
Prepare release

sven1103:


authentication/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>datamanager</artifactId>
+    <groupId>life.qbic</groupId>
+    <version>0.4.3</version>
```
Steffengreiner, line 8: Why is the version 0.4.3 if 0.4.4 is already released?

Steffengreiner (approved):
Heya Sven there seems to be a versioning mismatch between the already released version 0.4.4 and the version in this pom 0.4.3. 
However this can be fixed by triggering the release workflow version with 0.4.5

PR #73, opened 2022-07-27
Open field access for public response type

sven1103:
Opens the ApplicationResponse.Type enum, since it is a return type of the public API.


KochTobi (commented):
Why not introduce accessor methods
`isSuccess()` and `isFailure()`

KochTobi (approved):

PR #74, opened 2022-07-29
Add logging compontent

sven1103:
Provides the central logging component that can be used  by all other components in the application to log certain events.

The component offers to hook into log messages and create plugins to take action on certain events, for example on error messages to write an email to all developers.


sven1103:
Logging component architecture layout:

image

logging/src/main/java/life/qbic/logging/api/LogMessage.java
```diff
@@ -0,0 +1,17 @@
...
+ *   <li>log level</li>
+ *   <li>message</li>
+ *   <li>throwable</li>
+ * </ul>
+ *
+ * @since 1.0.0
+ */
+public record LogMessage(String application, LogLevel logLevel, String message,
+                         Throwable throwable) {
+
```
KochTobi, line 16: What is the throwable meant to represent? Is it the cause for the message, is it related to the message directly? Can there be messages without throwables?

logging/src/main/java/life/qbic/logging/api/Logger.java
```diff
@@ -0,0 +1,44 @@
...
+  void error(String message);
+
+  /**
+   * Logs a message on ERROR level with a {@link Throwable} provided
+   *
+   * @param message the message content
+   * @param t       the throwable to report
+   * @since 1.0.0
+   */
+  void error(String message, Throwable t);
```
KochTobi, line 34: What is the throwable meant to be here? Is it the cause for the error, is it something else? do we want to specify that further?

logging/src/main/java/life/qbic/logging/api/Logger.java
```diff
@@ -0,0 +1,44 @@
+package life.qbic.logging.api;
+
+/**
+ * Simplistic interface of a {@link Logger}, that enables basic logging functionality for different
+ * log levels.
+ *
+ * @since 1.0.0
+ */
+public interface Logger {
```
KochTobi, line 9: I am missing the inclusion of a throwable to debug. Further I did not see any access to the warn log level here. Is this intentional?

logging/src/main/java/life/qbic/logging/impl/logger/LoggerFacade.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public static LoggerFacade from(String name, Publisher publisher) {
+    Objects.requireNonNull(publisher);
+    return new LoggerFacade(name, publisher);
+  }
+
+  public static LoggerFacade from(Class<?> clazz, Publisher publisher) {
+    Objects.requireNonNull(publisher);
+    return from(clazz.getName(), publisher);
+  }
```
KochTobi, line 40: Why do we need to expose both endpoints? Wouldn't the class one suffice?

logging/src/main/java/life/qbic/logging/impl/logger/LoggerFacade.java
```diff
@@ -0,0 +1,94 @@
...
+    logError(message);
+  }
+
+  private LogMessage create(LogLevel logLevel, String message) {
+    return create(logLevel, message, null);
+  }
+
+  private LogMessage create(LogLevel logLevel, String message, Throwable t) {
+    return new LogMessage(slf4jWrapper.name(), logLevel, message, t);
+  }
```
KochTobi, line 65: What is the throwable t here?

logging/src/main/java/life/qbic/logging/impl/publisher/PublisherFactory.java
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Creates a simple publisher that includes a {@link EmailOnErrorSubscriber} subscription.
+   *
+   * @return an instance of type {@link Publisher}
+   * @since 1.0.0
+   */
+  public static Publisher createSimplePublisher() {
+    var publisher = new SimplePublisher();
+    var emailOnErrorSubscriber = new EmailOnErrorSubscriber();
+    publisher.subscribe(emailOnErrorSubscriber);
```
KochTobi, line 22: I would not expect the returned publisher to have a subscriber already attached to it. Can you motivate your thoughts here a bit?

logging/src/main/java/life/qbic/logging/impl/publisher/SimplePublisher.java
```diff
@@ -0,0 +1,119 @@
...
+  protected synchronized LogMessage nextLogMessage() {
+    while (logMessages.isEmpty()) {
+      try {
+        wait();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    }
+    return logMessages.remove();
+  }
```
KochTobi, line 67: As you modify the `logMessages` queue here, do you need to notify any other thread?

logging/src/main/java/life/qbic/logging/impl/publisher/SimplePublisher.java
```diff
@@ -0,0 +1,119 @@
...
+      var logMessage = publisher.nextLogMessage();
+      if (isNull(logMessage)) {
+        return;
+      }
+      submit(logMessage);
+    }
+
+    private void submit(LogMessage logMessage) {
+      synchronized (subscribers) {
+        subscribers.forEach(s -> s.onNewMessage(logMessage));
```
KochTobi, line 114:
```suggestion
        subscribers.forEach(subscriber -> subscriber.onNewMessage(logMessage));
```
Please try to avoid single letter variable names, even in lambda expressions.

logging/src/main/java/life/qbic/logging/impl/slf4j/Slf4jWrapper.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @since 1.0.0
+ */
+public class Slf4jWrapper {
+
+  private final Logger logger;
+
+  public static Slf4jWrapper create(String name) {
+    return new Slf4jWrapper(LoggerFactory.getLogger(name));
+  }
```
KochTobi, line 18: Why do we need the String constructor?

logging/src/main/java/life/qbic/logging/impl/slf4j/Slf4jWrapper.java
```diff
@@ -0,0 +1,49 @@
...
+  private Slf4jWrapper(Logger logger) {
+    this.logger = logger;
+  }
+
+  public void debug(String message) {
+    logger.debug(message);
+  }
+
+  ;
+
```
KochTobi, line 33: ? 
```suggestion
```

logging/src/main/java/life/qbic/logging/service/LoggerFactory.java
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Creates a simple logger that carries the provided name.
+   *
+   * @param name the logger instance name provided by the client
+   * @return a {@link Logger} instance
+   * @since 1.0.0
+   */
+  public static Logger logger(String name) {
+    return LoggerFacade.from(name, publisher);
+  }
```
KochTobi, line 27: Why do we need both, the string and the class methods?

logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
+class LoggerFacadeSpec extends Specification {
+
+    def "On info message, publish the log message and call the logger binding"() {
```
KochTobi, line 16:
```suggestion
    def "When an info message is logged, all subscribers are notified and the message is contained in the log"() {
```

logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy
```diff
@@ -0,0 +1,76 @@
...
+        when:
+        logger.info("test")
+
+        then:
+        1 * publisher.publish(_ as LogMessage)
+        byteOutputStream.toString().contains("test")
+
+    }
+
+    def "On error message, publish the log message and call the logger binding"() {
```
KochTobi, line 36:
```suggestion
    def "When an error message is logged, all subscribers are notified and the message is contained in the log"()```

File: logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy [hunk 14]
```diff
@@ -0,0 +1,76 @@
...
+        when:
+        logger.error("test")
+
+        then:
+        1 * publisher.publish(_ as LogMessage)
+        byteOutputStream.toString().contains("test")
+
+    }
+
+    def "On debug message, publish the log message and call the logger binding"() {
```
KochTobi, line 56:
```suggestion
    def "When a debug message is logged, all subscribers are notified and the message is contained in the log"() {
    ```

File: logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy [hunk 15]
```diff
@@ -0,0 +1,76 @@
...
+import spock.lang.Specification
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
+class LoggerFacadeSpec extends Specification {
```
KochTobi, line 14: There are no tests containing throwables? Was this by design?

logging/src/test/groovy/life/qbic/logging/impl/publisher/SimplePublisherSpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.charset.Charset
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
+class SimplePublisherSpec extends Specification {
```
KochTobi, line 17: There is no test for the other log levels and a throwable payload. Is this by design?

KochTobi (changes requested):
Hi Sven,
That is a lot of code! :D 
I left some comments with open questions. Thank you for being so open and taking the time to clarify them.

logging/src/main/java/life/qbic/logging/api/LogMessage.java
```diff
@@ -0,0 +1,17 @@
...
+ *   <li>application</li>
+ *   <li>log level</li>
+ *   <li>message</li>
+ *   <li>throwable</li>
+ * </ul>
+ *
+ * @since 1.0.0
+ */
+public record LogMessage(String application, LogLevel logLevel, String message,
+                         Throwable throwable) {
```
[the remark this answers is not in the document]
KochTobi: Just noticed that this is a record. equals and hashCode are implemented by default.
Open question: do we need an id? otherwise the content is used to compare log messages

logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy
KochTobi: NIT

logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy
KochTobi: NIT

logging/src/test/groovy/life/qbic/logging/impl/logger/LoggerFacadeSpec.groovy
KochTobi: NIT

vaadinfrontend/src/main/java/module-info.java
```diff
@@ -23,14 +22,15 @@
...
-  requires org.apache.logging.log4j;
   requires spring.core;
   requires spring.web;
   requires org.apache.tomcat.embed.core;
   requires spring.data.commons;
   requires life.qbic.mariadbconnector;
   requires spring.data.jpa;
   requires life.qbic.commons;
+  requires life.qbic.logging;
+  requires jakarta.activation;
```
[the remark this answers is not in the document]
sven1103: Nice spot, this is an artifact. I will remove it.

newshandler/src/main/java/module-info.java
```diff
@@ -1,9 +1,10 @@
 module life.qbic.newshandler {
   requires spring.beans;
   requires spring.context;
-  requires jakarta.mail;
   requires life.qbic.commons;
-  requires org.slf4j;
+  requires life.qbic.logging;
+  requires jakarta.mail;
+  requires jakarta.activation;
```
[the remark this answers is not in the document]
sven1103: This was introduced since I observed that sending emails after introducing modules did not work anymore. That
was a bugfix on the fly.

logging/src/test/groovy/module-info.java
```diff
@@ -0,0 +1,9 @@
+open module life.qbic.logging {
+  requires org.slf4j;
+  requires logback.core;
+  requires org.spockframework.core;
+  requires org.apache.groovy;
+  requires java.desktop;
+
+  exports life.qbic.logging.impl.slf4j;
```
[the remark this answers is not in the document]
sven1103: Artifact, will be removed

logging/src/main/java/life/qbic/logging/api/LogMessage.java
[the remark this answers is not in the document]
sven1103: I don't yet see a use case to introduce an ID for LogMessages. If you think that helps mapping logging events to log entries, happy to add one.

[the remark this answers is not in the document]
KochTobi: No just wanted to bring the question up in case they need to be identified by something other than their values. As I do not see a reason for that either, I will resolve this comment.

logging/src/main/java/life/qbic/logging/api/Logger.java
sven1103: I will rename it to `cause`, to make it more explicit.

logging/src/main/java/life/qbic/logging/impl/logger/LoggerFacade.java
sven1103: The `Class` gives you the fully qualified identifier of the class, which sometimes can be undesired. Therefore I provide both possibilities as API and following the convention of the slf4j API.

logging/src/main/java/life/qbic/logging/impl/logger/LoggerFacade.java
sven1103: the cause, i address it

logging/src/main/java/life/qbic/logging/impl/publisher/PublisherFactory.java
sven1103: Sure :) We need to encapsulate our subscriber policies somewhere and not give the responsibility to the application. I imagine to have some default subscriber like `EmailOnError` and `SlackMessageOnError` that are hooked in all the time and the application can provide implementations (still need to define the interface for these) or if not, provide no-op implementations. With this we can decide within the logging component, that we want a slack message subscriber executed for all errors, write an email for warnings to our mailing list or whatever. We can decide on application level to provide an implementation for sending emails or slack messages or default to no-op.
But the rules are defined in the component, that is where i think it belongs. Therefore I preconfigure the publisher in the factory :)

logging/src/main/java/life/qbic/logging/impl/publisher/SimplePublisher.java
sven1103: I think this is implemented in the ArrayBlockingQueue class of the Java collections package

logging/src/main/java/life/qbic/logging/impl/publisher/PublisherFactory.java
KochTobi: Thanks for the elaborate motivation :+1:
I it makes sense to have a default subscription strategy in place.

logging/src/main/java/life/qbic/logging/impl/logger/LoggerFacade.java
KochTobi: The `Class::getName` provides you with a full identifier of the class. In this case `life.qbic.logging...LoggerFacade`
Is this the desired information?

logging/src/main/java/life/qbic/logging/impl/publisher/SimplePublisher.java
sven1103: reading the API i switch it to `poll` instead of `remove`, because i dont want an ElementNotFoundException thrown here, but return null if the queue is empty.

KochTobi: I could not infer that from the documentation.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ArrayBlockingQueue.html
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/AbstractQueue.html#remove()

logging/src/main/java/life/qbic/logging/impl/slf4j/Slf4jWrapper.java
sven1103: see Logger Facade API

logging/src/main/java/life/qbic/logging/service/LoggerFactory.java
sven1103: yes :) see my API explanation above

logging/src/test/groovy/life/qbic/logging/impl/slf4j/Slf4jWrapperSpec.groovy
```diff
@@ -0,0 +1,79 @@
...
+
+        and:
+        def errorMessage = "Error message"
+        def throwable = new Throwable("Out of coffee")
+
+        when:
+        wrapper.error(errorMessage, throwable)
+
+        then:
+        byteOutputStream.toString().contains(errorMessage) && byteOutputStream.toString().contains(throwable.toString())
```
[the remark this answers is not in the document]
sven1103: stack traces are properly handled by the logging libraries (logback, slf4j2, logging, etc)

KochTobi (approved):

PR #75, opened 2022-08-03
Remove logback dependency from logging component

sven1103:
Removes the logback dependency from the logging component
and uses the slf4j api only. 

This allows us to define the logging framework on application level
and are fully flexible in the configuration.


KochTobi (approved):
LGTM

PR #76, opened 2022-08-04
Add email on error subscriber provider

sven1103:
For configuration see the `mail.properties` file in the vaadin-frontend module resource folder.


sven1103:
image

Steffengreiner (approved):
Heya @sven1103 Thanks for the detailed explanation! 
I can confirm that the logging email was sent as expected and therefore think this PR is g2g!

PR #77, opened 2022-09-09
Create project domain objects

KochTobi:


projectmanagement/src/main/java/life/qbic/projectmanagement/Project.java
```diff
@@ -0,0 +1,57 @@
...
+   * Creates a new project with code and project intent
+   *
+   * @param projectIntent the intent of the project
+   * @return a new project instance
+   */
+  public static Project create(ProjectIntent projectIntent) {
+    UUID uuid = UUID.randomUUID();
+    return new Project(uuid, projectIntent);
+  }
+
```
sven1103, line 34: Missing javadocs

projectmanagement/src/main/java/life/qbic/projectmanagement/ProjectIntent.java
```diff
@@ -0,0 +1,14 @@
+package life.qbic.projectmanagement;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A project intent contains information on the project that is related to the intent of the
+ * project.
+ */
+public record ProjectIntent(ProjectTitle projectTitle) {
```
sven1103, line 9: Afaik, records introduce public constructors, which I would like to avoid to create expressive object APIs with named static factory methods.

projectmanagement/src/main/java/life/qbic/projectmanagement/offer/OfferId.java
```diff
@@ -0,0 +1,39 @@
...
+    String[] splits = offerId.split("_");
+    if (splits.length != 4) {
+      throw new ProjectManagementDomainException(
+          "OfferId does not contain 4 parts. Provided: " + offerId);
+    }
+    ProjectConservedPart projectConservedPart = new ProjectConservedPart(splits[1]);
+    RandomPart randomPart = new RandomPart(splits[2]);
+    long versionNumber = Long.parseLong(splits[3]);
+    Version version = new Version(versionNumber);
+    return new OfferId(projectConservedPart, randomPart, version);
```
sven1103, line 36: This is business logic from the finance and accounting domain and should not reside in the project management domain.

I suggest to use a simple internal character sequence representation of the offer id, until we have a reason to change it.

projectmanagement/src/main/java/life/qbic/projectmanagement/offer/RandomPart.java
```diff
@@ -0,0 +1,17 @@
...
+import life.qbic.projectmanagement.ProjectManagementDomainException;
+
+public record RandomPart(String randomPart) {
+
+  public RandomPart {
+    requireNonNull(randomPart);
+    if (randomPart.isEmpty()) {
+      throw new ProjectManagementDomainException(
+          "Random part of an offer id must not be empty. Provided: " + randomPart);
+    }
```
sven1103, line 14: Business logic from another domain

projectmanagement/src/main/java/life/qbic/projectmanagement/offer/Version.java
```diff
@@ -0,0 +1,14 @@
...
+
+import life.qbic.projectmanagement.ProjectManagementDomainException;
+
+public record Version(long number) {
+
+  public Version {
+    if (number <= 0) {
+      throw new ProjectManagementDomainException(
+          "Offer version must be greater than 0. Provided: " + number);
+    }
```
sven1103, line 11: Business logic from another domain

sven1103 (changes requested):
Dear @KochTobi , thanks a lot for your contribution. I left some comments to discuss. I have only major concerns with the domain logic of the offer id, which does not belong into the project management domain, but in the finance and accounting domain. This can be greatly simplified.

projectmanagement/src/main/java/life/qbic/projectmanagement/ProjectIntent.java
KochTobi: I think that `records` here are perfect for implementing value objects.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Record.html
Records have components (in this case a ProjectTitle). As every component needs to be provided for a record to be complete (or in this case a project intent). You can, however, write custom constructors or provide static factory methods. For value objects, we need to provide all components. Every internal value of the value object must be verified in any case, for this the compact constructors of records are perfect IMHO.

KochTobi: We further benefit from the guaranteed correct implementation of `equals and hashCode` as well as the `toString` and immutability of java records

projectmanagement/src/main/java/life/qbic/projectmanagement/offer/OfferId.java
KochTobi: Yes this domain did not exist when I implemented these classes. That is why I implemented them :D

projectmanagement/src/main/java/life/qbic/projectmanagement/ProjectIntent.java
sven1103: @KochTobi ok, convinced. But let's keep an eye on the object APIs

PR #79, opened 2022-09-13
Provide application service for offer search

sven1103:


projectmanagement/src/main/java/life/qbic/projectmanagement/application/OfferSearchService.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * <b>Offer Search Service</b>
+ *
+ * Enables search queries for offers.
+ *
+ * @since 1.0.0
+ */
+@Service
+public class OfferSearchService {
```
jenniferboedker, line 18: I expected this Service to be the ProjectCreationService. This service then uses the offersearchservice to query for an offer. For me its not so intuitive that we have two OfferSearchServices. Why do we need a service that calls an equally called service from another layer?

sven1103: Ah yes, that makes a lot of sense. I was struggling with finding the proper class name.

jenniferboedker (approved):
🚀  looks good

PR #86, opened 2022-09-16
Provide reusable result class to improve method signatures

sven1103:
Introduces a new class "Result" that enables to wrap result values after successful method execution or exceptions, that can further be processed with a fluent API.


commons/src/test/groovy/life/qbic/application/commons/ResultSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+class ResultSpec extends Specification {
+
+    def "run"() {
+        when:
+        String s = "hi"
+
+        then:
+        assert s == "hi"
+
+    }
```
jenniferboedker, line 18:
```suggestion
```

jenniferboedker (approved):

commons/src/main/java/life/qbic/application/commons/Result.java
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param function a function transforming data of type <code>V</code> to <code>U</code>
+   * @return a new result object instance of type <code>U,E</code>
+   */
+  public <U> Result<U, E> map(Function<V, U> function) {
+    Objects.requireNonNull(function);
+    Result<U, E> result = null;
+    switch (this.type) {
+      case EXCEPTION -> result = Result.exception(this.exception);
+      case SUCCESS -> result = apply(function, this.getValue());
```
KochTobi, line 151: this might return another exception type

KochTobi: duplicated

commons/src/main/java/life/qbic/application/commons/Result.java
```diff
@@ -0,0 +1,242 @@
...
+   * Static constructor method for creating a result object instance of type <code>V,E</code>
+   * wrapping an actual value <code>V</code>.
+   *
+   * @param value the notorious value to get wrapped in a result object
+   * @return a new result object instance
+   */
+  public static <V, E extends Exception> Result<V, E> success(V value) {
+    Objects.requireNonNull(value);
+    return new Result<>(value);
+  }
```
KochTobi, line 61:
```suggestion
  public static <V, E extends Exception> Result<V, E> value(V value) {
    Objects.requireNonNull(value);
    return new Result<>(value);
  }
```

commons/src/main/java/life/qbic/application/commons/Result.java
```diff
@@ -0,0 +1,242 @@
...
+   * <code>U,E</code>.</p>
+   *
+   * <p>If the current result contains an error (and therefore has no value), the created result
+   * object will contain the error of type <code>E</code> of the input result object.
+   * </p>
+   *
+   * @param function a function transforming data of type <code>V</code> to <code>U</code>
+   * @return a new result object instance of type <code>U,E</code>
+   */
+  public <U> Result<U, ? extends E> map(Function<V, U> function) {
```
KochTobi, line 147:
```suggestion
  public <U> Result<U, ? extends Exception> map(Function<V, U> function) {
```
This needs to extend Exception itself. If you have a Result<Integer,IllegalArgumentException> res and you fail to do res.map(it -> it.fooBar()) then you get a Result<Integer,NoSuchOperationException> or similar. These exceptions do not have any relationship except that they are exceptions

commons/src/main/java/life/qbic/application/commons/Result.java
```diff
@@ -0,0 +1,242 @@
...
+    Objects.requireNonNull(function);
+    Result<U, ? extends E> result = null;
```
KochTobi, line 149:
```suggestion
    Result<U, ? extends Exception> result = null;
```

commons/src/main/java/life/qbic/application/commons/Result.java
```diff
@@ -0,0 +1,242 @@
...
+
+  private <U> Result<U, E> apply(Function<V, U> function, V value) {
+    Result<U, E> result;
+    try {
+      result = Result.success(function.apply(value));
+    } catch (Exception e) {
+      result = (Result<U, E>) Result.failure(e);
+    }
+    return result;
+  }
```
KochTobi, line 241:
```suggestion
  private <U> Result<U, ? extends Exception> apply(Function<V, U> function, V value) {
    Result<U, ? extends Exception> result;
    try {
      result = Result.success(function.apply(value));
    } catch (Exception e) {
      result = Result.failure(e);
    }
    return result;
  }
```

commons/src/main/java/life/qbic/application/commons/Result.java
sven1103: we create either a Result object of type success or failure

KochTobi (approved):
LGTM

PR #88, opened 2022-09-19
Implement Create Project View

KochTobi:
image


jenniferboedker (commented):
Hi 👋🏼 ,

regarding the UI: I think I would create a new page to put the content in to have it separated from the other views. I would see the project creation as an own page. Furthermore, it would be cool if the content would be placed in the card like component (see prototype).

vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/CreateProjectHandlerInterface.java
```diff
@@ -0,0 +1,7 @@
+package life.qbic.datamanager.views.project.create;
+
+public interface CreateProjectHandlerInterface {
+
+  void handle(CreateProjectLayout createProjectLayout);
+
+}
```
jenniferboedker, line 7: Java doc

jenniferboedker (approved):

PR #96, opened 2022-09-22
Extend project creation by optional experimental design

KochTobi:


projectmanagement/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
```diff
@@ -31,32 +35,74 @@ public ProjectCreationService(ProjectRepository projectRepository) {
    */
   public Result<Project, ApplicationException> createProject(String title, String objective) {
     ProjectTitle projectTitle;
+    try {
+      projectTitle = ProjectTitle.create(title);
+    } catch (RuntimeException e) {
+      log.error(e.getMessage(), e);
+      return Result.failure(new ProjectManagementException(ErrorCode.INVALID_PROJECT_TITLE,
+          ErrorParameters.of(200, title)));
```
sven1103, line 43: missing the exception handling for other Throwables here

sven1103, line 43: Not sure I got the meaning of the code `200`

KochTobi: This is not a code but the max length of project titles (stubbed, needs to come from the title once implemented)

KochTobi: I think all throwables are not handled. Checked exceptions are not handled as they should not occur and are truly exceptional.
Maybe I misunderstood you here?

sven1103 (dismissed):
LGTM, great work. 

I would just like to know the meaning of the error code, which i seem to have missed

KochTobi:
@sven1103 The error code is used to decide which error message to present the user with. It uses a MessageProvider to read from a properties file containing the error messages.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/MessageSource.html

sven1103 (approved):
Nice !

PR #97, opened 2022-09-23
Fixes Subscription service provider 

sven1103:
After the removal of Java modules support, the Subscription provider implementation broke. 

This enables the findability of the service via the legacy `META-INF/services` approach to expose service providers.

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/Application.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/Application.java
@@ -82,6 +82,7 @@ public class Application extends SpringBootServletInitializer implements AppShel
     messageBus.subscribe(whenPasswordResetRequestSendEmail(appContext), PASSWORD_RESET);
 
     setupUseCases(appContext);
+    logger.error("Oh oh");
   }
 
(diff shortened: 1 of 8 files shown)


KochTobi (commented):
I do not understand the problem this PR fixes. Could you elaborate?

vaadinfrontend/src/main/java/life/qbic/datamanager/Application.java
```diff
@@ -82,6 +82,7 @@ public static void main(String[] args) {
     messageBus.subscribe(whenPasswordResetRequestSendEmail(appContext), PASSWORD_RESET);
 
     setupUseCases(appContext);
+    logger.error("Oh oh");
```
Steffengreiner, line 85: I think this is a remnant from testing? :) 
```suggestion
```

Steffengreiner (changes requested):
Nicely done! The PW Reset and account creation email sending works as intended. 
I would suggest removing the testing logging remnant and merging development into the branch then this can be merged 👍

Steffengreiner (approved):
Ship it! 🚢

PR #108, opened 2022-10-04
Restructure components

sven1103:
Removes the preview feature from the domain repo and puts it into an
application service.


KochTobi (approved):
LGTM

PR #112, opened 2022-10-05
Format modification time according to user timezone

KochTobi:
This PR tries to query the client details for displaying the date in the zone of the client. It falls back to UTC

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/ClientDetailsProvideImpl.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/ClientDetailsProvideImpl.java
@@ -1,0 +1,27 @@
+package life.qbic.datamanager;
+
+import com.vaadin.flow.component.page.ExtendedClientDetails;
+import com.vaadin.flow.component.page.Page.ExtendedClientDetailsReceiver;
+import com.vaadin.flow.spring.annotation.SpringComponent;
+import java.util.Optional;
+
+/**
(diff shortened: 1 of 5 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/ClientDetailsProvideImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import com.vaadin.flow.component.page.ExtendedClientDetails;
+import com.vaadin.flow.component.page.Page.ExtendedClientDetailsReceiver;
+import com.vaadin.flow.spring.annotation.SpringComponent;
+import java.util.Optional;
+
+/**
+ * Receives client details from vaadin and provides them internally.
+ */
+@SpringComponent
+public class ClientDetailsProvideImpl implements ExtendedClientDetailsReceiver,
```
wow-such-code, line 12: Should be named "Provider" instead of "Provide", no?

wow-such-code:
Please provide a description for this PR, the changes seem more complex than the title suggests.

vaadinfrontend/src/main/java/life/qbic/datamanager/ClientDetailsProvideImpl.java
KochTobi: yes

wow-such-code (approved):
thanks, lgtm!

PR #117, opened 2022-10-06
Specify overflow property to avoid vertical and horizontal overflow in CardLayout

Steffengreiner:
**What was changed** 
Specifies the overflow property in the card-layout.css file to auto to avoid vertical and horizontal overflow. 

**Technical details**
The overflow css property specifies what should happen if a layout is too small to contain its content. 
It's default behaviour is set to visible(See [here](https://www.w3schools.com/csS/css_overflow.asp) for details), which leads to showing the content in the textfield outside of the cardlayout. 
Setting it to auto keeps the content within the window and makes a scrollbar appear if layout is too small for the content within.

### DIFF
--- a/authentication/pom.xml
+++ b/authentication/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>datamanager</artifactId>
     <groupId>life.qbic</groupId>
-    <version>0.7.0</version>
+    <version>0.7.1</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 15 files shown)


jenniferboedker (commented):
Very nice!

Question: can we add spacing so that if the window is resized the text has a fixed spacing in the card? I mean the spacing especially on the left and right border. Additionally it would be cool to have it on the top and bottom as well

Screenshot 2022-10-07 at 08 28 37

jenniferboedker (changes requested):
I would require the fixed spacing on the bottom and top since it appeared with this fix and would appreciate if we can tackle the same issue for the left and right side in this PR as well :)

Steffengreiner:
Heya @jenniferboedker, I added a small padding on the bottom of the formlayout which will lead to the scrollbar be triggered before the formlayout is reached. As we discussed maybe a complete overhaul of the layout structure within the cardlayout makes sense but would be outside of the scope of this PR

jenniferboedker (approved):

PR #118, opened 2022-10-06
Load mail.properties from resources

sven1103:
Fixes the app not being able to read from the mail.properties file when
compiled and package to a running jar file.


jenniferboedker (approved):
Good catch!

PR #122, opened 2022-10-10
Introduce project code

sven1103:


projectmanagement/src/test/java/life/qbic/projectmanagement/domain/project/ProjectSpec.groovy
```diff
@@ -19,8 +19,8 @@ class ProjectSpec extends Specification {
     ProjectIntent intentTwo = ProjectIntent.of(new ProjectTitle("Another project"), new ProjectObjective("another objective"))
 
     expect: "projects with the same uuid are equal"
-    def projectOne = Project.of(ProjectId.of(uuid), intentOne)
-    def projectTwo = Project.of(ProjectId.of(uuid), intentTwo)
+    def projectOne = Project.of(ProjectId.of(uuid), intentOne, ProjectCode.random())
+    def projectTwo = Project.of(ProjectId.of(uuid), intentTwo, ProjectCode.random())
```
jenniferboedker, line 23: should this be true? I think if the code is different the projects are different?

sven1103: The invariant is possible, though we don't want it to be happening in the domain. So the domain service that creates the project entity has the responsibility imo.

jenniferboedker (approved):
Works neatly 🎉

PR #139, opened 2022-10-21
Adjust default sorting of projects

KochTobi:
This PR adjusts the default sorting of projects to show most recently changed projects first. 
It furthermore provides sorting functionality to the grid.


mariadb-connector/src/main/java/life/qbic/persistence/OffsetBasedRequest.java
```diff
@@ -52,11 +66,14 @@ public long getOffset() {
   }
 
   @Override
+  @NonNull
   public Sort getSort() {
+    Objects.requireNonNull(sort);
```
Steffengreiner, line 71: You could also think about moving the @Nonnull annotation to the sort field value with a description such as:
 @NotNull(message = "Sorting may not be null")

projectmanagement/src/main/java/life/qbic/projectmanagement/application/SortOrder.java
```diff
@@ -0,0 +1,25 @@
...
+    return new SortOrder(propertyName, false);
+  }
+
+  public SortOrder descending() {
+    return new SortOrder(propertyName, true);
+  }
+
+  public static SortOrder of(String propertyName) {
+    return new SortOrder(propertyName, true);
+  }
```
Steffengreiner, line 24: This will make the sortorders default value to descending if i understood it correctly. Would it make more sense to also requiring specifying the sort order when creating a new instance?

Steffengreiner (approved):
Heya @KochTobi LGTM i have some questions concerning the nonnull annotation and the default ordering but it works as intended 👍

mariadb-connector/src/main/java/life/qbic/persistence/OffsetBasedRequest.java
KochTobi: TBH. the sort can never be null as it is set in all constructors, so I would not expect a null pointer here 😄 
I added the annotations because IntelliJ hinted me to the fact that they are expected by the interface method.

projectmanagement/src/main/java/life/qbic/projectmanagement/application/SortOrder.java
KochTobi: You can specify both over the constructor. For simplicity I provide this method here so that you do not need to decide upon creation on the direction of the sorting. The `descending()` and `ascending()` methods allow you to modify that later on. If you already know both, you can use the constructor `SortOrder(String propertyName, boolean isDescending)`

PR #140, opened 2022-10-24
Reset functionality also resets textfield state

Steffengreiner:
Addresses [DM-572 ](https://qbicsoftware.atlassian.net/browse/DM-572) by allowing the Reset functionality to also reset textfield state (Remove invalid marker and highlighting) in the projectinformation dialogue window. 
Additionally the reset is triggered independent on how the dialogue window is closed

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
@@ -67,7 +67,6 @@ public class ProjectInformationDialog extends Dialog {
     cancelButton = new Button("Cancel");
 
     configureDialogLayout();
-    configureDialogClosure();
     initForm();
     styleForm();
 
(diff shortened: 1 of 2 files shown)


KochTobi (changes requested):
I like the way you reset the dialog! This PR breaks existing behavior. To avoid that in the future, I suggest encapsulating the closing behavior in the component itself and not making the cancel button visible to the outside. (There is no need for this button to be visible IMHO)

KochTobi (approved):
LGTM
please consider running optimize imports before commiting.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
```diff
@@ -123,23 +123,15 @@ public String getObjective() {
     return projectObjective.getValue();
   }
 
-  public String getExperimentalDesign() {
-    return experimentalDesignField.getValue();
+  public String getExperimentalDesign() { return experimentalDesignField.getValue(); }
```
KochTobi, line 126: What happened here?
```suggestion
public String getExperimentalDesign() {
 return experimentalDesignField.getValue(); 
}
```
Please do not remove the new-line characters. Your IDE will automatically display it in one line but our code style expects the body in a new line.

KochTobi (changes requested):

KochTobi (approved):

PR #154, opened 2022-12-05
Make experimental design required during project creation and update

Steffengreiner:
**What was changed** 
Removes the optional wrapping of the experimentaldesigndescription in the domain and application logic since it's a mandatory field now. 

**Additional changes**
Also updated the database by filling in the empty experimental design fields for all currently created projects. 

**More Information**
Addresses DM-635

### DIFF
--- a/vaadinfrontend/src/main/resources/messages.properties
+++ b/vaadinfrontend/src/main/resources/messages.properties
@@ -1,4 +1,4 @@
 GENERAL=Ooooooops! Sorry! -> Please contact support@qbic.zendesk.com.
 INVALID_PROJECT_TITLE=The project title {1} is invalid. -> A valid project title is not empty and is shorter than {0} characters.
-INVALID_EXPERIMENTAL_DESIGN=The experimental design description is too long. -> A description of an experimental design must contain at most {0} characters.
+INVALID_EXPERIMENTAL_DESIGN=The experimental design description is invalid -> A vaild description of an experimental design must not be empty and contain at most {0} characters.
 INVALID_PROJECT_OBJECTIVE=The project objective {0} is invalid. -> A valid objective is not empty.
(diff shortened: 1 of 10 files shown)


vaadinfrontend/src/main/resources/messages.properties
```diff
@@ -1,4 +1,4 @@
 GENERAL=Ooooooops! Sorry! -> Please contact support@qbic.zendesk.com.
 INVALID_PROJECT_TITLE=The project title {1} is invalid. -> A valid project title is not empty and is shorter than {0} characters.
-INVALID_EXPERIMENTAL_DESIGN=The experimental design description is too long. -> A description of an experimental design must contain at most {0} characters.
+INVALID_EXPERIMENTAL_DESIGN=The experimental design description is invalid -> A vaild description of an experimental design must not be empty and contain at most {0} characters.
```
jenniferboedker, line 3:
```suggestion
INVALID_EXPERIMENTAL_DESIGN=The experimental design description is invalid -> A vaild description of an experimental design must not be empty and contain at least the number of samples.
```

jenniferboedker (changes requested):
Hi Steffen, thanks for the work 💪🏼 
Looks very good and works!

Is it possible to keep the dialog open and show the error in the component. Otherwise this leads to bad UX since I have to click again to open the dialog. It kind of breaks the user flow

jenniferboedker:
I just noticed that holds true for all fields but did not catch my attention. If its a quick fix it would be awesome to fix it for all required fields so the user is supported in creating projects fast

vaadinfrontend/src/main/resources/messages.properties
Steffengreiner: @jenniferboedker I think this is the description of the experimental design ergo meaning the textual explanation what is to happen with the samples. 
Currently we inform the user if that description is over a specific amount of characters. 
Do we change the definition of the experimental design to only account for the number of samples?

KochTobi: ? what do you mean with that @jenniferboedker ?

jenniferboedker (approved):

PR #155, opened 2022-12-12
Reset ProjectDialogue after successful project Creation

Steffengreiner:
**What was changed**
Resets the project creation dialogue if the project was created successfully 👍


Steffengreiner:
Since the reset is executed in the ifSuccessorElse method the fields will only be reset for a successful project creation

KochTobi:
> Since the reset is executed in the ifSuccessorElse method the fields will only be reset for a successful project creation

Did you try that out? (e.g. not providing any experimental design)

Steffengreiner:
https://user-images.githubusercontent.com/29627977/207038086-0099e699-7aac-4d16-867c-e4459543ea03.mov
I tried it with all the fields(shown here) and comboboxes(not shown here)

KochTobi (approved):

PR #162, opened 2023-01-20
add openbis connector and functionality to fetch vocabularies

wow-such-code:


projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/vocabulary/ControlledVocabulary.java
```diff
@@ -0,0 +1,39 @@
...
+  final private String identifier;
+
+  private Map<String,String> vocabularyTermsByName;
+
+  public String getIdentifier() {
+    return identifier;
+  }
+
+  public List<String> getVocabularyTermNames() {
+    return new ArrayList<>(vocabularyTermsByName.keySet());
```
Steffengreiner, line 29: Question: I'd expect the map ordering to be switched with the database identifier as key and the name as value, since it's the value behind the identifiers that is of interest of us?

openbis-connector/src/main/java/life/qbic/projectmanagement/persistence/repository/OpenbisConnector.java
```diff
@@ -0,0 +1,86 @@
...
+    criteria.withVocabulary().withCode().thatEquals(vocabularyCode);
+
+    VocabularyTermFetchOptions options = new VocabularyTermFetchOptions();
+    SearchResult<VocabularyTerm> searchResult =
+        openBisClient.getV3().searchVocabularyTerms(openBisClient.getSessionToken(), criteria, options);
+
+    Map<String,String> termsByName = new HashMap<>();
+    for (VocabularyTerm term : searchResult.getObjects()) {
+      String code = term.getCode();
+      String label = code;
```
Steffengreiner, line 46: Why did you assign the code to the label here?

openbis-connector/src/main/java/life/qbic/projectmanagement/persistence/repository/OpenbisConnector.java
```diff
@@ -0,0 +1,86 @@
...
+      if (term.getLabel() != null && !term.getLabel().isEmpty()) {
+        label = term.getLabel();
```
Steffengreiner, line 48: Question for clarification: Is the label the value of interest for us? If so then i'd be nitty here and keep the naming consistent between the ControlledVocabulary terminology and this method.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/vocabulary/ControlledVocabulary.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public ControlledVocabulary(String identifier,
+      Map<String,String> vocabularyTermsByName) {
+    this.identifier = identifier;
+    this.vocabularyTermsByName = vocabularyTermsByName;
+  }
+
+  final private String identifier;
+
+  private Map<String,String> vocabularyTermsByName;
```
Steffengreiner, line 22: NIT: This could also be called vocabularyEntries

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/vocabulary/ControlledVocabulary.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public String getIdentifier() {
+    return identifier;
+  }
+
+  public List<String> getVocabularyTermNames() {
+    return new ArrayList<>(vocabularyTermsByName.keySet());
+  }
+
+  public String getVocabularyTermCode(String name) {
```
Steffengreiner, line 32: NIT: I'd keep the naming consistent between identifier or code(I'd argue identifier is more explicit)

Steffengreiner (changes requested):
Heya @wow-such-code ! Very nice implementation. 
I have some questions and found nitpicks about the used terms but nothing major 👍

openbis-connector/src/main/java/life/qbic/projectmanagement/persistence/repository/OpenbisConnector.java
wow-such-code: Vocabulary terms are stored in code/label pairs, but labels are optional. So we return the code, in case there is no label.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/vocabulary/ControlledVocabulary.java
wow-such-code: That's fair. I did it this way because I re-use it to map it back to the code later. (method that is not used atm)

PR #167, opened 2023-01-30
Feature/dm 677/add empty project

wow-such-code:
draft to create an empty project in openbis and check for existing projects with that code. wiring needs work


projectmanagement/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
```diff
@@ -26,9 +28,11 @@ public class ProjectCreationService {
   private static final Logger log = logger(ProjectCreationService.class);
 
   private final ProjectRepository projectRepository;
+  private final ProjectDataRepository projectDataRepository;
 
-  public ProjectCreationService(ProjectRepository projectRepository) {
+  public ProjectCreationService(ProjectRepository projectRepository, @Autowired ProjectDataRepository projectDataRepository) {
```
jenniferboedker, line 33: Still open to discuss if we use autowired or not. Maybe @KochTobi or @sven1103 can state what is best practice here

projectmanagement/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
```diff
@@ -61,7 +66,7 @@ public Result<Project, ApplicationException> createProject(String title, String
 
   private ProjectCode createRandomCode() {
     ProjectCode code = ProjectCode.random();
-    while (!projectRepository.find(code).isEmpty()) {
+    while (!projectRepository.find(code).isEmpty() || projectDataRepository.projectExists(code)) {
```
jenniferboedker, line 69: Do we want to only check openbis? Should they be out of sync? Can there be projects that are in the project repostiory but not stored in openbis? 

I have the feeling only one condition is enough

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/repository/ProjectDataRepository.java
```diff
@@ -0,0 +1,33 @@
...
+ * <b>Project Data Storage Interface</b>
+ *
+ * <p>Provides access to the persistence layer that handles the {@link Project} data storage.
+ *
+ * @since 1.0.0
+ */
+public interface ProjectDataRepository {
+
+  /**
+   * Saves a {@link Project} entity permanently.
```
jenniferboedker, line 17: Does it save a project? It rather creates space to store data for a project of a specific code, right?

projectmanagement/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
wow-such-code: Unless we change project creation in the openbis connector this will lead to an error, if a project exists there. They should not be out of sync, but I think it's important to check. Maybe also concerning old projects?

jenniferboedker: hm ok, I actually thought we added the check in the project repository only because we did not have the openbis implementation at that time.

But we also dont lose anything to keep it like this. Wont object.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/repository/ProjectDataRepository.java
```diff
@@ -14,17 +14,17 @@
 public interface ProjectDataRepository {
 
   /**
-   * Saves a {@link Project} entity permanently.
+   * Creates a reference to a {@link Project} permanently, in order to connect project data.
```
jenniferboedker, line 17:
```suggestion
   * Creates a reference to a {@link Project} in the data repository to connect project data.
```

KochTobi (approved):
LGTM

PR #172, opened 2023-02-23
add project code field and auto-generate button to create codes.

wow-such-code:
- adds auto-generation button and the possibility to input own project code in project creation
- adds error/missing data handling to the project code field
- cleans up validation of project code, which used conflicting information in different methods
- adds required sign for experimental design input field, as the logic requires it at the moment

### DIFF
--- a/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/ProjectCode.java
+++ b/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/ProjectCode.java
@@ -58,7 +58,9 @@ public class ProjectCode {
    * @since 1.0.0
    */
   public static ProjectCode parse(String str) throws IllegalArgumentException {
-    if (isInvalid(str)) {
+    // project codes are always upper case
+    str = str.toUpperCase();
+    if (!(str.startsWith(PREFIX) && (str.length() == LENGTH))) {
       throw new IllegalArgumentException(String.format("%s is not a valid project code", str));
     }
     if (containsInvalidCharacters(str)) {
@@ -72,7 +74,7 @@ public class ProjectCode {
   }
 
   private static boolean isValid(String code) {
-    return code.startsWith(PREFIX) && (code.length() == LENGTH);
+    return code.startsWith(PREFIX) && (code.length() == LENGTH) && !containsInvalidCharacters(code);
   }
 
   private static boolean containsInvalidCharacters(String code) {
@@ -122,10 +124,6 @@ public class ProjectCode {
       }
     }
     return false;
-  }
-
-  private static boolean isInvalid(String code) {
-    return !isValid(code);
   }
 
   private static boolean isBlackListed(String word) {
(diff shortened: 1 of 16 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
```diff
@@ -181,6 +207,12 @@ public void reset() {
         .forEach(component -> ((HasValue<?, ?>) component).clear());
     formLayout.getChildren().filter(component -> component instanceof HasValidation)
         .forEach(component -> ((HasValidation) component).setInvalid(false));
+
+    // reset components in sub-layout
+    codeAndTitleLayout.getChildren().filter(component -> component instanceof HasValue<?, ?>)
+        .forEach(component -> ((HasValue<?, ?>) component).clear());
+    codeAndTitleLayout.getChildren().filter(component -> component instanceof HasValidation)
+        .forEach(component -> ((HasValidation) component).setInvalid(false));
```
jenniferboedker, line 215: NIT since this is repetitive code maybe a function can facilitate the reset of (sub)components

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/ProjectCode.java
```diff
@@ -72,7 +74,7 @@ public static ProjectCode parse(String str) throws IllegalArgumentException {
   }
 
   private static boolean isValid(String code) {
-    return code.startsWith(PREFIX) && (code.length() == LENGTH);
+    return code.startsWith(PREFIX) && (code.length() == LENGTH) && !containsInvalidCharacters(code);
```
jenniferboedker, line 77: Why do we need to check for invalid characters here as well? In line 66 there is this check already done :)

jenniferboedker (changes requested):
Left some questions, otherwise looks good! Thanks for working on this 🎉

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/ProjectCode.java
wow-such-code: Good point, this method was used differently before - now it is unused and parse(str) takes over all validation. It would save some space, but I think the name is wrong if it does not check for all statements that make a code illegal. Would you agree that we can remove the method?

jenniferboedker: the isValid method?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
wow-such-code: good idea, changed

jenniferboedker (approved):

PR #173, opened 2023-02-24
Catch and describe openBIS connection issues verbose in logs

sven1103:


wow-such-code (approved):
looks like a good solution, thanks

PR #174, opened 2023-02-27
Set projects overview as root route

sven1103:
A parent layout should not be the root route, since tthis will show the navigation page after login in development mode, and not work at all in production mode (route cannot be found!).

Every Vaadin App should have a defined root route, that represents the default entry page for a user.


jenniferboedker (approved):
Looks good and works 👍🏼

PR #175, opened 2023-02-27
Centralize app routes in global constant class

sven1103:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/AppRoutes.java
```diff
@@ -1,24 +1,44 @@
...
 public class AppRoutes {
 
   /**
-   * The "Main" page of the data manager app. Location, where you want the user to be
-   * directed to, if no explicit path has been provided.
+   * The "Main" page of the data manager app. Location, where you want the user to be directed to,
+   * if no explicit path has been provided. Shows all the projects a user has access to.
    */
-  public static final String MAIN = "";
+  public static final String PROJECTS = "";
```
Steffengreiner, line 17: I'd argue that MAIN, HOME or ROOT was actually correct here since it's intention is to be the route for directing the user to the current landing page.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/overview/ProjectOverviewPage.java
```diff
@@ -20,8 +20,8 @@
  * @since 1.0.0
  */
 @PageTitle("Project Overview")
-@Route(value = AppRoutes.MAIN, layout = MainLayout.class)
-@RouteAlias(value = AppRoutes.MAIN_ALIAS)
+@Route(value = AppRoutes.PROJECTS, layout = MainLayout.class)
+@RouteAlias(value = AppRoutes.PROJECTS_ALIAS)
```
Steffengreiner, line 24: Would it make sense to also have additional aliases similiar to "home" or "main"? 
https://vaadin.com/docs/next/routing/route

vaadinfrontend/src/main/java/life/qbic/datamanager/views/AppRoutes.java
```diff
@@ -1,24 +1,44 @@
...
- * shall be visible by all components within the app.
- *
+ * Defines the data manager's public application routes, that shall be visible by all components
+ * within the app.
+ * <p>
  * Local routes of components should be listed here with care!
  *
  * @since 1.0.0
  */
 public class AppRoutes {
```
Steffengreiner, line 11: What is the reason for not defining the AppRoutes as an Enum?

Steffengreiner (changes requested):
Heya @sven1103! Nicely done, i have some minor questions and will approve once they are answered :)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/AppRoutes.java
sven1103: Why enums? we need the character sequence representation for the route annotation value property

vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/overview/ProjectOverviewPage.java
Steffengreiner: I understand but would argue that in the context of an application redirecting the user back to the main-page is a sensible operation. 
Without a "home" or "main" alias the knowledge that the ProjectOverviewPage functions as the landing page is implictit rather than explicit. 
However this is more of a NIT than a blocking issue so i'd approve the PR :)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/AppRoutes.java
Steffengreiner: Also more of a NIT but: 
The purpose of an enum is to make sure only the defined values are to be used, which makes sense in this context since only these values will provide sensible routing. 
Additionally by combining the Enum Constant with a String value similiar to [previous implentations](https://github.com/qbicsoftware/data-model-lib/blob/5ddc9812bb2445d3b22f1bbf985fc90e40f5fd18/src/main/groovy/life/qbic/datamodel/dtos/business/AcademicTitle.groovy) you'd still get the same result.

Steffengreiner (approved):
I still have some NITS, but agree that they should not block this PR 👍

Steffengreiner (approved):
LGTM 👍

PR #176, opened 2023-02-27
Provide javadoc for ToggleDisplayEditComponent

Steffengreiner:
**What was changed**
Add javadocs to provide more insights into the inner workings of the ToggleDisplayEditComponent


sven1103 (changes requested):
Some suggestion for readability and navigation

sven1103 (approved):

PR #183, opened 2023-03-20
Fix Format in project creation dialog

jenniferboedker:
This PR aims at cleaning up the Create Project Dialog window.
Changes made:

- fix length: make various components to have the same lenght
- Add Project Contact Heading: allow visual separation of experiment
- Change Experimantal Design to Experiment
- Move experimental design description in experiment section
- move code next to offer ID (title is important and otherwise might be overseen)
- make code to be autogenerated automatically (to make project creation faster)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/project/create/ProjectInformationDialog.java
```diff
@@ -216,6 +246,13 @@ public void reset() {
   private void resetChildValues(Component component) {
     component.getChildren().filter(comp -> comp instanceof HasValue<?, ?>)
         .forEach(comp -> ((HasValue<?, ?>) comp).clear());
+
+  }
+
+  private void defaultProjectCodeCreation() {
+    this.addOpenedChangeListener(opend -> {
+      codeField.setValue(ProjectCode.random().value());
+    });
```
KochTobi, line 255: The opened change listener also fires on close.
```suggestion
  private void defaultProjectCodeCreation() {
    this.addOpenedChangeListener(opend -> 
    {
    if opened is window.opened // or similar
      codeField.setValue(ProjectCode.random().value());
    });
```

PR #187, opened 2023-03-22
Fix project information component layout

jenniferboedker:
Attempts to fix discrepancy between prototype and implementation:

- Adds 'Projects' title to list of projects (as unrelated quick fix)
- Reorganise the layouts of the project information view (remove unnecessary lables, move labels above fields)
- Add project title to page (only a dummy as of now)

Todo:
Adjust project title to the actual title and code


sven1103:
Closing, since this PR has stalled.

PR #192, opened 2023-03-30
Introduce experimental groups in domain and persistence layer

wow-such-code:
experimental groups*


projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/Condition.java
```diff
@@ -121,47 +99,25 @@ public ConditionLabel label() {
    * @return true if <code>other</code> has the same levels defined; false otherwise
    */
   public boolean hasSameLevelsDefined(Condition other) {
```
sven1103, line 101: The function does now exactly the same as the overridden equals method. So redundant and can be removed.

sven1103 (changes requested):
Thank you for your contribution! Only two minor things, else looks good to me :)

sven1103 (approved):

PR #193, opened 2023-03-31
Increase number of iterations to 100k for PBKDF2

sven1103:
Since we do not enforce passwords of length greater 12 chars and special chars, lets at least increase the number of iterations from 10k to 100k :)


KochTobi (approved):
LGTM

PR #196, opened 2023-04-03
add experimental group handling to experiment information service

wow-such-code:


projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/ExperimentalVariable.java
```diff
@@ -85,9 +92,10 @@ Result<VariableLevel, Exception> addLevel(ExperimentalValue experimentalValue) {
   }
 
   private boolean hasDifferentUnitAsExistingLevels(ExperimentalValue experimentalValue) {
-    Predicate<ExperimentalValue> hasDifferentUnit = input -> !input.unit()
-        .equals(experimentalValue.unit());
-    return levels.stream().anyMatch(hasDifferentUnit);
+    if(levels.isEmpty())
```
KochTobi, line 95: dangerous. if there is already a level with a unit you would give your ok here although the experimental value has a different unit (being no unit)

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/ExperimentalVariable.java
```diff
@@ -53,7 +54,13 @@ private ExperimentalVariable(String name, ExperimentalValue... levels) {
...
     }
     this.name = VariableName.create(name);
-    this.levels.addAll(List.of(levels));
+    // necessary to test unit cohesion
+    for(ExperimentalValue level : levels) {
+      Result<VariableLevel, Exception> addLevelResult = addLevel(level);//TODO this does not seem like the correct way to use this
+      if (addLevelResult.isFailure()) {
+          throw new IllegalArgumentException(addLevelResult.exception().getMessage());
+      }
+    }
```
KochTobi, line 63: Why not simply check for the unit and throw the illegal argument exception?

KochTobi (changes requested):

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/ExperimentalVariable.java
wow-such-code: levels.isEmpty() does not check the unit, it checks if there **are** already levels at all. the first level that is added is always okay.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/ExperimentalVariable.java
wow-such-code: because that would be some code duplication. Might be better than before, though...

KochTobi (approved):

PR #207, opened 2023-04-19
Update data-manager-application to Vaadin 24 and Spring Boot 3

Steffengreiner:
**What was changed**
Update Vaadin version to current version 24.04 which required an update of the spring boot version to 3.X.X. 
Detailed information can be found here in the [docs](https://vaadin.com/docs/latest/upgrading)

This has opened the wormhole to update all dependencies associated with spring such as: 

1.) Replace javax package with jakarta package
2.) Bump Maven wrapper version 
3.) Update groovy-all version and spock version to fix tests 
4.) Update logging to slfj to 2.0.7 and logback classic to 1.46 and fix pattern in logback.xml file to fix logging 
5.) Explictly set openbis-api version to 19.06 since the previous version used a removed spring utils method. 

Additionally minor css adaptions had to be taken to adhere to changes in the vaadin css: 

1.) Multiselect-combo-box chips are now a seperate entity within the multiselect combobox(can't be styled with ::part within the shadowdom anymore) 
2.) H2 Element Margins had to be reintroduced since they were removed during the update


vaadinfrontend/pom.xml
```diff
@@ -188,7 +188,6 @@
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <version>1.2.11</version>
```
wow-such-code, line 191: why was this version specified before but now removed?

pom.xml
```diff
@@ -124,16 +124,9 @@
       <dependency>
         <groupId>org.spockframework</groupId>
         <artifactId>spock-core</artifactId>
-        <version>2.2-M3-groovy-4.0</version>
+        <version>2.4-M1-groovy-4.0</version>
         <scope>test</scope>
       </dependency>
-      <!--If not specified, the 2.7.6 spring-boot-starter-parent dependency will introduce transitively the 1.30 version of snakeyaml, for which a multitude of security concerns are reported
```
wow-such-code, line 130: Does spring-boot 3 fix this problem?

projectmanagement/pom.xml
```diff
@@ -10,15 +10,6 @@
...
   <url>http://maven.apache.org</url>
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>javax.persistence</groupId>
-        <artifactId>javax.persistence-api</artifactId>
-        <version>2.2</version>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
```
wow-such-code, line 21: why doesn't this have to be defined in the new version?

wow-such-code (changes requested):
looks good, but I have a few questions

projectmanagement/pom.xml
wow-such-code: is there no import needed for jakarta.persistence-api or similar?

pom.xml
Steffengreiner: Thanks to your review i doublechecked the dependency tree and I was wrong to assume that this was addressed. [Apparently](https://stackoverflow.com/questions/75870282/is-snakeyaml-2-0-added-in-the-new-spring-boot-versions) spring boot stayed with 1.33 to avoid dependency issues. 
Therefore i readded this dependency 👍

wow-such-code (approved):
nice

PR #208, opened 2023-04-21
User is able to create additional experiments within a project

Steffengreiner:
**What was changed**
1.) Adapt routing to experimentpage to include experimentId with rerouting to active experiment if id is invalid. 
2.) Include possibility to add additional experiments to a project via experimentList component 
3.) Switch to newly created experiment after experiment creation
4.) Switch between experiments via clicking on an experimentcard within experimentList component


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/ProjectInformationDialog.java
wow-such-code, line 1: file is empty, I think you can delete it

wow-such-code:
everything seems to work very well, but I am not finished with the review, yet.

a small point:
rollover pointer for experiment switching is missing

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/ProjectInformationDialog.java
Steffengreiner: Nice catch 👀

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/ProjectCreationDialog.java
```diff
@@ -0,0 +1,375 @@
...
+  private final Button cancelButton = new Button("Cancel");
+  private final Handler handler;
+
+  public ProjectCreationDialog(ExperimentalDesignSearchService experimentalDesignSearchService) {
+
+    initCodeAndTitleLayout();
+    projectObjective.setRequired(true);
+
+    defineExperimentComponent = new DefineExperimentComponent(experimentalDesignSearchService);
+    defineExperimentComponent.hideNameField();
```
wow-such-code, line 76: why hide it?

wow-such-code: does the first experiment have no name?

Steffengreiner: Yes, that's the actual reason for the hide method in the first place (which also looks contrived to me and i'm very happy to hear an alternate solution). For now we don't want the user to have the possibility an experiment name during project creation according to the prototype

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/ProjectCreationDialog.java
```diff
@@ -0,0 +1,375 @@
...
+    responsiblePerson.setClearButtonVisible(true);
+
+    projectManager.setPlaceholder("Select a project manager");
+    projectManager.setRequired(true);
+
+    configureDialogLayout();
+    initForm();
+    styleForm();
+    handler = new Handler();
+    handler.handle();
```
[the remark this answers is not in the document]
Steffengreiner: Tbh the handle method is unnecessary anyway imo. We stopped defining it in the newer components and therefore i'll remove it here. Nice catch 👀

PR #217, opened 2023-04-26
Feature/dm 750/update experimental group card design

wow-such-code:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupsLayout.java
```diff
@@ -114,31 +135,28 @@ private void fillWithVariableLevels(FlexLayout tagsContainer, VariableLevel[] va
           .sorted((VariableLevel l1, VariableLevel l2) -> l1.variableName().value()
               .compareToIgnoreCase(l2.variableName().value())).toArray(VariableLevel[]::new);
       for (VariableLevel variableLevel : variableLevels) {
+        String formattedValue = ExperimentValueFormatter.format(
+            variableLevel.experimentalValue());
         Tag tag = new Tag(
-            variableLevel.variableName().value() + ":" + ExperimentValueFormatter.format(
-                variableLevel.experimentalValue()));
+            variableLevel.variableName().value() + ":" + formattedValue);
+        tag.addClassNames(TextOverflow.ELLIPSIS);//this does not seem to work, any ideas?
```
KochTobi, line 142: Is this something that should be adjusted in every tag? See the tag class for general adjustment.

wow-such-code: I'm not sure where to set the ELLIPSIS option (meaning cut off values end with "..." instead of a hard cut-off), but the cut off itself is set in the parent container. I don't think this is needed for all tags. In fact @Steffengreiner suggested maybe not using the tags for these groups at all, since it is hard to set tooltips for them. What do you think?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupsLayout.java
```diff
@@ -84,12 +89,30 @@ public interface ExperimentalGroupCommandListener {
...
+    private void setCardLayoutStyle() {
+      addClassNames(
+          "rounded-m",
+          "box-border",
+          "rounded-m",
+          "shadow-xs",
+          "p-m"
+      );
+    }
+  }
```
Steffengreiner, line 113: Why was this moved into a seperate Class? The Width and Height is hard set so it can't be reused in other classes. An Idea would be to allow the user to set the width and height so it can be used in other components?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupsLayout.java
```diff
@@ -98,14 +121,12 @@ public ExperimentalGroupCard(int sampleSize, VariableLevel... variableLevels) {
       tagsContainer.setFlexWrap(FlexWrap.WRAP);
       tagsContainer.setFlexDirection(FlexDirection.ROW);
       tagsContainer.setAlignContent(ContentAlignment.START);
+      tagsContainer.addClassNames(Overflow.HIDDEN);
+      tagsContainer.setWidth("165px");//TODO better way to hide overflow?
```
Steffengreiner: This should be set to use the full width of the card container 
```suggestion
      tagsContainer.setWidthFull()
```

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupsLayout.java
Steffengreiner: I agree with @KochTobi that this logic can be moved to the Tag itself since we want to allow this behavior everywhere

Steffengreiner (changes requested):
Heya @wow-such-code I made some changes to the Tag CSS layout to ensure that the width stays as intended. Feel Free to adapt it to your needs. 
I'll approve this PR as soon as I understand the need for a BaseExperimentalGroupCard. 
I'd argue it makes more sense to provide Cardlayout based Components for 2 cases:

1.) Cards styled to be used within a page component
2.) A Page component styled as a card 

The BaseExperimentalGroupCard does not fulfill this functionality since it has hardcoded needs designed specifically for the ExperimentalGroupCard imo. 

Let me know what you think.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupsLayout.java
wow-such-code: the new class is used because the card with groups and the card with the "add group" button were using a lot of the same code. I don't foresee a third use case for these specific cards but with other sizes at the moment.

PR #222, opened 2023-05-02
Add missing jackson dependency

KochTobi:
For version 2.15.0 jackson-core is required to run.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
     <maven.compiler.target>17</maven.compiler.target>
     <jakarta.persistence.version>3.1.0</jakarta.persistence.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jackson.version>2.15.0</jackson.version>
   </properties>
 
   <parent>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -137,27 +138,26 @@
         <type>pom</type>
         <scope>test</scope>
       </dependency>
-      <!--If not specified, the 2.7.6 spring-boot-starter-parent dependency will introduce transitively the 1.30 version of snakeyaml, for which a multitude of security concerns are reported -->
-      <dependency>
-        <groupId>org.yaml</groupId>
-        <artifactId>snakeyaml</artifactId>
-        <version>2.0</version>
-      </dependency>
```
Steffengreiner, line 145: Why was this removed? 
If the version is not specified then sprnig-boot will use the 1.33 version for which security issues have been reported:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-1471

KochTobi: >2.7.6 spring-boot-starter-parent

KochTobi: we are now using 3.x

Steffengreiner (dismissed):
LGTM 🚀

pom.xml
KochTobi: I was wrong, turns out it is still pulled but now with version 1.33 (which has 1 CVE)

Steffengreiner (approved):
LGTM 🚀

PR #243, opened 2023-05-22
Remove german exception messages as vaadin always uses machine not client locale

KochTobi:
Removes the messages_de.properties file which provides german translations for error messages. As vaadin currently only supplies the HTTPRequest with the server locale, this does not take effect and leads to confusion.


Steffengreiner (approved):
LGMT 🚀

PR #244, opened 2023-05-22
Introduce Basic Card component 

Steffengreiner:
**What was Changed**
Introduce Basic Card component which hosts the styles within its own custom css class.

### DIFF
--- a/vaadinfrontend/frontend/themes/datamanager/card.css
+++ b/vaadinfrontend/frontend/themes/datamanager/card.css
@@ -1,0 +1,27 @@
+.card {
+  display: flex;
+  flex: auto;
+  flex-direction: column;
+  border-color: var(--lumo-contrast-10pct);
+  padding: var(--lumo-space-l);
+  font-size: var(--lumo-font-size-s);
+  box-shadow: var(--lumo-box-shadow-s);
+  border-radius: var(--lumo-border-radius-m);
+  box-sizing: border-box;
+  background-color: var(--lumo-base-color);
+  margin: 5px;
+  gap: 1em;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.card .title {
+  white-space: nowrap;
+  font-weight: bold;
+  font-size: var(--lumo-font-size-m);
+  color: var(--lumo-header-text-color);
+}
+
+.card .content {
+
+}
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
@@ -1,0 +1,45 @@
+package life.qbic.datamanager.views.layouts;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.Composite;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.Span;
+
+/**
+ * <b>Card, A {@link Composite} component containing a {@link Div} hosting the content to be shown
+ * on a card within a {@link PageComponent} in the data-manager-app</b>
+ */
+public class Card extends Composite<Div> {
+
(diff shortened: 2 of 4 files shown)


vaadinfrontend/frontend/themes/datamanager/card.css
```diff
@@ -0,0 +1,27 @@
...
+  border-color: var(--lumo-contrast-10pct);
+  padding: var(--lumo-space-l);
+  font-size: var(--lumo-font-size-s);
+  box-shadow: var(--lumo-box-shadow-s);
+  border-radius: var(--lumo-border-radius-m);
+  box-sizing: border-box;
+  background-color: var(--lumo-base-color);
+  margin: 5px;
+  gap: 1em;
+  overflow: hidden;
```
KochTobi, line 14: Why is this? I thought it is `flex` and it expands until the content is covered.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
```diff
@@ -0,0 +1,45 @@
...
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.Span;
+
+/**
+ * <b>Card, A {@link Composite} component containing a {@link Div} hosting the content to be shown
+ * on a card within a {@link PageComponent} in the data-manager-app</b>
+ */
+public class Card extends Composite<Div> {
+
+  private final Span title = new Span("");
```
KochTobi, line 14: I would argue a simple card is empty.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
```diff
@@ -0,0 +1,45 @@
...
+    add(title);
+    setDefaultCardStyles();
+  }
+
+  public Card(String title, Component... content) {
+    this.title.setText(title);
+    getContent().add(content);
+    add(this.title);
+    setDefaultCardStyles();
+  }
```
KochTobi, line 26: I thought a simple card is empty and provides only border and shadow?

vaadinfrontend/frontend/themes/datamanager/card.css
Steffengreiner: The card is styled as a flexcomponent, however I'd argue we don't want to allow the card to expand beyond the outer container boundaries, which is why we set the overflow to hidden. 

The background color is set as the default color chosen by vaadin and the gap property allows the card to behave spacing wise similiar to a vaadin layout, which is the current style on which we based the data-manager 

I'm more than open to discuss if the margin property is necessary in the card itself, since it's there to allow for space between the cards if multiple cards are aligned next to each other

vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
Steffengreiner: I thought the idea was to enable the user to add specific content to the card which should be automatically styled?
I'm happy to remove the title from this card, however we would then need to decide on how we deal with small changes to the basic card design (e.g. by introducing multiple components for small changes such as titlecard, cardwithbuttons etc...)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/layouts/Card.java
Steffengreiner: I'd have assumed that the Card needs a way to host the content we want to show within the card. This constructor is just a more convienient way to set the title and content during the card creation instead of setting the title and content later. I can remove the constructor if you wish 👍

Steffengreiner:
After our discussion today we agreed on the new component and css structure which can be seen here: https://github.com/qbicsoftware/data-manager-app/pull/245
Therefore this Suggestion can be closed

PR #245, opened 2023-05-23
feature/dm-759-as-a-pm-i-want-to-delete-or-edit-an-experimental-group

sven1103:


projectmanagement/src/main/java/life/qbic/projectmanagement/application/ExperimentInformationService.java
```diff
@@ -60,18 +62,19 @@ private Experiment loadExperimentById(ExperimentId experimentId) {
...
     Objects.requireNonNull(experimentId, "experiment id must not be null");
 
     Experiment activeExperiment = loadExperimentById(experimentId);
-    AddExperimentalGroupResponse response = activeExperiment.addExperimentalGroup(
+    Result<ExperimentalGroup, ResponseCode> result = activeExperiment.addExperimentalGroup(
         experimentalGroup.levels(), experimentalGroup.sampleSize());
-    if (response.responseCode() == ResponseCode.SUCCESS) {
+    if(result.isValue()) {
       experimentRepository.update(activeExperiment);
+      return result;
```
wow-such-code, line 75:
```suggestion
```

wow-such-code: result is returned below the if block anyway.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/experiment/ExperimentalGroup.java
```diff
@@ -3,6 +3,7 @@
 import jakarta.persistence.CascadeType;
 import jakarta.persistence.Entity;
 import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
```
wow-such-code, line 6:
```suggestion
```
unused import afai can see

wow-such-code (changes requested):
Found two problems while trying it out:

- Group cards are added to the view every time the experiment tab is clicked (even when you are already there). This leads to an accumulation of cards, see screenshot:
image
I suspect resetting them when navigating to the experiment tab first, will fix the problem. This problem persists even between projects btw.

- Variable names are not displayed - they used to be in the last iteration of this feature. I think this might be important in case the variable levels are named similarly (e.g. mutant and KO, but two different genes).

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentDetailsComponent.java
```diff
@@ -165,14 +170,30 @@ private void initExperimentalVariableCard(
     experimentalVariableCard.setAddButtonAction(addVariablesDialog::open);
   }
 
-  private void initExperimentalGroupsBoard() {
-    experimentalGroupsLayoutBoard.setWidthFull();
+  public void experimentId(ExperimentId experimentId) {
+    this.handler.setExperimentId(experimentId);
```
wow-such-code, line 174: groups should be reset somewhere here - the "add card" has to be added again.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupCard.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private Div condition() {
+    var variableLevels = experimentalGroup.condition().getVariableLevels();
+    Div layout = new Div();
+    String tagFormat = "%s %s"; // "<value> [<unit>]"
+    List<Tag> tags = variableLevels.stream()
+        .sorted(Comparator.comparing(o -> o.variableName().value()))
+        .map(variableLevel -> new Tag(tagFormat.formatted(variableLevel.experimentalValue().value(),
+            variableLevel.experimentalValue().unit().orElse("").trim()))).toList();
```
wow-such-code, line 86: nice

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentDetailsComponent.java
wow-such-code: also: shouldn't this be called setExperimentId(...)? Or am I out of the loop concerning the naming conventions? :)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/general/Card.java
```diff
@@ -0,0 +1,25 @@
...
+import com.vaadin.flow.component.Tag;
+import java.io.Serial;
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ */
+@Tag(Tag.DIV)
```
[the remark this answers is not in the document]
sven1103: As far as I understood @Steffengreiner , we need the tag to tell Vaadin to render instances of this class as Div in the HTML DOM. 

Not sure what you mean with cards below?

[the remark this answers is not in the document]
wow-such-code: Sorry, I mean the classes inheriting from Card. I'm probably misunderstanding how this works, but it sounds like they ARE a Div (thanks to the tag) but then they also have Div fields?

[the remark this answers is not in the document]
wow-such-code: NIT, obviously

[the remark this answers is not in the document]
sven1103: all inherited classes will be a Div in the DOM

image

wow-such-code:
Nice addition of the card leading back to variable definition if none are defined.
Unfortunately there are still two problems:

- Groups are not shown until adding another group
- Changing projects (back button and selecting another one) shows groups of the previous project

wow-such-code (approved):
works now 👍

PR #252, opened 2023-05-31
Base css styling definitions on the extended java component

Steffengreiner:
**What was changed** 

1.) Introduce dedicated component css stylings ("card.css"), which will host all the css to be used within the card component and it's extended variants 
2.) Refactor how the individual classes are called within the java classes and css


sven1103:
Great, thank you @Steffengreiner !

PR #256, opened 2023-06-06
Add experimental variables component

sven1103:
Without words:

image

### DIFF
--- a/vaadinfrontend/frontend/themes/datamanager/components/card.css
+++ b/vaadinfrontend/frontend/themes/datamanager/components/card.css
@@ -146,3 +146,63 @@
 .experiment-info .title {
   font-weight: bold;
 }
+
+.card.experimental-variables {
+  flex: auto;
+  font-size: var(--lumo-font-size-s);
+  padding: var(--lumo-space-l);
+}
+
+.experimental-variables .header {
+  color: var(--lumo-header-text-color);
+  font-size: var(--lumo-font-size-m);
+  font-weight: bold;
+  display: flex;
+  justify-content: space-between;
+  align-items: baseline;
+}
+
+.experimental-variables .content {
+  display: flex;
+  justify-content: space-between;
+  align-items: baseline;
+  overflow: auto;
+}
+
+.experimental-variables-fact-sheet {
+  flex: auto;
+  font-size: var(--lumo-font-size-s);
+  min-width: 100px;
+  margin-right: 50px;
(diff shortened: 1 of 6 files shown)


vaadinfrontend/frontend/themes/datamanager/components/card.css
```diff
@@ -146,3 +146,63 @@
 .experiment-info .title {
   font-weight: bold;
 }
+
+.card.experimental-variables {
+  flex: auto;
+  font-size: var(--lumo-font-size-s);
+  padding: var(--lumo-space-l);
+}
+
```
wow-such-code, line 155:
```suggestion
.card.experimental-variables {
  flex: auto;
  font-size: var(--lumo-font-size-s);
  padding: var(--lumo-space-l);
  margin: var(--lumo-space-s);
}
```
I suggest adding a margin so the cards are not stuck together. I like how the small margin looks, but previously it was larger (see testing instance).

wow-such-code: You would need to add the same to
`.card.experiment-info`
further up, so they are aligned.

wow-such-code: That's the margin between variables: 
image
The margin between both cards is missing.

wow-such-code (changes requested):
Very nice design! Apart from the small margin issue, SonarCloud complains about code duplication. I think the bigger problem in that respect is 
ExperimentInfoComponent.java that was not part of this PR, so your choice. 🤷

vaadinfrontend/frontend/themes/datamanager/components/card.css
sven1103: The margin between components will be addressed in an own PR, since we need refactor the complete Experimental Page anyway. I work bottom up: first the individual pieces and orchestrate everything at the end :)

sven1103:
Thank you @wow-such-code for your review! 

The sonar cloud warning is a hint, that at some point it might be worth to introduce a new intermediate component: a component that allows user interaction (interactive component). 

But I am not sure yet, if I want to introduce another class inheritance step, I would like to wait for the user interview results.

sven1103:
I introduced a custom comparator, since I was not happy with numbers / text variable levels not sorted properly in the variable sheets:
image

wow-such-code (approved):
good to go!

PR #257, opened 2023-06-06
create attachments component css

wow-such-code:
image
Disclaimer: attachments component is now not necessarily aligned on the bottom with the component on the left. We think this is something that has to be solved with the board component and it probably makes sense to solve it after all components have been refactored.

### DIFF
--- a/vaadinfrontend/frontend/themes/datamanager/components/custom.css
+++ b/vaadinfrontend/frontend/themes/datamanager/components/custom.css
@@ -1,8 +1,8 @@
 /* This file defines the styling and layout for all custom components. The components are sorted alphabetically */
 
 /*Make sure to import the css classes for each custom component to avoid cluttering the master-style-sheet */
-@import "card.css";
 @import "page-area.css";
+@import "card.css";
 
 .card-deck {
   display: flex;
--- a/vaadinfrontend/frontend/themes/datamanager/components/custom.css
+++ b/vaadinfrontend/frontend/themes/datamanager/components/custom.css
@@ -2,6 +2,7 @@
 
 /*Make sure to import the css classes for each custom component to avoid cluttering the master-style-sheet */
 @import "card.css";
+@import "page-area.css";
 
 .card-deck {
   display: flex;
--- a/vaadinfrontend/frontend/themes/datamanager/components/page-area.css
+++ b/vaadinfrontend/frontend/themes/datamanager/components/page-area.css
@@ -1,0 +1,42 @@
+/* All custom css for components which extend the page-area component go in here! */
+
+.page-area {
+  background-color: var(--lumo-base-color);
+  border: 1px solid;
+  border-color: var(--lumo-contrast-10pct);
+  border-radius: var(--lumo-border-radius-m);
+  box-shadow: var(--lumo-box-shadow-m);
+  box-sizing: border-box;
+  text-overflow: ellipsis;
+  width: 100%;
+  height: 100%;
+  margin: var(--lumo-space-s);
+  padding: var(--lumo-space-l);
+}
+
+.attachments-area {
+  display: flex;
+  gap: var(--lumo-space-m);
+  flex-direction: column;
+  /*
+  needed so the component ends on the right and bottom
+
+  width: unset;
+  height: unset;
+
+   */
+}
+
+.attachments-area .title {
+  font-weight: bold;
+  color: var(--lumo-secondary-text-color);
+  font-size: var(--lumo-font-size-xxl);
+  margin-bottom: 0.5rem;
+}
+
+.attachments-area .delete-button {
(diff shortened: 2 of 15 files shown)


vaadinfrontend/frontend/themes/datamanager/components/custom.css
```diff
@@ -2,6 +2,7 @@
 
 /*Make sure to import the css classes for each custom component to avoid cluttering the master-style-sheet */
 @import "card.css";
+@import "page-area.css";
```
KochTobi, line 5: Hi @wow-such-code 
As the last definition always has precedence, I think it would be good to have the most specific CSS at the bottom and general CSS on top. So reversing the order of the import might matter

vaadinfrontend/frontend/themes/datamanager/components/page-area.css
```diff
@@ -0,0 +1,35 @@
...
+  color: var(--lumo-secondary-text-color);
+  font-size: var(--lumo-font-size-xxl);
+  margin-bottom: 0.5rem;
+}
+
+.attachments-area .delete-button {
+  color: var(--lumo-error-text-color);
+  --lumo-button-size: var(--lumo-size-m);
+  cursor: pointer;
+  background-color: inherit;
```
KochTobi, line 34: Why do you inherit the background color? As the initial value for background-color is `transparent` this should not be necessary?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectLinksComponent.java
```diff
@@ -50,11 +56,23 @@ public ProjectLinksComponent(@Autowired ProjectLinkingService projectLinkingServ
...
-    getContent().addTitle(TITLE);
-    getContent().addContent(offerSearch, projectLinks);
+
+    initLayout();
+
     projectLinksComponentHandler = new ProjectLinksComponentHandler(projectLinkingService);
   }
 
+  private void initLayout() {
+    addClassName("attachments-area");
```
KochTobi, line 66: NIT: I think the class name should go to the top of the constructor as it does not add to the structure of the component but rather to its style.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectLinksComponent.java
```diff
@@ -50,11 +56,23 @@ public ProjectLinksComponent(@Autowired ProjectLinkingService projectLinkingServ
...
+    Span titleSpan = new Span();
+    titleSpan.add(TITLE);
```
KochTobi, line 68: should be `setText` I think

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectLinksComponent.java
```diff
@@ -103,7 +120,7 @@ public List<String> linkedOffers() {
   }
 
   public void setStyles(String... componentStyles) {
-    getContent().addClassNames(componentStyles);
+    addClassNames(componentStyles);
   }
```
KochTobi, line 124: what does this method do?

KochTobi (changes requested):
Some minor requests and questions :)

vaadinfrontend/frontend/themes/datamanager/components/page-area.css
wow-such-code: This is from the CSS created by adding the different button settings with Lumo. If I remove this setting it looks like this:
image
So the default value seems to be gray - I think it's the button background?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectLinksComponent.java
wow-such-code: This is directly taken from the way we did it before. Did you test if setText works as well?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectLinksComponent.java
wow-such-code: It adds css styles or Lumo stuff from the outside. It's not used at the moment - should I remove it?

wow-such-code: Ha, okay, I see. Since we are not using this Component with a wrapper anymore, this is not needed.

KochTobi (approved):

PR #258, opened 2023-06-07
Introduce support section component

wow-such-code:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/general/PageArea.java
```diff
@@ -5,11 +5,11 @@
...
+ * <b>A specific display area of the app that is neither header nor footer</b>
  *
- * <p><More detailed description - When to use, what it solves, etc.></p>
+ * <p>Contains main content OR support content (side-bar) of the respective page.</p>
  *
- * @since <version tag>
+ * @since <1.0.0>
  */
 @Tag(Tag.DIV)
 public class PageArea extends Component {
```
Steffengreiner, line 15: Shouldn't the PageArea Component implement the HasComponents interface So child element can be added?

Steffengreiner (changes requested):
Heya @wow-such-code, 

I have a question concerning the purpose of the support area: IMO we have 2 routes into which we can go direct here: 

1.) Make PageArea Abstract and introduce ContentArea and SupportArea components since we enforce the structure
2.) Don't distinguish between PageArea and the ContentArea /SupportArea

Steffengreiner (approved):
Nice work 🤔

PR #259, opened 2023-06-09
Minor UI improvements in ExperimentDetailsComponent

Steffengreiner:
1.) Change Numberfield Label in AddVariableDialog
2.) Change Cursor over menubar hovering
3.) Add Blocking Variable Card below experimentvariablecard 

**Edit** 

Additionally cleaned up css structure, removed outdated and not used css


vaadinfrontend/frontend/themes/datamanager/components/card.css
```diff
@@ -9,6 +9,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   margin: 4px 4px 4px 4px;
+  cursor: pointer;
```
sven1103, line 12: Why a pointer? It is not guaranteed that the user can interact with a card, thus I would not make it as default for the complete card.

sven1103 (changes requested):
Hi @Steffengreiner, I am not convinced, that we need the cursor, or I miss something here?

sven1103 (approved):
Great 🎉

PR #260, opened 2023-06-12
Ensure openbis connection before accessing the openbis api

KochTobi:


sven1103 (changes requested):
Dear @KochTobi i think this route will create us head-ache, when the app grows. It is easy to forget the call to the openBIS client to ensure trigger the session recreation.

I think this needs to go in the openBIS client itself, or you wrap every execution to the API into an own class and trigger it there everytime we request a token.

KochTobi:
However, I am unsure if we should depend on the Openbis client as it still uses the parent poms. We should clean it up if this is the direction we intend to take.

Steffengreiner (approved):
LGTM 🚀

PR #271, opened 2023-06-20
Add vaadin license key

KochTobi:


Steffengreiner (approved):
LGTM as long as the key is version specific otherwise i'd recommend to name the secret just 
VAADIN_SERVER 🚀

KochTobi:
It is a Vaadin server 23.2+ key

PR #277, opened 2023-06-22
Generalizes dialog styling and only allows closing via cancel button

wow-such-code:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/DefineExperimentComponent.java
```diff
@@ -28,9 +29,8 @@
  *
  * @since 1.0.0
  */
-public class DefineExperimentComponent extends Composite<VerticalLayout> {
+public class DefineExperimentComponent extends Div {
```
sven1103, line 32: I think we need a `@Tag(Tag.DIV)` annotation here, otherwise it wont be treated as native HTML div element

sven1103 (dismissed):
The dialog behavior is perfect, only the rendering of the form input elements in the project creation dialog seem to be destroyed.

After you have fixed that, I will happily approve

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/DefineExperimentComponent.java
sven1103: please ignore, since we extend `Div`, no need to add the tag explicitly (it is done in the `Div.class` already)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/overview/components/ProjectOverviewComponent.java
```diff
@@ -91,6 +94,17 @@ public ProjectOverviewComponent(@Autowired ClientDetailsProvider clientDetailsPr
...
+    //TODO remove this, only for testing
+    Button testButton = new Button("test experiment creation");
+    getContent().add(testButton);
+    testButton.addClickListener(new ComponentEventListener<ClickEvent<Button>>() {
+      @Override
+      public void onComponentEvent(ClickEvent<Button> event) {
+        ExperimentCreationDialog experimentCreationDialog = new ExperimentCreationDialog(experimentalDesignSearchService);
+        experimentCreationDialog.open();
+      }
+    });
```
sven1103, line 107: Since it is only for testing purposes, we should remove it before merging it :)

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/AddExperimentalGroupsDialog.java
```diff
@@ -23,16 +21,12 @@
  *
  * @since <version tag>
```
sven1103, line 22: Missing JDS

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/BatchRegistrationDialog.java
```diff
@@ -22,7 +20,7 @@
  *
  * <p>Component to register {@link Sample} with their associated metadata information</p>
  */
-public class BatchRegistrationDialog extends Dialog {
+public class BatchRegistrationDialog extends DialogWindow {
```
sven1103, line 23: div tag needed here?

sven1103 (commented):
Some last minor requests / quesions

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/BatchRegistrationDialog.java
wow-such-code: Good point! I tried it and it leads to Javascript errors and the Component being rendered below everything (not in a Dialog Window). So my take is that a Dialog cannot/should not be a Div.

PR #291, opened 2023-07-04
change error layout and lengthen notification for invalid sample input

wow-such-code:


Steffengreiner (approved):
LGTM 🚀 
The question is if this message should also be shown if a duplicate condition was selected during experimentalgroup selection

PR #295, opened 2023-07-05
Batch registration spreadsheet is only reset if the set experiment has changed

Steffengreiner:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleSpreadsheetLayout.java
```diff
@@ -103,14 +104,19 @@ public boolean isInputValid() {
...
+    this.experiment = experiment;
     experimentName.setText(experiment.getName());
+    SampleRegistrationSpreadsheet.setExperimentMetadata(experiment);
   }
 
   public List<SampleRegistrationContent> getContent() {
     return sampleInformationLayoutHandler.getContent();
   }
 
+  public Experiment getExperiment() {
```
sven1103, line 116: Since experiment is an entity not an DTO, this is not side effect free. Why is this needed?

sven1103 (changes requested):

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleSpreadsheetLayout.java
Steffengreiner: Good point I'll expose only the experimentID

sven1103 (approved):

PR #309, opened 2023-07-11
Allow user to filter the samples for each experiment via searchfield in the SampleOverViewComponent

Steffengreiner:
**What was Changed**

1.) Remove selection of the to be filtered experiment:
Input into the searchfield will trigger the filtering of samples in all grids via dedicated event propagation. 
2.) Introduce JPA specifications {and a lot of table joins} to enable the filtering of samplepreviews for all fields. 

**Unclear how to solve** 
It is unclear how the experiment tabs should behave if they have not been selected yet. (Selection of tab triggers lazy loading). Currently the itemcount of the tab is only updated if they have been selected before.

### DIFF
--- a/database-connector/src/main/java/life/qbic/projectmanagement/experiment/persistence/SamplePreviewJpaRepository.java
+++ b/database-connector/src/main/java/life/qbic/projectmanagement/experiment/persistence/SamplePreviewJpaRepository.java
@@ -7,9 +7,11 @@ import life.qbic.projectmanagement.application.SortOrder;
 import life.qbic.projectmanagement.application.sample.SamplePreview;
 import life.qbic.projectmanagement.application.sample.SamplePreviewLookup;
 import life.qbic.projectmanagement.domain.project.experiment.ExperimentId;
+import org.eclipse.jetty.util.StringUtil;
 import org.springframework.context.annotation.Scope;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.domain.Sort.Order;
+import org.springframework.data.jpa.domain.Specification;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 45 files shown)


database-connector/src/main/java/life/qbic/projectmanagement/experiment/persistence/SamplePreviewJpaRepository.java
```diff
@@ -40,13 +45,133 @@ public List<SamplePreview> queryByExperimentId(ExperimentId experimentId, int of
...
+  public int queryCountByExperimentId(ExperimentId experimentId, String filter) {
+    Specification<SamplePreview> filterSpecification = generateExperimentIdandFilterSpecification(
+        experimentId, filter);
+    return (int) samplePreviewRepository.count(filterSpecification);
+  }
+
+  private Specification<SamplePreview> generateExperimentIdandFilterSpecification(
+      ExperimentId experimentId, String filter) {
+    Specification<SamplePreview> isBlankSpec = SamplePreviewSpecs.isBlank(filter);
+    Specification<SamplePreview> experimentIdSpec = SamplePreviewSpecs.experimentIdEquals(
```
wow-such-code, line 65: is this experiment name or really id? and why is it needed (especially as "equals") for the sample table?

wow-such-code (approved):
looks good to me

database-connector/src/main/java/life/qbic/projectmanagement/experiment/persistence/SamplePreviewJpaRepository.java
Steffengreiner: This specification is the ID of the experiment since that is the unique identifier. The idea here is to only fetch the samplepreview information for that specific id(which is why it's an equal statement) :)

PR #310, opened 2023-07-11
Prepares embedding of project details component

sven1103:
Prepares the component such that we can embedd it properly in a PageArea, which is taking care of the correct rendering.

The rest of the component works smoothly, i did not wont to interefere here.


KochTobi (approved):
LGTM

PR #311, opened 2023-07-11
Extend database and domain for 'analysis performed' and 'comment'

wow-such-code:


KochTobi (changes requested):
Hi @wow-such-code 
your build fails ;) Please address the failure.

projectmanagement/src/main/java/life/qbic/projectmanagement/application/sample/SamplePreview.java
```diff
@@ -79,6 +82,9 @@ private SamplePreview(ExperimentId experimentId, SampleId sampleId, String sampl
     this.species = species;
     this.specimen = specimen;
     this.analyte = analyte;
+    // optional columns
+    this.comment = comment == null ? "" : comment;
```
KochTobi, line 86:
```suggestion
    this.comment = comment;
```
Please make the field nullable and change the getter. This is a good example of where the optional type is intended to be used.

projectmanagement/src/main/java/life/qbic/projectmanagement/application/sample/SamplePreview.java
```diff
@@ -79,6 +82,9 @@ private SamplePreview(ExperimentId experimentId, SampleId sampleId, String sampl
...
+    this.analysisType = analysisType == null ? "" : analysisType;
```
KochTobi, line 87:
```suggestion
    this.analysisType = analysisType;
```

projectmanagement/src/main/java/life/qbic/projectmanagement/application/sample/SamplePreview.java
```diff
@@ -146,6 +154,12 @@ public String specimen() {
   public String analyte() {
     return analyte;
   }
+  public String analysisType() {
+    return analysisType;
+  }
+  public String comment() {
+    return comment;
+  }
```
KochTobi, line 162: As they are optional, why not return an Optional here?
```suggestion
  public Optional<String> analysisType() {
    return Optional.ofNullable(analysisType);
  }
  public Optional<String> comment() {
    return Optional.ofNullable(comment);
  }
```

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/sample/SampleRegistrationRequest.java
```diff
@@ -15,21 +15,25 @@
...
-      SampleOrigin sampleOrigin) {
+      SampleOrigin sampleOrigin, String analysisType, String comment) {
     this.label = Objects.requireNonNull(label);
     this.assignedBatch = Objects.requireNonNull(assignedBatch);
     this.experimentId = Objects.requireNonNull(experimentId);
     this.experimentalGroupId = Objects.requireNonNull(experimentalGroupId);
     this.replicateReference = Objects.requireNonNull(replicateReference);
     this.sampleOrigin = Objects.requireNonNull(sampleOrigin);
+    this.comment = comment == null ? "" : comment;
+    this.analysisType = analysisType == null ? "" : analysisType;
```
KochTobi, line 37: why the default values?

KochTobi (changes requested):
Hi @wow-such-code 
Thank you for the changes. I'd like to motivate the use of the Optional class for these optional properties.
Looking forward to your changes :)

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/sample/SampleRegistrationRequest.java
```diff
@@ -15,21 +16,32 @@
...
     this.sampleOrigin = Objects.requireNonNull(sampleOrigin);
+    this.comment = comment;
+    this.analysisType = analysisType;
+  }
+
+  public Optional<String> getAnalysisType() {
+    return Optional.ofNullable(analysisType);
+  }
+  public Optional<String> getComment() {
+    return Optional.ofNullable(comment);
```
KochTobi, line 45: I read up on it. As soon as you have optional fields you should not use a record. The getter will still give back `null` so if you call `sampleRegistrationRequest.analysisType()` you would get `null`.
We should convert this to a class instead.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/sample/Sample.java
```diff
@@ -54,6 +58,8 @@ private Sample(SampleId id, SampleCode sampleCode, BatchId assignedBatch, String
     this.sampleOrigin = sampleOrigin;
     this.biologicalReplicateId = replicateReference;
     this.assignedBatch = assignedBatch;
+    this.analysisType = analysisType.orElse("");
+    this.comment = comment.orElse("");
```
KochTobi, line 62: I think the sample itself should have `null` values in these fields as they are optional for the sample.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/sample/SampleRegistrationRequest.java
wow-such-code: I had this problem when I used the original method names "analysisType()" and "comment()". The suggestion to fix this problem was to explicitly add the "get" to the method name. This difference is also displayed by the IDE:
image
See first reply here for the solution I used: https://stackoverflow.com/questions/62945049/java-records-with-nullable-components

KochTobi: yes that happens because there would be two methods with the same signature and name but different return type
`public String analysisType()` and `public Optional<String> analysisType()` which is not possible in Java. That is why I recommend overwriting the method and making it a class instead of a record or not using optional in this case but simply returning null and make use of optional in the Sample class.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/sample/Sample.java
KochTobi: the sample does have nullable fields and optional getters. the sample registration request can have null values. that is fine by me. 
In the initial review, I missed the record for the constructor.

KochTobi (approved):
Hi @wow-such-code thank you for your patience!

PR #320, opened 2023-07-17
Preserve dropdown menus after selection in spreadsheet

wow-such-code:
* Dropdown menus in the sample registration spreadsheet are now editable after the initial selection
* Filled or empty menus can be "pulled" (copied) to other editable cells
* This comes with the caveat that selection of a cell containing a dropdown menu is more difficult. It is however possible using the keyboard

**How is it done and why?**
* Counter-intuitively, the method _getCustomEditorForCell_ is not used - this method only shows editor cells (here: comboboxes) when a user clicks on the cell
* Copying of values is done in the following way: since components can not be copied, selecting an item from the dropdown also stores that value in the cell (as opposed to the component shown instead of the cell). The method _getCustomComponentForCell_ checks if the value in the cell is part of the Combobox item list. If it is, the Combobox for that cell is created with that item already pre-selected.


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SpreadsheetDropdownFactory.java
```diff
@@ -46,33 +43,45 @@ public Component getCustomComponentForCell(Cell cell, int rowIndex, int columnIn
...
+    comboBox.addValueChangeListener(e -> {
+      //when a selection is made, the value is set to the cell (in addition to the component)
+      //this is needed for copying of inputs to other cells works
       Cell cell = spreadsheet.getCell(rowIndex, columnIndex);
       cell.setCellValue(e.getValue());
-      spreadsheet.refreshCells(cell);
     });
-    return analysisType;
+    // allows copying of Comboboxes before a selection was made
+    Cell cell = spreadsheet.getCell(rowIndex, columnIndex);
```
KochTobi, line 81: initial empty value, correct?

KochTobi (approved):

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SpreadsheetDropdownFactory.java
KochTobi: NIT: you could simply use the emptyValue of the combobox here instead of ""

wow-such-code: Yes

PR #327, opened 2023-07-19
Display numbers of samples to create in custom column

wow-such-code:
* Hides standard letter and digit header and row ids of the spreadsheet
* Creates custom column (by adding "ROW" to header Enums) that shows a bold enumeration in the (currently) leftmost column
* Fixes an issue where columns that had pre-filled data were editable before and - in some cases - after validation
* Removes some methods used to figure out which columns should be editable - replaces them by a method that decides by the header of that column
* Renames some methods to better reflect their functionality

image

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
@@ -32,6 +32,7 @@ import org.apache.poi.ss.usermodel.BorderStyle;
 import org.apache.poi.ss.usermodel.Cell;
 import org.apache.poi.ss.usermodel.CellStyle;
 import org.apache.poi.ss.usermodel.Font;
+import org.apache.poi.ss.usermodel.HorizontalAlignment;
 import org.apache.poi.ss.usermodel.IndexedColors;
 import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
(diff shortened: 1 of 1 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
```diff
@@ -269,22 +292,24 @@ private void prepareCommonSheetTasks() {
...
         Cell cell = this.createCell(row, column, "");
         cells.add(cell);
       }
     }
-    defaultStyleAndUnlockCells(cells);
+    defaultStyleAndUnlockEditableCells(cells);
   }
 
-  void defaultStyleAndUnlockCells(Set<Cell> cells) {
+  void defaultStyleAndUnlockEditableCells(Set<Cell> cells) {
```
Steffengreiner, line 306: What does defaultStyle mean since the only change i can see is that the cells are not locked?

Steffengreiner (approved):
Heya @wow-such-code , Nice changes!
I'm curious what does defaultstyle means in this context? 

Also the spreadsheet seems completely broken: I found 2 weird occurences in the spreadsheet which i'm unsure about that are already on development: 
1.) **Open and Close Dialog and open dialog again error seems to be back**
Bildschirmfoto 2023-07-20 um 13 44 52

2.) **There are many empty columns after the comment column in the spreadsheet** 
Bildschirmfoto 2023-07-20 um 13 49 49

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
wow-such-code: I introduced this name in the validation highlighting PR. It means cells with "normal" text and without any highlighting.

PR #329, opened 2023-07-20
User can view edit and delete batch information for experiments 

Steffengreiner:
**What was changed** 
Implemented a basic batchdetailscomponent, allowing the user to edit the name of a batch and delete a batch in the samplemaincomponent

**It's a draft**
Currently it's unclear what editability the component should provide and how much space the component should take. The original vision was made for a singular experiment which is not the case anymore


wow-such-code:
I'm so sorry

PR #336, opened 2023-07-28
Inform User that row deletion is not possible if only one row remains in the spreadsheet. 

Steffengreiner:
**What was changed**
The user is now unable to delete the final row(outside of the columnheaders) and is informed via a notification that this is not possible. 

**Technical Note**
I wanted to propagate the notification via an applicationexception within the layout, however this leads to the spreadsheet component breaking entirely, which is why a dedicated notification was necessary.

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
@@ -220,6 +220,10 @@ public class SampleRegistrationSpreadsheet extends Spreadsheet implements Serial
    */
   public void deleteRow(int index) {
     //delete row
+    if (getRows() == 1) {
+      // only one row remaining -> the header row
+      return;
+    }
     deleteRows(index, index);
     //move other rows up
     if (index + 1 < getRows()) {
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
@@ -219,10 +219,6 @@ public class SampleRegistrationSpreadsheet extends Spreadsheet implements Serial
    * @param index 0-based row index of the row to remove
    */
   public void deleteRow(int index) {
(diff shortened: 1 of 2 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
```diff
@@ -219,10 +219,6 @@ private void generateRowHeaderCell(int colIndex, int rowIndex) {
    * @param index 0-based row index of the row to remove
    */
   public void deleteRow(int index) {
-    if (getRows() == 1) {
-      // only one row remaining -> the header row
-      return;
-    }
```
KochTobi, line 225: this needs to stay in. The last row can never be deleted (no matter what assumption is taken before).

KochTobi (changes requested):
Hi @Steffengreiner 
Nice work! Two minor things.

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleSpreadsheetLayout.java
```diff
@@ -76,14 +84,28 @@ private void initButtonLayout() {
...
+        });
     registerButton.addThemeVariants(ButtonVariant.LUMO_PRIMARY);
     sampleInformationButtons.add(backButton, addRowButton, deleteRowButton, cancelButton,
         registerButton);
     add(sampleInformationButtons);
   }
 
+  private void initNotifications() {
+    addComponentAsFirst(lastRowDeletionNotification);
+  }
```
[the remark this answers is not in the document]
Steffengreiner: Good point. Do you think it makes sense to throw an exception here?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/registration/batch/SampleRegistrationSpreadsheet.java
Steffengreiner: Good point. Do you think it makes sense to throw an exception here?

KochTobi: no it is not that exceptional and I do not think anyone wants to delete the header

KochTobi (approved):

PR #338, opened 2023-08-02
Adds prefill option to sample batch registration

wow-such-code:
Small open issues/ideas with this:
* A reload() is necessary to make this work with the back button (user loads sheet, goes back, then selects prefill option) - not sure if this is computationally expensive/problematic
* Setting the experiment to null upon closing the dialog might be a bad idea. @Steffengreiner proposed a complete recreation of the Dialog upon close and reopen, which might make this issue obsolete.


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -85,6 +85,8 @@ public class SampleDetailsComponent extends PageArea implements Serializable {
   public final Button registerButton = new Button("Register");
   private final Button metadataDownloadButton = new Button("Download Metadata");
   private final TabSheet sampleExperimentTabSheet = new TabSheet();
+
+  //This should be rebuilt from scratch to avoid reset issues if it was opened before
```
Steffengreiner, line 89: We'll tackle that as some would say "soon"™️ 
```suggestion
```

Steffengreiner (approved):
LGTM 🚀

PR #346, opened 2023-08-23
Data-manager-application pages use the entire screen width and height independent of content 

Steffengreiner:
**What was changed**
This PR intends to set the height and width of the main component and its containing page areas to employ the entire screen independent of the content within. For this the screen height and width has to be propagated through all the relevant div containers.


wow-such-code (approved):
works on my machine!

PR #348, opened 2023-08-24
Show disclaimer if no samples have been registered in an experiment 

Steffengreiner:
**What was changed**
1.) Fix logic so no samples registered disclaimer is shown when no samples have been registered in an experiment instead of an empty grid. 
2.) Rewrite disclaimer message according to [figma](https://www.figma.com/file/mYcW5viu5GpPZ852VHYBrV/Start-Page?type=design&node-id=3694-97721&mode=design&t=P6Hd6bthfqiKbqG7-0) prototype.


wow-such-code (approved):
very nice

wow-such-code (approved):

PR #353, opened 2023-08-29
Set Generationtype strategy to AUTO for id generation

Steffengreiner:
**What was changed** 
1.) The GenerationType strategy for the ids within experimental variable and experimental groups should be set to Auto so it's incremented correctly. Otherwise the generation of a new group/variable will fail since there is no default value provided 

2.) Remove unnecessary postfiler on samplePreview query since there are no permissions implemented yet


wow-such-code (approved):
works, thank you!

KochTobi:
@Steffengreiner please revert

PR #354, opened 2023-08-29
Revert last commit and set identitygeneration back to identity strategy for ids 

Steffengreiner:
**What was changed** 
We decided that the tables should be responsible for auto incrementing instead of hibernate


KochTobi (approved):
Thank you @Steffengreiner

PR #357, opened 2023-08-30
User can see the button in the spreadsheet layout independent of number of rows

Steffengreiner:
**What was changed**
This PR achieves two things:

1.) Adapt styling so the dialog buttons are always visible in the spreadsheetlayout 

2.) Ensure the spreadsheet is prefilled correctly(or not prefilled) and streamline reload logic in spreadsheetlayout

**How to Test**

1.) Select "My new Experiment" in "SampleRegistrationSheeTtester" project on smaller screen size on qpylon and locally and try to see if the buttons are shown. 

2.) Select "Experiment 0" in "SampleRegistrationSheeTtester" project and observe "orphaned" entries in replicate column of the spreadsheet even if prefill was not selected: 
Bildschirmfoto 2023-08-30 um 15 11 01

### DIFF
--- a/vaadinfrontend/frontend/themes/datamanager/components/dialog.css
+++ b/vaadinfrontend/frontend/themes/datamanager/components/dialog.css
@@ -27,8 +27,14 @@ vaadin-dialog-overlay::part(title) {
   height: 100%
 }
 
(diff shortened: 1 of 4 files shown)


KochTobi (approved):
LGTM

PR #363, opened 2023-09-06
align add-variable dialog with UI of add-group dialog

wow-such-code:
* remove one of the two titles in the add variable dialog
* style adding of variables like in add-group dialog
* if the last variable row is removed in "add mode" (as opposed to edit mode), a new empty row is automatically added
* remove some unnecessary CSS


KochTobi (approved):
LGTM great job!

PR #364, opened 2023-09-07
Separate Experiment from Project creation

KochTobi:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/create/DefineExperimentComponent.java
```diff
@@ -1,184 +0,0 @@
-package life.qbic.datamanager.views.projects.create;
```
wow-such-code, line 1: why was this component deleted - is it not needed to create experiments independently of projects?

KochTobi: There is no `define Experiment Component` anymore. This was an abstraction so the same could be displayed in multiple dialogs. As it is only needed in one place it is not only in the experiment dialog

wow-such-code:
branch renamed to better reflect the changes. see https://github.com/qbicsoftware/data-manager-app/pull/371

PR #365, opened 2023-09-08
Notify the user via email about granted project access 

sven1103:
After a user has been granted access to a project, they will now get
notified with an email.

The email also contains a link to the project to simplify access for the user.


authorization/src/main/java/life/qbic/authorization/application/AppContextProvider.java
```diff
@@ -0,0 +1,21 @@
+package life.qbic.authorization.application;
```
Steffengreiner, line 1: Since then navigation is based on URL and the user authorization is not checked in the contextprovider i'd recommend moving the interface from the `authorization `package to the more general `life.qbic.datamanager` package

authorization/src/main/java/life/qbic/authorization/application/AppContextProvider.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * <b>App Context Provider</b>
+ * <p>
+ * Provides some utility methods to create navigation targets within the application.
+ *
+ * @since 1.0.0
+ */
+public interface AppContextProvider {
+
+  /**
```
Steffengreiner, line 12: This looks similiar to the `PasswortResetLinkSupplier` class maybe this could be streamlined?

authorization/src/main/java/life/qbic/authorization/application/policy/directive/InformUserAboutGrantedAccess.java
```diff
@@ -0,0 +1,77 @@
...
+        Please click the link below to access the project after login:
+             
+        %s
+              
+        Need help? Contact us for further questions at support@qbic.zendesk.com
+              
+        Best regards,\
+        The QBiC team
+        """, recipient.fullName(), projectTitle, appContextProvider.urlToProject(projectId));
+  }
```
Steffengreiner, line 58: We have a `EmailFactory` class which i'd expect to provide the content of the email messages. Is there a reason for the seperation here?

domain-concept/src/main/java/life/qbic/domain/concepts/communication/EmailService.java
```diff
@@ -3,4 +3,6 @@
 public interface EmailService {
 
   void send(Email email);
+
+  void send(String recipient, String recipientFullName, String subject, String message);
```
Steffengreiner, line 7: IMO this is unnecessary since we can just generate and provide the email record. (Even better if that's handled by the UserContactService similiar to the password reset email generation)

authorization/src/main/java/life/qbic/authorization/application/policy/directive/InformUserAboutGrantedAccess.java
```diff
@@ -0,0 +1,77 @@
...
+  public void handleEvent(ProjectAccessGranted event) {
+    jobScheduler.enqueue(() -> notifyUser(event.forUser(), event.forProject(), event.withTitle()));
+  }
+
+  @Job(name = "Notify user about granted project access")
+  public void notifyUser(String userId, String projectId, String projectTitle)
+      throws RuntimeException {
+    var recipient = userRepository.findById(UserId.from(userId)).get();
+    emailService.send(recipient.emailAddress().get(), recipient.fullName().get(),
+        "Project access granted", composeMessage(projectId, recipient, projectTitle));
```
Steffengreiner, line 75: IMO we can just generate and provide the email record here
```suggestion
        new Email(composeMessage(projectId, recipient, projectTitle), "Project access granted",
            "no-reply@qbic.life",
            new Recipient(recipient.emailAddress().get(), recipient.fullName().get()),
            "text/plain"));
```

newshandler/src/main/java/life/qbic/newshandler/usermanagement/email/EmailSubmissionService.java
```diff
@@ -78,6 +79,16 @@ public void send(Email email) {
...
 
+  @Override
+  public void send(String recipientAddress, String recipientFullName, String subject, String message) {
+    var email = new Email(message, subject, "no-reply@qbic.life",
+        new Recipient(recipientAddress, recipientFullName),  "text/plain");
+    sendPlainEmail(email).ifSuccessOrElse(
+            successResponse -> reportSuccess(successResponse, email),
+            failureResponse -> reportFailure(failureResponse, email)
+    );
+  }
```
Steffengreiner, line 90: Unnecessary if we stick to the Email DTO which enables us to avoid redundancy here :) 
```suggestion
```

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/service/event/ProjectAccessGranted.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Objects;
+import life.qbic.domain.concepts.DomainEvent;
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
```
Steffengreiner, line 14: Missing JD

Steffengreiner (changes requested):
Heya @sven1103 ! Thanks for the awesome implementation. 
However i have some questions: 
1.) The email generation code looks very simliar to the code provided in the `UserContactService`, is there a reason for a distinction here? 
2.) I was unable to get an email if i granted project access to my user accounts or after project creation (@uni-tuebingen, @qbic.uni-tuebingen.de). Do i need to set additional environment variables or parameters to trigger the email sending?
I checked Jobrunner and found the following error: 

```
org.jobrunr.scheduling.exceptions.JobClassNotFoundException:
life.qbic.authorization.application.policy.directive.InformUserAboutGrantedAccess.notifyUser(java.lang.String,java.lang.String,java.lang.String)
```

authorization/src/main/java/life/qbic/authorization/application/AppContextProvider.java
sven1103: Dependency Inversion Principle. My only concern here was to introduce lose coupling so the auth module is not depending on the app UI module.

authorization/src/main/java/life/qbic/authorization/application/AppContextProvider.java
sven1103: Yes exactly, however this was my first effort to refactor it. The reset link supplier and confirmation link supplier would be also implemented in the app module and the interface extended accordingly.

authorization/src/main/java/life/qbic/authorization/application/policy/directive/InformUserAboutGrantedAccess.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @since 1.0.0
+ */
+@Component
+public class InformUserAboutGrantedAccess implements DomainEventSubscriber<ProjectAccessGranted> {
+
+  private final EmailService emailService;
+
+  private final JobScheduler jobScheduler;
+  private final UserRepository userRepository;
```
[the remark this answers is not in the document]
sven1103: I think this would be a strong violation of the Single Responsibility Pattern. I think it is not of the user contact service concern to know what it is about, but to deliver a message to a user with a given id. Imo, the service should care about finding the user's contact email address and sending the mail and only this. The content comes from the domain, where business logic policy demand to inform the user about an event.

authorization/src/main/java/life/qbic/authorization/application/policy/directive/InformUserAboutGrantedAccess.java
sven1103: Different domain.

domain-concept/src/main/java/life/qbic/domain/concepts/communication/EmailService.java
sven1103: The issue is, that Email is introduced as domain concept but with a technical low level detail: mime type. To not break existing functionality, I decided to override the method with simple String values.

PR #367, opened 2023-09-11
Preserve spacing between title and button in ExperimentList component 

Steffengreiner:
**What was changed**

1.) Move experiment item collection css into Div.css style sheet
2.) Add gap between button and title ensuring spacing on small screens


sven1103 (approved):
Great!

PR #370, opened 2023-09-12
Ensure sample table readability on smaller screens 

Steffengreiner:
**What was changed ** 
1.) Change Sample Overview Grid to wrap cell content on whitespace as a default behaviour. 
2.) Introduce tooltips since some values can overflow
3.) Enforce that Sample Id and conditions are always visible and not changed due to grid sizing


wow-such-code (approved):
good job!

PR #372, opened 2023-09-13
Remove navigation bar from project information page

sven1103:
Since we are going to introduce a different navigation scheme soon, the navbar is obsolete in the project information page, since the progress is only relevant on experiment level.


Steffengreiner (approved):
Nicely done 👍

PR #373, opened 2023-09-15
Introduce translation of terms in openBIS connector

wow-such-code:


sven1103 (approved):
Great, this helps in swapping that out for Ontology mappers.

PR #376, opened 2023-09-19
Fix OpenBis Analyte Mapping and restore BatchRegistrationDialog functionality

Steffengreiner:
**What was changed**
1.) Unnecessary @Override leading to close dialog recursion
2.) DialogWindow is not applicable for batchRegistrationDialog without major refactor since buttons functionality changes inbetween tabs
3.) OpenBis Mapping based on label but property storage is based on code


wow-such-code (approved):
thank you!

PR #381, opened 2023-09-20
Fix email communication service

sven1103:
Due to some syntax error, the application configuration variables
for the email service did not get resolved, resulting in a failing email communication service.

This PR fixes the issue and also provides a resolvable link to a project after granted access.


Steffengreiner (approved):
LGTM 🚀

PR #386, opened 2023-09-26
Add flask icon to experimentList item and update style

Steffengreiner:
**What was changed**

1.) Add Flask icon to experimentItem 
2.) Update css to set font-size and padding according to figma prototype
3.) Sort items in experimentList properly by experimentName, ignoring case of item.


sven1103 (approved):
LGTM

PR #387, opened 2023-09-28
Introduce funding section to provide project grant information

sven1103:
A new custom field for funding information is introduced to provide grant information, which enriches the available project information to store.

### DIFF
--- a/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/Funding.java
+++ b/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/Funding.java
@@ -1,6 +1,7 @@
 package life.qbic.projectmanagement.domain.project;
 
 import jakarta.persistence.Embeddable;
+import java.util.Objects;
 
 /**
  * A small business value object representing information about funding.
@@ -41,5 +42,23 @@ public class Funding {
 
   public String grantId() {
     return this.grantId;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    Funding funding = (Funding) o;
+    return Objects.equals(grant, funding.grant) && Objects.equals(grantId,
+        funding.grantId);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(grant, grantId);
   }
 }
--- a/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/Funding.java
+++ b/projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/Funding.java
@@ -1,0 +1,45 @@
+package life.qbic.projectmanagement.domain.project;
+
+import jakarta.persistence.Embeddable;
+
+/**
+ * A small business value object representing information about funding.
+ * <p>
+ * Funding is described by a free-text label and a funding body's reference identifier.
+ * <p>
+ * For example:
+ * <ul>
+ *   <li>SFB (label)</li>
+ *   <li>SFB 1101 (reference id)</li>
+ * </ul>
+ *
+ * @since 1.0.0
+ */
+@Embeddable
+public class Funding {
+
+  private String grant;
+
+  private String grantId;
+
+  protected Funding() {
+
+  }
+
+  private Funding(String grant, String grantId) {
+    this.grant = grant;
+    this.grantId = grantId;
+  }
+
+  public static Funding of(String grant, String grantId) {
+    return new Funding(grant, grantId);
+  }
+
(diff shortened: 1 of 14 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/ProjectFormLayout.java
```diff
@@ -74,6 +77,23 @@ public ProjectFormLayout() {
         .bind((ProjectInformation::getProjectObjective),
             ProjectInformation::setProjectObjective);
 
+    fundingField = new FundingField("Funding Information");
+    binder.forField(fundingField)
+        .withValidator(field -> {
```
KochTobi, line 82: The element you are passing into the lambda is not the field but rather its value. So I suggest renaming it to `fundingInformation` or `funding`

vaadinfrontend/src/main/java/life/qbic/datamanager/views/projects/ProjectFormLayout.java
```diff
@@ -74,6 +77,23 @@ public ProjectFormLayout() {
...
+              if (field == null) {
+                return true;
+              }
+              if (!field.getLabel().isBlank()) {
+                return !field.getReferenceId().isBlank();
+              } else if (!field.getReferenceId().isBlank()) {
+                return !field.getLabel().isBlank();
+              }
+              return true;
```
KochTobi, line 91: Should be split up into different validators to allow for different error messages depending on the cause. Right now it is not possible to say something like `please provide the funding reference id` when only funding label is filled out.

projectmanagement/src/main/java/life/qbic/projectmanagement/domain/project/Funding.java
```diff
@@ -0,0 +1,45 @@
...
+ * For example:
+ * <ul>
+ *   <li>SFB (label)</li>
+ *   <li>SFB 1101 (reference id)</li>
+ * </ul>
+ *
+ * @since 1.0.0
+ */
+@Embeddable
+public class Funding {
```
KochTobi, line 19: Please provide matching `equals` and `hashCode` methods.

vaadinfrontend/frontend/themes/datamanager/components/dialog.css
```diff
@@ -182,6 +182,12 @@ vaadin-dialog-overlay::part(title) {
   align-items: baseline;
 }
 
+.create-project-dialog .input-fields {
+   display: flex;
+   gap: 1em;
+   align-items: baseline;
+ }
+
```
KochTobi, line 190: why is this necessary?

KochTobi (changes requested):

KochTobi (changes requested):
Hi @sven1103 
The code looks good. However, I would argue that the validator can be split into two giving us more flexibility without increasing complexity much. 
I further have a question to the intended functionality. When funding information is provided at any point, it is impossible to remove it again. Instead an empty funding entry is stored. Is this desired?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/general/funding/FundingField.java
```diff
@@ -0,0 +1,65 @@
...
+    super();
+    setLabel(fieldLabel);
+    this.label = new TextField("Grant", "e.g. SFB");
+    this.referenceId = new TextField("Grant ID", "e.g. SFB 1101");
+    this.binder = new Binder<>();
+    layoutComponent();
+    configureBinder();
+  }
+
+  private void configureBinder() {
```
sven1103, line 34: can be removed

vaadinfrontend/frontend/themes/datamanager/components/dialog.css
KochTobi: Isn't that done already?

KochTobi:
```
.create-project-dialog .code-and-title,
.contact-field .input-fields {
  display: flex;
  gap: 1em;
  align-items: baseline;
}
```

vaadinfrontend/src/main/java/life/qbic/datamanager/views/general/funding/FundingField.java
```diff
@@ -45,6 +34,10 @@ private static Div layoutFundingInput(TextField label, TextField referenceId) {
     return layout;
   }
 
+  private void layoutComponent() {
```
KochTobi, line 37: You could inline this method.
Something like `add(layoutInputs)` is readable enough and when looking for the layout can be detected directly.

KochTobi (approved):
beautiful LGTM

PR #393, opened 2023-10-02
Enable User to navigate within a project with an application drawer component 

Steffengreiner:
**What was changed**

1.) Introduce appdrawer for navigation purposes with dedicated project and experiment section
 
**ProjectSection**
A) User can select from the 3 last modified project within a dedicated selectioncomponent
B) User can go to projectOverview Component within selection component

**ExperimentSection**
A) User can select an experiment from the list of experiments and is routed to the ExperimentMainComponent
2.) User can toggle appdrawer only within a project and the content is updated dependent on the project

**What is missing** 
BUG: Currently the selected experiment is not updated correctly if the user selects the experiment within the experimentlist (although it's set correctly if the user is routed directly via an URL)

**How to solve**
I'd suggest waiting for the API to catch up since it's already `[documented](https://github.com/vaadin/flow-components/issues/5093)` being tackled and hopefully added in a later release:


vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
```diff
@@ -0,0 +1,206 @@
...
+    Objects.requireNonNull(experimentInformationService);
+    this.addClassName("project-navigation-drawer");
+    this.projectInformationService = projectInformationService;
+    this.experimentInformationService = experimentInformationService;
+    projectSection.addClassName("project-section");
+    experimentSection.addClassName("experiment-section");
+    add(projectSection, experimentSection);
+    log.debug(
+        "New instance for ProjectSideNavigationComponent {} was created",
+        System.identityHashCode(this));
```
sven1103, line 81: Why is this necessary to log?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
```diff
@@ -0,0 +1,206 @@
...
+  @Override
+  public void beforeEnter(BeforeEnterEvent beforeEnterEvent) {
+    resetSections();
+    String projectId = beforeEnterEvent.getRouteParameters()
+        .get(PROJECT_ID_ROUTE_PARAMETER).orElseThrow();
+    ProjectId parsedProjectId = ProjectId.parse(projectId);
+    this.context = new Context().with(parsedProjectId);
+    beforeEnterEvent.getRouteParameters().get(EXPERIMENT_ID_ROUTE_PARAMETER)
+        .ifPresent(experimentId -> this.context = context.with(ExperimentId.parse(experimentId)));
+    setContext();
```
sven1103, line 93: loadProjectInformation?

vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  private void setContext() {
+    initializeProjectSection(context.projectId().orElseThrow());
+    initializeExperimentSection(context.projectId().orElseThrow());
+  }
+
+  private void initializeProjectSection(ProjectId projectId) {
```
sven1103, line 101: this method introduces side effects (manipulating project section header and project section). Please retrieve the required project information once, and pass required information to the methods that will build and return a configured side nav element. 
On top level, you can then compose the nav bar, with the advantage that your code is not as nested (which makes reading and debugging hard) and you have fewer side effects. Try to remove any side effect in your functions, as they easily lead to unexpected behaviour. 
Also, your function does more than one thing. Can you spot it? ;) Please honor the definition of a function.

sven1103 (approved):
Hi Steffen, 

thank you for your contribution!

I have some concerns regarding your code structure. Please honor the definition of a function and implement a composite approach of building your user interface elements. 
You code is currently very nested, which decreases readability, ease of debugging and introduces side effects.

sven1103 (changes requested):
Sry, wrong button clicked :/

vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
Steffengreiner: We log all our component initialization currently. I think it makes sense for debugging purposes to see if the components could be created correctly 🤔

vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
Steffengreiner: Good point I made the initialization clearer

vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
Steffengreiner: I rewrote how the section are initialized to remove their interdependency as much as possible and load the experiment/project information beforehand 👍

sven1103 (approved):

PR #394, opened 2023-10-04
Remove analysis type remnants and require non-null values for analysis method in samplePreview 

Steffengreiner:


KochTobi (approved):
LGTM

PR #400, opened 2023-10-10
Increase reusability for the spreadsheet component

KochTobi:
Introduces a Spreadsheet component with a custom API.
The spreadsheet can be tested using the test view at `/test`. This view will be removed.
image


user-interface/src/main/java/life/qbic/datamanager/views/general/spreadsheet/ColumnValidator.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.function.Predicate;
+
+/**
+ * Validates the content of a cell. Can be applied to a column.
+ * <p>
+ * One column can have multiple validators. The spreadsheet can use the validators to evaluate the
+ * validity of cells in the column.
+ */
+public class ColumnValidator<T2> {
```
wow-such-code, line 11: NIT: why T2 and not just T?

user-interface/src/main/java/life/qbic/datamanager/views/general/spreadsheet/Spreadsheet.java
```diff
@@ -0,0 +1,671 @@
...
+
+  private class HeaderRow extends Row {
+
+  }
+
+  /**
+   * This SpreadsheetComponentFactory handles components in the spreadsheet. When a custom editor is
+   * retrieved, the editor is taken from the corresponding column.
+   */
+  private final class MyComponentFactory implements SpreadsheetComponentFactory {
```
wow-such-code, line 632: I would rename your implementation to something more meaningful

wow-such-code (changes requested):
This looks very promising and the test implementation works well!

To see if it covers all features needed in our domain, please refactor the sample registration spreadsheet to use your implementation.

user-interface/src/main/java/life/qbic/datamanager/views/general/spreadsheet/ColumnValidator.java
KochTobi: Can be `T`. :+1: This class was originally in the Spreadsheet class. The spreadsheet class provides a typed parameter `T`. Originally `T2` was chosen to avoid conflicts with `T`. As the class `ColumnValidator` moved out of the `Spreadsheet` class, `T` can now be used.

user-interface/src/main/java/life/qbic/datamanager/views/general/spreadsheet/Spreadsheet.java
KochTobi: Suggestion?

KochTobi:
> To see if it covers all features needed in our domain, please refactor the sample registration spreadsheet to use your implementation.

Not in this PR. There is a separate task of changing the dialog.

user-interface/src/main/java/life/qbic/datamanager/views/general/spreadsheet/Spreadsheet.java
wow-such-code: How about DatamanagerSpreadsheetComponentFactory? A bit long, but maybe a subset of those words

wow-such-code (changes requested):
Then I see the card as solved. A last thing, though: please handle the code smells where possible (except in the test class and I guess the "transient" issues, as that seems to be an error sometimes)

wow-such-code (approved):
thank you, good to go!

PR #403, opened 2023-10-17
Reset appdrawer content correctly to avoid node tree exception

Steffengreiner:
**What was changed**
Since the content div is static, it's not possible to add it to different state trees. Trying to do so leads to a Statetree exception. 
Bildschirmfoto 2023-10-17 um 11 12 14

Therefore we need to make sure the content div is reset completely and not added multiple times to the parent container 

**How to reproduce** 
1.) Login to the application and select a project so the drawer is initialized
2.) Copy the URL into a different tab and load the page 
3.) Observe state tree exception since the application tries to add a static content div to multiple instances

### DIFF
--- a/vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
+++ b/vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
@@ -4,6 +4,7 @@ import static org.slf4j.LoggerFactory.getLogger;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEvent;
+import com.vaadin.flow.component.UI;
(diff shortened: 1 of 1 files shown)


vaadinfrontend/src/main/java/life/qbic/datamanager/views/navigation/ProjectSideNavigationComponent.java
```diff
@@ -71,8 +71,8 @@ public ProjectSideNavigationComponent(
     addClassName("project-navigation-drawer");
     this.projectInformationService = projectInformationService;
     this.experimentInformationService = experimentInformationService;
+    resetContent();
```
KochTobi, line 74:
```suggestion
```

KochTobi (dismissed):
We do not modify static information in a non-static constructor.

sven1103 (approved):

PR #407, opened 2023-10-19
Rearrange user module to identity module

sven1103:
The identity-infrastructure module still points to the finances module as transitive dependency from project-management. 

I need to remove the UserJpaRepository dependency on SidRepository, since it pollutes identity with authorization concerns, which we want to leave in the project management domain.

image


identity/src/main/java/life/qbic/identity/application/user/IdentityService.java
```diff
@@ -33,27 +32,15 @@
...
-    var notificationId = notificationService.newNotificationId();
-    var notification =
-        Notification.create(
-            event.getClass().getSimpleName(),
-            event.occurredOn(),
-            notificationId,
-            event);
-    notificationService.send(notification);
-  }
-
```
KochTobi, line 56: I assume this functionality was moved somewhere? Can you point me in the right direction?

KochTobi (changes requested):
One small question

identity/src/main/java/life/qbic/identity/application/user/IdentityService.java
sven1103: it was removed since not used

PR #408, opened 2023-10-19
Use user ID instead of email for authentication

wow-such-code:


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/access/ProjectAccessComponent.java
```diff
@@ -154,28 +155,30 @@ private void loadInformationForProject(ProjectId projectId) {
...
+    }
+    List<String> authorities = projectAccessService.listAuthorities(projectId).stream().distinct()
         .toList();
+    var entries = users.stream().map(userDetail -> {
+      var roles = getProjectRoles(authorities, userDetail);
+      roles = roles.stream()
+          .map(this::formatAuthorityToReadableString)
+          .toList();
+      String fullName = userInformationService.findById(userDetail.getUserId()).get().fullName();
+      return new UserProjectAccess(fullName, userDetail.getEmailAddress(), String.join(", ", roles));
```
sven1103, line 178: Since this is our implementation, why do we need the email address here and further use the concept user name?

sven1103 (changes requested):

sven1103 (changes requested):
Only one minor feedback request, to possibly eliminate the concept 'username'.

sven1103 (approved):
Awesome work!

PR #409, opened 2023-10-19
Refactor project management domain

sven1103:
Before:
image

After:
image


project-management-infrastructure/src/main/java/life/qbic/controlling/infrastructure/sample/SamplePreviewJpaRepository.java
```diff
@@ -98,7 +97,7 @@ public static Specification<SamplePreview> isDistinct() {
     //If no filter was provided return all SamplePreviews
     public static Specification<SamplePreview> isBlank(String filter) {
       return (root, query, builder) -> {
-        if (StringUtil.isBlank(filter)) {
+        if (filter.isBlank()) {
```
KochTobi, line 100: Good spot! However, StringUtil::isBlank checks for null as well according to the JavaDoc.
If the filter is null you would produce a NullPointer with String::isBlank

project-management-infrastructure/src/main/java/life/qbic/controlling/infrastructure/sample/SamplePreviewJpaRepository.java
```diff
@@ -107,7 +106,7 @@ public static Specification<SamplePreview> isBlank(String filter) {
 
     public static Specification<SamplePreview> experimentIdEquals(ExperimentId experimentId) {
       return (root, query, builder) ->
-          StringUtil.isBlank(experimentId.value()) ?
+          experimentId.value().isBlank() ?
```
KochTobi, line 109: As before, you need to check for null here if the value can be null.

KochTobi (changes requested):

KochTobi (approved):
LGTM

PR #412, opened 2023-10-20
Allow tooltips in summary page, add some tooltips

wow-such-code:


sven1103 (approved):
Cool, simple and effective solution

PR #416, opened 2023-10-24
Create SID entry for a new user after registration

wow-such-code:
This fixes an issue where new users could not be added to projects, since the SID entry was not yet created.

### DIFF
--- a/user-interface/src/main/java/life/qbic/datamanager/Application.java
+++ b/user-interface/src/main/java/life/qbic/datamanager/Application.java
@@ -95,6 +95,7 @@ public class Application extends SpringBootServletInitializer implements AppShel
     newPassword.setUseCaseOutput(newPasswordHandler);
   }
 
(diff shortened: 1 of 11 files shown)


user-interface/src/main/java/life/qbic/datamanager/Application.java
```diff
@@ -88,6 +95,23 @@ private static void setupUseCases(ConfigurableApplicationContext context) {
     newPassword.setUseCaseOutput(newPasswordHandler);
   }
 
+  private static MessageSubscriber whenUserRegisteredAddSid(SidRepository sidRepository) {
```
sven1103, line 98: Please add another TODO statement here, so we really address it. The Application.java class must not contain policies in the future to avoid making this class a spaghetti script of initialization.

sven1103 (approved):

PR #421, opened 2023-10-31
Address indexing error within acl deletion when user is removed from a project

Steffengreiner:
**What was changed** 
Revoking access to a project is done via deleting the acl entries within a mutable acl object. However the current implementation leads to an indexing issue since we're deleting entries from a list that is not in sync with the copied aclEntries list. This can be avoided by starting the deletion and iteration from the highest number in the list.


wow-such-code (approved):
wat

PR #423, opened 2023-11-07
Enable user to navigate within the pages of an experiment via navigation stepper

Steffengreiner:
**What was changed**

**1.) Introduce dedicated AppLayout `ExperimentMainLayout` for all pages within an Experiment**
This AppLayout hosts the `ProjectSideNavigationComponent` in its Drawer Section and the `ExperimentNavigationComponent' in its Navbar section. 

**2.) Adapt Routing to the `SampleMainComponent`** 
from `projects/:projectId?/samples`to `projects/:projectId?/experiments/:experimentId?/samples`

Since the component is planned to show the batch information within an experiment instead of all experiment information within a project. This is also necessary so the `ExperimentNavigationComponent` knows the pages of which experiment it should route to. 

**3.) Remove `ProjectNavigationBarComponent`** 
No longer necessary since it's replaced with 'ExperimentNavigationComponent', this also means that the CSS for all main components was adapted to remove the space within the pages CSS layout grid allocated to the `ProjectNavigationBarComponent`. 

**4.) Minor adaptions to the component routing**
We can now expect an `ExperimentId` to be present when navigating between the pages of an experiment, therefore the routing in all components and considering the beforeEnterEvents will throw an exception if no valid` ExperimentId` was required.  

**Technical Limitations and Bugs** 

1.) Active Experiment within the `SideNav` component is not highlighted correctly if the user switches to the `ExperimentInformationPage` within the experiment via the `ExperimentNavigationComponent`: 

https://github.com/qbicsoftware/data-manager-app/assets/29627977/577a3522-fb33-474c-86e9-6b28e2f5252f

This seems to be a [bug](https://github.com/vaadin/flow-components/issues/5227#issuecomment-1766321386) within the SideNavComponent which will hopefully be addressed by Vaadin in the future. 
Sadly currently the SideNav component also does not provide the [API](https://github.com/vaadin/flow-components/issues/5091) to set an active Item directly so we can't do an easy workaround for now. 

2.) Selecting an experiment via the tab within the SampleGrid does not update the active experiment in the SideNavComponent

https://github.com/qbicsoftware/data-manager-app/assets/29627977/6cee2709-9aed-4f3e-b328-84d84a76cc20

This was not implemented since we plan to replace the experiment tabs within the sample grid with tabs showing batch information instead. 

Bildschirmfoto 2023-11-07 um 14 12 43


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/experiments/ExperimentInformationMain.java
```diff
@@ -49,44 +40,26 @@ public class ExperimentInformationMain extends MainComponent implements BeforeEn
...
-    requireNonNull(experimentInformationService);
-    this.projectNavigationBarComponent = projectNavigationBarComponent;
     this.experimentContentComponent = experimentContentComponent;
     this.experimentSupportComponent = experimentSupportComponent;
-    this.projectInformationService = projectInformationService;
-    this.experimentInformationService = experimentInformationService;
-    layoutComponent();
+    addClassName("experiment");
     log.debug(String.format(
-        "New instance for ExperimentInformationMain (#%s) created with ProjectNavigationBar Component (#%s), ExperimentMain component (#%s) and ExperimentSupport component (#%s)",
```
sven1103, line 78: Please use variables to reference class names, since they are mutable and can change. `getClass().getName()` sth like that

sven1103 (approved):
Great job !

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/experiments/ExperimentInformationMain.java
Steffengreiner: Oh good point! I took the liberty and adapted all class instantiation logs in this[commit](https://github.com/qbicsoftware/data-manager-app/pull/423/commits/fe3adef6dab78a40461cb086316da3a74a76f7f7) in a similiar way.

sven1103 (approved):

PR #425, opened 2023-11-13
Release/0.30.0

Steffengreiner:


KochTobi:
Some things I noticed:
* Go to samples notification no longer links to the correct page after group editing.
* Download Metadata button without any functionality in step _Register Sample Batch_
* In _Register Sample Batch_ there are 2 buttons with different text but the same function. `Register` on top and `Register batch` in the middle.
* Column width in the sample grid on _Register Sample Batch_ page is to small and no longer fits its content.

KochTobi (approved):
LGTM

PR #427, opened 2023-11-14
Ensure that notification anchor after group registration always points to SampleInformationMain

Steffengreiner:
**What was changed**

After the routing update introduced with the experimentnavigation component, the hardcoded URL in the notification was no t working anymore. 
To futureproof this the String was replaced with the route retrieved from the class in question as suggested by the [vaadin docs](https://vaadin.com/docs/latest/routing/retrieving-routes)


wow-such-code (approved):
nice work

PR #431, opened 2023-11-20
Allow Users to create projects within a multistep dialog 

Steffengreiner:
**What was changed**
Entire rewrite of the project creation dialog. 

1.) Introduction of a generic stepper component which can be reused as necessary.
2.) Project Creation is seperated into different layouts within a dialog which are swapped as the user goes along the project creation process.
3.) For this a clear seperation between the Logic, Layouts, DTOs and Events of the project creation and project editing use cases, since their UI is now completely different


KochTobi (changes requested):
Dear @Steffengreiner 
I did not have a look at the code yet. Running your branch, I noticed a couple of things I am uncertain about:

1. When navigating through the dialog, I cannot navigate back when information in the current step is not valid. I think this can not be the desired behaviour. When I click next and notice I forgot to add something in an earlier step, I should be able to simply go back and change it. Do you agree? What do you think?
2. Entering an empty experiment name is possible by simply using whitespace.
3. The styling of the edit dialog is broken now. Can you check to verify that you did not change the styling for the edit dialog?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/ExperimentalInformationLayout.java
```diff
@@ -0,0 +1,268 @@
...
+    analyteBox.setItemLabelGenerator(Analyte::label);
+    analyteBox.setPlaceholder("Please select one or more analytes for your samples");
+    experimentalInformationBinder.forField(analyteBox)
+        .asRequired("Please select at least one analyte")
+        .bind(experimentalInformation -> new HashSet<>(experimentalInformation.getAnalytes()),
+            ExperimentalInformation::setAnalytes);
+    add(speciesBox, specimenBox, analyteBox);
+  }
+
+  private <T> void initComboBoxWithDatasource(MultiSelectComboBox<T> box, List<Ontology> ontologies,
```
wow-such-code, line 106: please replace with new helper class OntologyFilterConnector.initComboBoxWithOntologyDatasource(...)
(I kept box.addClassNames(CHIP_BADGE, WIDTH_INPUT) here due to the specific references)

wow-such-code (changes requested):
Nice work, the points by @KochTobi have been addressed as well, from what I can tell.

One small thing: please adapt the ontology comboboxes with the newly merged component - this will also add the helper text that is now missing from experiment creation.

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/ExperimentalInformationLayout.java
Steffengreiner: Thanks for the feedback, it is done 👍

wow-such-code (approved):

PR #433, opened 2023-11-22
Exception when pasting into the spreadsheet during sample registration

KochTobi:
When pasting more columns or rows than present, the spreadsheet broke. This is caught and all input exceeding the spreadsheet dimension is ignored.


Steffengreiner (approved):
LGTM.  🚀
However currently the user does not know why the paste didn't work. In future we could maybe introduce a notification that informs him about pasting too many rows 🤔

Steffengreiner (approved):
Still works 🚢 
However the content will not be pasted if the user pastes content exceeding the row limit and the user is not informed why the content was not pasted 🤔

wow-such-code (approved):
lgtm

PR #438, opened 2023-11-27
Enable upload of service purchase information

sven1103:
Introduces an offer upload dialog for signed/unsigned offers:

image


KochTobi:
```
org.apache.tomcat.util.http.fileupload.impl.SizeLimitExceededException: the request was rejected because its size (11255463) exceeds the configured maximum (10485760)
```
When trying to upload a file of 11.3 MB. The UI states a maximum file size of 14 MB.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/purchase/ServicePurchase.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * <b>Service Purchase</b>
+ *
+ * <p>A service purchase deliverable for a project</p>
+ *
+ * @since 1.0.0
+ */
+@Entity
+@Table(name = "service_purchase")
+public class ServicePurchase {
```
KochTobi, line 23: I am missing equals and hashCode methods for this class.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/purchase/Offer.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * <b>Offer</b>
+ *
+ * An offer in the context of project management, not finance and accounting.
+ *
+ * @since 1.0.0
+ */
+@Entity(name = "purchase_offer")
+public class Offer {
```
KochTobi, line 21: Nice. From the equals method I gather that this is a value object.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/purchase/Offer.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  private Long id;
+
+  private boolean signed;
+
+  private String fileName;
+
+  @Lob
```
KochTobi, line 31: As this is a large object, maybe we should fetch it lazily as the JavaDoc implies.
```java
@Basic(fetch=LAZY)
```

project-management/src/main/java/life/qbic/projectmanagement/application/purchase/ProjectPurchaseService.java
```diff
@@ -0,0 +1,54 @@
...
+    var purchaseDate = Instant.now();
+
+    ServicePurchase purchase = ServicePurchase.create(projectReference, purchaseDate, theOffer);
+
+    try {
+      storage.storePurchase(purchase);
+    } catch (PurchaseStoreException e) {
+      log.error("Purchase storage failed", e);
+      throw new ApplicationException("Saving the offer failed");
+    }
```
KochTobi, line 52: Why don't you let it throw directly? If you want to add additional debug context you could also throw `ApplicationException.wrapping("Could not store purchase " + purchase , e)`

project-management/src/main/java/life/qbic/projectmanagement/application/purchase/ProjectPurchaseService.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * <b>Project Purchase Service</b>
+ * <p>
+ * A service that enables actions on project purchase services.
+ *
+ * @since 1.0.0
+ */
+@Service
+public class ProjectPurchaseService {
```
KochTobi, line 23: I like that you put this into its own service :+1:

project-management/src/main/java/life/qbic/projectmanagement/application/purchase/OfferDTO.java
```diff
@@ -0,0 +1,12 @@
+package life.qbic.projectmanagement.application.purchase;
+
+/**
+ * <b><record short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
+public record OfferDTO(boolean signed, String fileName, byte[] content) {
```
KochTobi, line 10: When using arrays you have to overwrite equals and hashcode as by default array content is not compared. This only compares the arrays and might lead to misleading equals calls.

project-management/src/main/java/life/qbic/projectmanagement/application/api/PurchaseStoreException.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * <b>Purchase Store Exception</b>
+ * <p>
+ * Exception that shall be thrown if a
+ * {@link life.qbic.projectmanagement.domain.model.project.purchase.ServicePurchase} item cannot be
+ * stored in the persistence storage implementation.
+ *
+ * @since 1.0.0
+ */
+public class PurchaseStoreException extends RuntimeException {
```
KochTobi, line 14: As you are only catching this exception to log and rethrow you could just not catch the exception and let the global exception handler deal with it. You would not need this special exception then.

project-management/src/main/java/life/qbic/projectmanagement/application/api/ProjectPurchaseStorage.java
```diff
@@ -0,0 +1,23 @@
...
+public interface ProjectPurchaseStorage {
+
+  /**
+   * Stores a {@link ServicePurchase} item persistently.
+   *
+   * @param purchase the service purchase item to store
+   * @throws PurchaseStoreException is thrown if the storage fails
+   * @since 1.0.0
+   */
+  void storePurchase(ServicePurchase purchase) throws PurchaseStoreException;
```
KochTobi, line 21: you could just fail with any exception as you do not really handle the exception thrown.

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/project/purchase/PurchaseStore.java
```diff
@@ -0,0 +1,40 @@
...
+  @Autowired
+  public PurchaseStore(ProjectPurchaseJpa persistenceStore) {
+    this.persistenceStore = persistenceStore;
+  }
+
+  @Override
+  public void storePurchase(ServicePurchase purchase) throws PurchaseStoreException {
+    try {
+      persistenceStore.save(purchase);
+    } catch (Exception e) {
```
KochTobi, line 35: Please never catch `Exception` itself. 
```suggestion
    } catch (RuntimeException e) {
```
Why not let it fly instead?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectContentComponent.java
```diff
@@ -29,13 +35,39 @@ public class ProjectContentComponent extends Div {
...
+  private final ProjectPurchaseService projectPurchaseService;
+  private Context context;
+
   public ProjectContentComponent(
-      @Autowired ProjectDetailsComponent projectDetailsComponent) {
+      @Autowired ProjectDetailsComponent projectDetailsComponent,
+      @Autowired ProjectPurchaseService projectPurchaseService) {
     Objects.requireNonNull(projectDetailsComponent);
+    this.projectPurchaseService = Objects.requireNonNull(projectPurchaseService);
+    UploadPurchaseDialog uploadPurchaseDialog = new UploadPurchaseDialog();
```
KochTobi, line 46: NIT: I'd suggest to re-create the dialog every time it is opened. This way you can be certain that it is refreshed.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectContentComponent.java
```diff
@@ -29,13 +35,39 @@ public class ProjectContentComponent extends Div {
...
+    uploadPurchaseDialog.setOpened(false);
+    add(uploadPurchaseDialog);
+    Button uploadOffer = new Button("Upload offer");
+    uploadOffer.addClickListener(listener -> uploadPurchaseDialog.open());
+    uploadPurchaseDialog.subscribe(
```
KochTobi, line 51: please rename the method to `addConfirmListener` to be in sync with the other dialogs.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectContentComponent.java
```diff
@@ -29,13 +35,39 @@ public class ProjectContentComponent extends Div {
...
+        uploadPurchaseDialogConfirmEvent -> {
+          var offers = uploadPurchaseDialog.purchaseItems();
+          uploadPurchaseDialog.close();
```
KochTobi, line 54: You probably want to close the dialog only after the purchase item is added so that the user can change input if something goes wrong?

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/EditableMultiFileMemoryBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * <b>Editable multi-file memory buffer</b>
+ * <p>
+ * An improved {@link MultiFileReceiver} enables removing cached file content, which Vaadin's
+ * {@link com.vaadin.flow.component.upload.receivers.MultiFileMemoryBuffer} does not provide.
+ *
+ * @since 1.0.0
+ */
+public class EditableMultiFileMemoryBuffer implements MultiFileReceiver {
```
KochTobi, line 22: very nice class :+1:

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/PurchaseItem.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.Serial;
+import java.util.Objects;
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
```
KochTobi, line 16: some javadoc would be nice :smile:

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  @Override
+  protected void onConfirmClicked(ClickEvent<Button> clickEvent) {
+    fireConfirmEvent();
+  }
+
+  @Override
+  protected void onCancelClicked(ClickEvent<Button> clickEvent) {
+    this.close();
```
KochTobi, line 113: should be handled by the calling component in my opinion. This is how it is done with other dialogs.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  private void fireConfirmEvent() {
```
KochTobi, line 116: you do not need this custom implementation.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  private void processClientFileRemoveEvent(DomEvent event) {
+    JsonObject jsonObject = event.getEventData();
+    var fileName = jsonObject.getString(VAADIN_FILENAME_EVENT);
+    removeFile(fileName);
+  }
+
+  private void removeFile(String fileName) {
+    removeFileFromBuffer(fileName);
+    removeFileFromDisplay(fileName);
+    toggleFileSectionIfEmpty();
```
KochTobi, line 130: I think the toggle should go one level up to `processClientFileRemoveEvent`

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  protected void onCancelClicked(ClickEvent<Button> clickEvent) {
+    this.close();
+  }
+
+  private void fireConfirmEvent() {
+    var confirmEvent = new ConfirmEvent<>(this, true);
+    confirmListeners.forEach(listener -> listener.onComponentEvent(confirmEvent));
+  }
+
+  private void processClientFileRemoveEvent(DomEvent event) {
```
KochTobi, line 121: NIT: renaming of the method to `onClientFileRemoveEvent`

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  private void removeFile(String fileName) {
+    removeFileFromBuffer(fileName);
+    removeFileFromDisplay(fileName);
+    toggleFileSectionIfEmpty();
+  }
+
+  private void toggleFileSectionIfEmpty() {
+    if (uploadedPurchaseItems.getChildren().toList().isEmpty()) {
+      uploadedPurchaseItems.addClassName("hidden");
+      disclaimer.addClassName("hidden");
```
KochTobi, line 136: you can use vaadins `setVisible(false)`. This way you can get rid of the if-else as well.
Visibility in vaadin is handled special as some components are removed from the Tree completely on `setVisible(false)`. I do not know whether setting the visibility hidden CSS property has some side-effects in the framework.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+    } else {
+      uploadedPurchaseItems.removeClassName("hidden");
+      disclaimer.removeClassName("hidden");
```
KochTobi, line 139: you can use vaadins `setVisible(true)`.
This way you can get rid of the if-else as well.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  private void removeFileFromBuffer(String fileName) {
+    multiFileMemoryBuffer.remove(fileName);
+  }
+
+  private void removeFileFromDisplay(String fileName) {
+    var result = purchaseItemsCache.stream().filter(purchaseItem -> purchaseItem.fileName().equals(
+        fileName)).findAny();
+    result.ifPresent(purchaseItem -> {
```
KochTobi, line 150: The variable `result` does not add to the readability. You can either rename the variable to reflect the stored value or remove it and call `Optional::ifPresent` directly.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+      purchaseItemsCache.remove(purchaseItem);
+      uploadedPurchaseItems.remove(purchaseItem);
+    });
+  }
+
+  public List<OfferDTO> purchaseItems() {
+    return purchaseItemsCache.stream().map(this::convertToOffer).toList();
```
KochTobi, line 157: nice

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private OfferDTO convertToOffer(PurchaseItem purchaseItem) {
+    try {
+      var fileName = purchaseItem.fileName();
+      var content = multiFileMemoryBuffer.inputStream(fileName)
+          .orElse(new ByteArrayInputStream(new byte[]{})).readAllBytes();
+      return new OfferDTO(purchaseItem.isSigned(), fileName, content);
+    } catch (IOException e) {
+      log.error("Failed to read uploaded purchase item content", e);
+      throw new ApplicationException("Failed to read offer content");
```
KochTobi, line 168: you can

```java
throw new ApplicationException.wrapping("Could not upload " + purchasedItem.fileName(), e);
```

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+  }
+
+  public void subscribe(ComponentEventListener<ConfirmEvent<UploadPurchaseDialog>> listener) {
```
KochTobi, line 172: please rename to `addConfirmListener` to match our general listener method style.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+    this.confirmListeners.add(listener);
```
KochTobi, line 173: you can replace this by the framework calling `addListener(ConfirmEvent.class, listener)` see other dialogs for reference.

KochTobi (changes requested):
Hi Sven,
Very impressive implementation. I noted a couple of minor things I would like you to address. Feel free to ignore `NIT` comments.
Looking forward to seeing your dialog in the development branch :+1:

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectContentComponent.java
sven1103: The saving of offer items is blocking, hence the dialog remains open. This can lead to multiple uploads, if the user hits the save button multiple times.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/purchase/Offer.java
sven1103: exactly

project-management/src/main/java/life/qbic/projectmanagement/application/api/PurchaseStoreException.java
sven1103: I object to promote explicit language against general technical ones :)

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectContentComponent.java
sven1103: agree, but since this is not the final implementation and just a showcase, this can be addressed in the final implementation

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
```diff
@@ -0,0 +1,193 @@
...
+  private EditableMultiFileMemoryBuffer multiFileMemoryBuffer;
+  private Div uploadedPurchaseItems;
+
+  private List<PurchaseItem> purchaseItemsCache = new ArrayList<>();
+
+  private Div disclaimer;
+
+  private Div titleBox;
+
+  private List<ComponentEventListener<ConfirmEvent<UploadPurchaseDialog>>> confirmListeners = new ArrayList<>();
```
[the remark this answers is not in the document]
sven1103: how?

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
sven1103: I strongly object, since the abstract parent class method needs to be overridden. An empty code block implementation is bad practise. Either we remove this method from the abstract class to shift the responsibility to the client or we implement it. 
However I would argue that a cancel event is so trivial and the expected behaviour is that the dialog is closed.

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
sven1103: Thank you! I simplified the method block

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
sven1103: agreed and changed

user-interface/src/main/java/life/qbic/datamanager/views/projects/purchase/UploadPurchaseDialog.java
sven1103: does not work for me :/

PR #439, opened 2023-11-27
Mock openbis connector 

KochTobi:


KochTobi:
Just noticed you can define conditional beans. I will change this.

PR #441, opened 2023-11-28
Propagate batch registration dialog opening from notification in sampledetails component correctly

Steffengreiner:


KochTobi (approved):
LGTM

PR #444, opened 2023-11-29
Small UI and usability changes for ontology term selection

wow-such-code:
* Do not show terms until user types at least two letters
* State this in helper text
* Return only 50 results per search to prevent scrolling lag


Steffengreiner (approved):
LGTM 🚀

PR #449, opened 2023-12-04
Feature/dm 1072/store full ontology class

wow-such-code:
* Creates DTO Object to interchange concrete Ontology Term information between Sample and Experiment objects in the database and the views
* When an experiment is created or updated, OntologyClasses from the lookup table are transformed into the DTO
* Samples or experiment objects now store the DTO as JSON in the database
* In most cases, only the label is shown for now, but methods to access additional information have been added
* In the Sample Grid, some additional information about the selected terms is shown in a tooltip

The following columns in the database need to be updated (change VARCHAR type to TEXT):
experiments_datamanager_species.species
experiments_datamanager_specimens.specimens
experiments_datamanager_analytes.analytes
sample.analyte
sample.species
sample.specimen


project-management/src/main/java/life/qbic/projectmanagement/domain/model/experiment/vocabulary/OntologyClassDTO.java
```diff
@@ -0,0 +1,119 @@
+package life.qbic.projectmanagement.domain.model.experiment.vocabulary;
+
+import java.util.Objects;
+import life.qbic.projectmanagement.application.OntologyClassEntity;
+
+public class OntologyClassDTO {
+
+  String ontology;
+
```
Steffengreiner, line 9: Since this is a key DTO in our application some JD would be good to explain it's usage and context 🥇

user-interface/types.d.ts
```diff
@@ -0,0 +1,10 @@
+// This TypeScript modules definition file is generated by vaadin-maven-plugin.
+// You can not directly import your different static files into TypeScript,
+// This is needed for TypeScript compiler to declare and export as a TypeScript module.
+// It is recommended to commit this file to the VCS.
+// You might want to change the configurations to fit your preferences
+declare module '*.css' {
+  import { CSSResultGroup } from 'lit';
+  const content: CSSResultGroup;
+  export default content;
+}
```
Steffengreiner, line 10: This should not be pushed since it's locally rebuilt every time by vaadin

user-interface/vite.config.ts
```diff
@@ -0,0 +1,9 @@
+import { UserConfigFn } from 'vite';
+import { overrideVaadinConfig } from './vite.generated';
+
+const customConfig: UserConfigFn = (env) => ({
+  // Here you can add custom Vite parameters
+  // https://vitejs.dev/config/
+});
+
```
Steffengreiner, line 8: This should not be pushed since it's locally rebuilt every time by vaadin

Steffengreiner (changes requested):
Heya @wow-such-code awesome work! However i found some minor issues and have questions concerning the tooltip generation: 

1.) All Projects except `"taking the whole ontology"` lead to the error 500 exception page with the following message:
`life.qbic.datamanager.exceptionhandling.routing.exception.ExceptionErrorPage
Illegal pop() with non-matching JdbcValuesSourceProcessingState`
If this is unavoidable, I'd argue the older projects should be deleted.

2.) IMO the content shown within a tooltip is the responsibility of the UI not the responsibility of the samplepreview

3.) The Tooltips are very long which is not recommended according to the best practices by [vaadin ](https://vaadin.com/docs/latest/components/tooltip). 
Bildschirmfoto 2023-12-05 um 15 34 50
An alternative could be replacing the string value within the grid with a custom renderer showing the Onthology Icon/Id and label? 

4.) Additionally they are not consistently employed within the application:
Bildschirmfoto 2023-12-05 um 15 34 13
Bildschirmfoto 2023-12-05 um 15 34 02

wow-such-code:
Thank you for the quick review!

> Heya @wow-such-code awesome work! However i found some minor issues and have questions concerning the tooltip generation:
> 
> 1.) All Projects except `"taking the whole ontology"` lead to the error 500 exception page with the following message: `life.qbic.datamanager.exceptionhandling.routing.exception.ExceptionErrorPage Illegal pop() with non-matching JdbcValuesSourceProcessingState` If this is unavoidable, I'd argue the older projects should be deleted.

This is unavoidable as the respective fields are now parsed as JSON. I agree at least the experiments should be deleted - but I think this can wait until the merge OR even until deployment - in the meantime people might be testing/working with the old projects.

> 2.) IMO the content shown within a tooltip is the responsibility of the UI not the responsibility of the samplepreview

Good point! I will change it.

> 3.) The Tooltips are very long which is not recommended according to the best practices by [vaadin ](https://vaadin.com/docs/latest/components/tooltip). An alternative could be replacing the string value within the grid with a custom renderer showing the Onthology Icon/Id and label?

I will look into that!

> 4.) Additionally they are not consistently employed within the application(The tags here don't show a tooltip)

True, changes to experiments will be done in another PR that will also focus on Ontology Term selection.

> 5.) Finally the search functionality now searches for information not shown in the grid since it looks in the entire ontologyDTO while the grid only shows the label which is confusing to me

Also a good point, I will see if I can search in the label substring (or whatever we will end up using in the view in the end).

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/sample/SamplePreviewJpaRepository.java
```diff
@@ -149,19 +149,23 @@ public static Specification<SamplePreview> conditionContains(String filter) {
       };
     }
 
+    private static Specification<SamplePreview> ontologyColumnContains(String col, String filter) {
+      return (root, query, builder) -> {
+        Expression<String> expression = root.get(col).as(String.class);
+        return builder.like(expression, "%label%"+filter+"%name%");
+      };
+    }
```
Steffengreiner, line 157: While this is an awesome quick fix, i fear that this hardcoded search gets lost when we want to adapt it later. Therefore i looked into how this could be achieved and would suggest the following:

```suggestion
    private static Specification<SamplePreview> ontologyColumnContains(String col, String filter) {
      return (root, query, builder) -> {
        /*We're currently only interested in the label within the json file*/
        Expression<String> function =  builder.function("JSON_EXTRACT", String.class, root.get(col), builder.literal("$.label"));
        return builder.like(function,
            "%" + filter + "%");
    }
```

Steffengreiner (approved):
I proposed a JPA based solution on querying the JSON property which i find a better solution. (Also check the sonarcloud errors please)
 However I'm open for discussion 🤔

Steffengreiner (approved):
Send it 🚢

PR #458, opened 2024-01-10
Project creation dialog shows descriptions for each layout and adheres to sizing scheme for our dialogs

Steffengreiner:
**What was changed**
1.) Instead of dummy description texts, the layouts now show the descriptions defined in the figma prototype
2.) The header of the person responsible was renamed
3.) CSS style changes so the dialog is sized as intended 

**Prototype**
The figma prototype can be found [here](https://www.figma.com/file/mYcW5viu5GpPZ852VHYBrV/Start-Page?type=design&node-id=4788-99390&mode=dev)


wow-such-code (approved):
lookin' good

PR #459, opened 2024-01-10
sort samples in edit spreadsheet like in the sample table

wow-such-code:
Samples in both views are now sorted by the sample code.


Steffengreiner (approved):
Nicely done 🥇

PR #465, opened 2024-01-17
Fix sample editing and project search in openBIS

wow-such-code:
* Fixes and enables sample metadata update in openBIS, which was caused by an immutable list
* Fixes project search in openBIS, which was not searching for the project code
* Fixes a typo in Analysis Type

### DIFF
--- a/project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/sample/openbis/OpenbisConnector.java
+++ b/project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/sample/openbis/OpenbisConnector.java
@@ -45,7 +45,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import life.qbic.logging.api.Logger;
 import life.qbic.openbis.openbisclient.OpenBisClient;
 import life.qbic.projectmanagement.domain.model.project.Project;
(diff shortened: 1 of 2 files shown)


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/sample/openbis/OpenbisConnector.java
```diff
@@ -437,7 +436,7 @@ public void delete(ProjectCode projectCode) {
 
   @Override
   public boolean projectExists(ProjectCode projectCode) {
-    return !searchProjectsByCode(projectCode.toString()).isEmpty();
+    return !searchProjectsByCode(projectCode.value()).isEmpty();
```
KochTobi, line 439: good catch

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/sample/openbis/OpenbisConnector.java
```diff
@@ -314,10 +315,7 @@ public void updateAll(
       Collection<life.qbic.projectmanagement.domain.model.sample.Sample> samples)
       throws SampleNotUpdatedException {
     try {
-      /*FixMe Throws org.springframework.remoting.RemoteAccessException: Could not access HTTP invoker remote service
-          and invalid stream header: 3C68746D
-      */
-      // updateOpenbisSamples(convertSamplesToSampleUpdates(samples));
+      updateOpenbisSamples(convertSamplesToSampleUpdates(samples));
```
KochTobi, line 318: as it is `ch.ethz.sis.openbis.generic.asapi.v3.dto.sample.update.UpdateSamplesOperation` that expects a mutable list, maybe add a comment to `updateOpenbisSamples` and ensure that it is a mutable list in `updateOpenbisSample` as well. This does remove the need of having an modifiable `SampleUpdate`s list.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/sample/AnalysisMethod.java
```diff
@@ -60,7 +60,7 @@ public enum AnalysisMethod {
   Proteomics related analysis methods
    */
 
-  PROTEOMICS("PROTEOMIICS", "Proteomics", ""),
+  PROTEOMICS("PROTEOMICS", "Proteomics", ""),
```
KochTobi, line 63: good catch

KochTobi (changes requested):
Nice job detecting the issue with list mutability. I suggest moving a conversion like `new ArrayList(sampleUpdates)` into the method that calls the openbis operation. This would make our code independent of their expectation here and we could freely use `toList`.

KochTobi (approved):

PR #466, opened 2024-01-19
only load unique ontology terms

wow-such-code:
* queries to the ontology database (via filter input) now return a list of unique values, fixing an issue with the displayed components
* this is only an additional safeguard, as duplicates will be removed from the table and the scripts to fill the table adapted

See also: https://github.com/qbicsoftware/dm-onthology-terms/pull/2


KochTobi (approved):
LGTM

PR #468, opened 2024-01-22
Show number of samples, creation- and modificationdate in batchdetailscomponent

Steffengreiner:
**What was changed** 
This PR introduces the logic to show and adjust the number of samples, the modification date and creation date for a batch in the batchgrid. 

**Technical Issues**
1.) To show the correct number of samples after a batch editing has taken place, a seperate "SampleDeletion" domain event has to be published and consumed as a job by jobrunr correctly.
This works locally, however if the job event is taken by qpylon then the job can't be run since the necessary methods are only implemented in this branch

2.) Because of the "lag" introduced through the job consumption via jobrunr during sample deletion or creation, the sample number is only adjusted if the page is reloaded after the job has run in jobrunr successfully.

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/policy/directive/DeleteSampleFromBatch.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/policy/directive/DeleteSampleFromBatch.java
@@ -14,7 +14,7 @@ import org.jobrunr.scheduling.JobScheduler;
 import org.springframework.stereotype.Component;
 
 /**
- * <b>Directive: Delete Sample to Batch</b>
+ * <b>Directive: Delete Sample from Batch</b>
  * <p>
  * After a sample has been deleted, we need to update the batch and remove
  * the sample reference of the deleted sample.
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/policy/directive/DeleteSampleFromBatch.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/policy/directive/DeleteSampleFromBatch.java
@@ -1,0 +1,56 @@
+package life.qbic.projectmanagement.application.policy.directive;
+
+import static life.qbic.logging.service.LoggerFactory.logger;
+
+import life.qbic.domain.concepts.DomainEvent;
+import life.qbic.domain.concepts.DomainEventSubscriber;
+import life.qbic.logging.api.Logger;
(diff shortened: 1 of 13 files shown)


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -193,10 +193,8 @@ public void setContext(Context context) {
     if (context.projectId().isEmpty()) {
       throw new ApplicationException("no project id in context " + context);
     }
-
     this.context = context;
-    ExperimentId experimentId = context.experimentId()
-        .orElseThrow(() -> new ApplicationException("no experiment id in context " + context));
+    ExperimentId experimentId = context.experimentId().get();
```
wow-such-code, line 197: what is the idea behind this change?

project-management/src/main/java/life/qbic/projectmanagement/application/policy/directive/DeleteSampleFromBatch.java
```diff
@@ -0,0 +1,56 @@
...
+import life.qbic.projectmanagement.application.batch.BatchRegistrationService;
+import life.qbic.projectmanagement.domain.model.batch.BatchId;
+import life.qbic.projectmanagement.domain.model.sample.SampleId;
+import life.qbic.projectmanagement.domain.model.sample.event.SampleDeleted;
+import org.jobrunr.jobs.annotations.Job;
+import org.jobrunr.scheduling.JobScheduler;
+import org.springframework.stereotype.Component;
+
+/**
+ * <b>Directive: Delete Sample to Batch</b>
```
wow-such-code, line 17:
```suggestion
 * <b>Directive: Delete Sample from Batch</b>
```

wow-such-code (changes requested):
lookin' good, just a small documentation fix and a question

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
Steffengreiner: Since we already throw an exception in the beginning of the setContext method if no experimentId was provided it's unnecessary to check and do so again here

wow-such-code (approved):

PR #469, opened 2024-01-23
add optional sample property 'organism id'

wow-such-code:
Adds the property 'Organism Id' to Sample objects, Sample creation and Sample editing.

This optional property allows to discriminate between samples that are from the same patient or model organism and samples that are from different individuals.

Since the property is optional, these changes are backwards compatible and older batches/samples can still be displayed/edited. However, the following column has to be added to the **sample** table:

`organism_id varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_general_ci DEFAULT NULL`


Steffengreiner (changes requested):
Heya @wow-such-code , the code itself is excellent 👌.
However, i'd request to fix the typo of the branch name, otherwise the github actions cannot generate the changelog correctly. 
(feautre/ should be feature/)

PR #473, opened 2024-01-30
Release 0.34.0

Steffengreiner:


KochTobi (changes requested):
The number of samples is not adjusted after sample batches are edited. Is this the expected and desired behaviour?

KochTobi (approved):

PR #478, opened 2024-02-01
Fixes JPA experiment column creation 

KochTobi:
This PR fixes the column creation for the experiment table.


Steffengreiner (approved):
LGTM 🚀

PR #480, opened 2024-02-05
Allow download of sample metadata sheet

wow-such-code:
* enables the download of a tab-separated-value spreadsheet for all samples of an experiment
* the spreadsheet also contains the batch name for each sample
* units of experimental variables are shown in the header to make data-rows more machine-usable
* introduces a DownloadContentProvider interface, a DownloadProvider using it to link downloads in the UI, as well as the content provider implementation for Sample metadata
* introduces a reusable TSVBuilder that allows creation of a TSV String representation from generic objects
* replaces experimental variable Set with a List to keep order of variables, as the Set methods were not used anyway

image


sven1103:
Hi @wow-such-code 

i was wondering from where this requirement came from:

- `replaces experimental variable set with a list to keep order of variables. the check for distinct variables is performed on creation of the condition object`

Not from me so far, so I assume you have some knowledge worth to share.

sven1103:
Looking over the code, I would like to have a more extendable and reusable solution of the implementation. We already know that we have downloads in several places in the app.

@KochTobi @Steffengreiner adding you here, because i think we need to resolve a good strategy. I envision the Download being sth that can be configured with e.g. "ContentProvider", that can multiple implementations, based on what needs to be done. For example we can have a SampleMetadataContentProvider that is set up with the sample metadata content and offers a common API with the ContentProvider Interface to enable the Download context to retrieve the content as e.g. ByteStream or whatever

wow-such-code:
> Hi @wow-such-code 
> 
> i was wondering from where this requirement came from:
> 
> - `replaces experimental variable set with a list to keep order of variables. the check for distinct variables is performed on creation of the condition object`
> 
> Not from me so far, so I assume you have some knowledge worth to share. 

Hi Sven, not a requirement, but something that was changed after some discussion to make implementation cleaner.
Header and data of the sheet have to be in the same order, thus the use of a list. The set functionality was never used in any way, either, despite the comment.

sven1103 (approved):
Bravo :clap:

PR #481, opened 2024-02-06
Show existing groups in add dialog

wow-such-code:
* Introduces the possibility to create an ExperimentalGroupsDialog with existing groups that cannot be edited
* This makes it easier for users to add unique experimental groups (not accidentally adding the same again)
* For the Add Groups use case, only new groups are returned
* For the Edit Groups use case, everything is returned as before, as all parts of the dialog are in "enabled" mode


Steffengreiner (changes requested):
Heya @wow-such-code! Very nice implementation! 
I'd request that the delete button either works or is not shown for the non-editable groups, since otherwise this could confuse the user. 
https://github.com/qbicsoftware/data-manager-app/assets/29627977/95c191d8-8da9-4636-8c45-a184bb8f22cf

Steffengreiner (approved):
Nicely Done! 🥳 
(Although the components are now a little misaligned, since the button space was taken by the experiment group. This will be fixed however with the big dialog overhaul PR 🤔 )

PR #482, opened 2024-02-06
Register measurements

sven1103:


sven1103:
I think we should also include in the validation, if an sample code provided is part of an experiment

sven1103 (commented):
Some smaller comments, else looks really good

PR #493, opened 2024-02-16
User is able to access Measurement page and download their template of interest

Steffengreiner:
**What was changed**

1.) Introduce MeasurementMain Component to which the user can navigate via the built in experimentnavigationcomponent. 
The main page is also accessible via the URL:
<host>/projects/<projectId>/experiments/<experimentId>/measurements

2.) Introduce MeasurementTemplateList Component which provides all measurementTemplates stored in the resource path of the application (Provided by @sven1103 in #484)


sven1103 (changes requested):
Dear @Steffengreiner , please enable the user to access the measurement page without typing in the url path.

sven1103 (approved):
The rendering of the templates is an improvement now to the file name display.

PR #495, opened 2024-02-19
Set spring properties to allow 16MB (Medium BLOB size) uploads

Steffengreiner:
**What was changed**
1.) application.properties settings are correctly set to allow file size uploads up to MEDIUMBLOB size.
2.) Apparently the file_content column in the quality_control_upload table was also incorrectly set to BLOB 
3.) Increased Max allowed packet variable to 50MB 

If files exceed the max allowed packet variable the following exception is thrown

`Caused by: org.springframework.orm.jpa.JpaSystemException: could not execute statement [Packet for query is too large (DOUBLEFILESIZE > MEDIUMBLOBSIZE). 
You can change this value on the server by setting the 'max_allowed_packet' variable.]`
*


wow-such-code (changes requested):
Thank you for tackling this. Since we didn't find a solution to the double-sized stream, I suggest to set MAX_FILE_SIZE_BYTES to 7 MiB for now in both uploads to avoid confusion for the user.

wow-such-code (approved):

PR #497, opened 2024-02-22
hide offer search from anyone but admin and project-manager

wow-such-code:
* when adding a project, only adds offer search field to the interface if the user has admin or project_manager role
* disables search methods for offers for normal users (maybe unnecessary?)

### DIFF
--- a/user-interface/src/main/java/life/qbic/datamanager/views/projects/overview/ProjectOverviewMain.java
+++ b/user-interface/src/main/java/life/qbic/datamanager/views/projects/overview/ProjectOverviewMain.java
@@ -6,6 +6,9 @@ import com.vaadin.flow.router.PageTitle;
 import com.vaadin.flow.router.Route;
 import jakarta.annotation.security.PermitAll;
 import java.io.Serial;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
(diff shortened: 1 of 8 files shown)


user-interface/src/main/java/life/qbic/datamanager/views/projects/overview/ProjectOverviewMain.java
```diff
@@ -75,6 +82,12 @@ public ProjectOverviewMain(@Autowired ProjectCollectionComponent projectCollecti
         System.identityHashCode(projectCollectionComponent)));
   }
 
+  private boolean isOfferSearchAllowed() {
+    Set<String> allowedRoles = new HashSet<>(Arrays.asList("ROLE_ADMIN", "ROLE_PROJECT_MANAGER"));
+    return SecurityContextHolder.getContext().getAuthentication().getAuthorities().stream()
+        .anyMatch(r -> allowedRoles.contains(r.getAuthority()));
+  }
```
Steffengreiner, line 89: NIT: 
Another alternative could be to access and store the user authorities before the user enters the main page via the BeforeEnterEvent and then check these stored events for the right authorities.

Steffengreiner (approved):
Heya @wow-such-code! Nicely done 👍

PR #498, opened 2024-02-22
allow users to edit access to their projects

wow-such-code:
* allows users access to the "Project Access Management" page for projects they have created
* they can add or remove other users
* other users can edit the respective projects, but cannot access the "Project Access Management" page themselves


user-interface/src/main/java/life/qbic/datamanager/security/UserPermissionsImpl.java
```diff
@@ -12,6 +12,8 @@ public class UserPermissionsImpl implements UserPermissions {
 
   final AclPermissionEvaluator aclPermissionEvaluator;
 
+  private static final String projectTargetType = "life.qbic.projectmanagement.domain.model.project.Project";
+
```
Steffengreiner, line 16: You can replace the hardcoded string name by using one of the java name methods for the class of interest outlined here:
https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

Steffengreiner (approved):
LGMT! I made a minor point on how you could improve the project class string otherwise it's perfect. Down the line we could discuss replacing the static methods in this class but that is outside of the scope of this PR  👍

PR #507, opened 2024-02-29
Remove proteomics fractionation type property

sven1103:
This property is not needed anymore according to the domain experts, so this PR removes it from the protemics template.

Also, the template is renamed to match the domain: ms => proteomics


KochTobi (approved):
The excel sheet is not displayed but I believe you :D

PR #508, opened 2024-03-01
Enable optimistic locking for concurrent sample batch updates

sven1103:
The current implementation allowed JobRunner processes to concurrently load batch entities upon the AddSampleToBatch directive execution.

However, this lead to unwanted side-effects, resulting in batch information with corrupted state. The samples connected to the batch were random after every batch registration.

The reason for this side-effect is the concurrent loading and manipulation of the batch entity. Multiple processes are competing to read and update the sample batch with the sample id they are assigned to add. 

Since this is a classical race condition on data base level, we enable optimistic locking for the batch update transaction.


wow-such-code (approved):
👍

PR #509, opened 2024-03-01
Enable username for new user registrations

sven1103:
This PR introduces the concept of usernames.

If a username already is in use, the user is getting the notification that they have to chose another one.


user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationLayout.java
```diff
@@ -92,8 +98,13 @@ private void createFieldLayout() {
     fieldLayout = new VerticalLayout();
     createEmailField();
     createNameField();
+    createUsernameField();
     createPasswordField();
-    fieldLayout.add(fullName, email, password);
+    fieldLayout.add(fullName, email, username, password);
```
Steffengreiner, line 103: NIT: I found it confusing that the username is inbetween the password and the email. I'd expected the string to login to be above the password 🤔

Steffengreiner (approved):
Heya @sven1103 Nicely done! I have a nitpick and would suggest to also check the codesmells provided by sonarcloud before merging 👍

PR #510, opened 2024-03-04
Use openbis 20.10.x

wow-such-code:
* Uses openBIS version 20 API
* Simplifies what and how things are stored in openBIS in accordance with new data model
* Fixes some workarounds (sample search) caused by old API incompatibility


KochTobi (approved):
LGTM

PR #514, opened 2024-03-11
Improve UI for ontology search and access token generation

wow-such-code:
* improves naming and description for ontology search
* actually counts available ontologies now
* indicates successful copying of the access token, as well as the ontology curie
* resets indicators after 1 second using a background thread


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/OntologyTermRepositoryJpaInterface.java
```diff
@@ -27,4 +27,6 @@ Page<OntologyClass> findByLabelFulltextMatching(
 
   List<OntologyClass> findOntologyClassEntitiesByCurie(String name);
 
+  @Query(value = "SELECT DISTINCT ontology from ontology_classes;", nativeQuery = true)
+  List<String> findUniqueOntologies();
```
KochTobi, line 31: A JPA interface should always return the same type. In your case I would expect a list of OntologyClasses returned.

project-management/src/main/java/life/qbic/projectmanagement/application/ontology/OntologyLookupService.java
```diff
@@ -51,4 +51,8 @@ public Optional<OntologyClass> findByCURI(String curie) {
     return ontologyTermLookup.query(new OntologyCurie(curie)).stream().findAny();
   }
 
+  public List<String> findNumberOfOntologies() {
```
KochTobi, line 54: The method is called findNumber... so why not return a long or int instead?

user-interface/src/main/java/life/qbic/datamanager/views/account/PersonalAccessTokenComponent.java
```diff
@@ -120,11 +120,14 @@ private void showCreatedPersonalAccessToken(String rawTokenText) {
...
     copyIcon.addClassName(IconSize.SMALL);
     copyIcon.addClassName("clickable");
-    copyIcon.addClickListener(
-        event -> UI.getCurrent().getPage().executeJs("window.copyToClipboard($0)", rawTokenText));
+
     Span rawToken = new Span(rawTokenText);
     Span personalAccessTokenWithIcon = new Span(rawToken, copyIcon);
     personalAccessTokenWithIcon.addClassName("token-text");
+
+    copyIcon.addClickListener(
```
KochTobi, line 128: NIT: why did you move this down? I think it makes more sense to keep the listener close to the icon for which it is added.

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/OntologyTermRepository.java
```diff
@@ -93,4 +93,8 @@ private static Sort sortByOrders(List<SortOrder> sortOrders) {
             : Order.asc(sortOrder.propertyName()))
         .toList());
   }
+
+  public List<String> findUniqueOntologies() {
```
KochTobi, line 97: doesn't find unique ontologies but only unique ontology abbreviations

user-interface/src/main/java/life/qbic/datamanager/views/account/PersonalAccessTokenComponent.java
```diff
@@ -137,6 +140,34 @@ private void showCreatedPersonalAccessToken(String rawTokenText) {
     updateUI();
   }
 
+private void handleCopyClicked(String rawTokenText) {
```
KochTobi, line 143: this effectively prevents users from copying the token multiple times. This is bad IMHO because if I accidentally overwrite it in my clipboard I lost the ability to copy it easily. How about leaving the copy icon in place and only changing text and background color?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/ontology/OntologyLookupComponent.java
```diff
@@ -159,13 +160,36 @@ private Span createHeader(String label, String curieText) {
...
-      Span header = new Span(title, curie, initCopyIcon(curieText));
+          event -> handleCopyClicked(header, curieText));
+      copyIcon.addClickListener(
+          event -> handleCopyClicked(header, curieText));
+
       header.addClassName("header");
       return header;
     }
 
+    private void handleCopyClicked(Span header, String curieText) {
```
KochTobi, line 176: Once I copied an ontology, I cannot do so again. This is bad. I want to copy it into multiple fields in my measurement sheet. If I copy a second term, the previous is still highlighted (do we want this?)

KochTobi (changes requested):

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/OntologyTermRepositoryJpaInterface.java
wow-such-code: Unfortunately that won't be possible using the DISTINCT keyword. Alternatives would be not to use the feature or create another interface?

KochTobi: you are right.

PR #516, opened 2024-03-12
Propagate to user if a project with the provided code already exists during project creation

Steffengreiner:
**What was changed**
Handle duplicate project exception in domain service and propagate the error message to the user in application service

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
@@ -1,7 +1,6 @@
 package life.qbic.projectmanagement.application;
 
 import static java.util.Objects.requireNonNull;
-import static life.qbic.logging.service.LoggerFactory.logger;
 
 import java.util.Objects;
 import java.util.Optional;
@@ -9,7 +8,6 @@ import life.qbic.application.commons.ApplicationException;
 import life.qbic.application.commons.ApplicationException.ErrorCode;
 import life.qbic.application.commons.ApplicationException.ErrorParameters;
 import life.qbic.application.commons.Result;
-import life.qbic.logging.api.Logger;
 import life.qbic.projectmanagement.domain.model.project.Contact;
 import life.qbic.projectmanagement.domain.model.project.Funding;
 import life.qbic.projectmanagement.domain.model.project.OfferIdentifier;
@@ -18,9 +16,7 @@ import life.qbic.projectmanagement.domain.model.project.ProjectCode;
 import life.qbic.projectmanagement.domain.model.project.ProjectIntent;
 import life.qbic.projectmanagement.domain.model.project.ProjectObjective;
 import life.qbic.projectmanagement.domain.model.project.ProjectTitle;
-import life.qbic.projectmanagement.domain.repository.ProjectRepository;
 import life.qbic.projectmanagement.domain.service.ProjectDomainService;
-import life.qbic.projectmanagement.domain.service.ProjectDomainService.ResponseCode;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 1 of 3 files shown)


sven1103 (dismissed):
👏🏽  thank you Steffen!

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
```diff
@@ -97,10 +97,15 @@ private Project createProject(String code, String title, String objective,
 
     var registrationResult = projectDomainService.registerProject(intent, projectCode, projectManager, principalInvestigator, responsiblePerson, funding);
 
-    if (registrationResult.isError() && registrationResult.getError().equals(ResponseCode.PROJECT_REGISTRATION_FAILED)) {
-      throw new ApplicationException("Project registration failed.", ErrorCode.GENERAL, ErrorParameters.of(code));
-    }
-
+    registrationResult.onError(responseCode -> {
+      if (responseCode.equals(ResponseCode.PROJECT_ALREADY_EXISTS)) {
+        throw new ApplicationException("Project code is already in use.",
```
KochTobi, line 102:
```suggestion
        throw new ApplicationException("Project code " + projectCode +" is already in use.",
```
you can add debug information here. This message is logged.

project-management/src/main/java/life/qbic/projectmanagement/domain/service/ProjectDomainService.java
```diff
@@ -70,7 +74,8 @@ public Result<Project, ResponseCode> registerProject(
   }
 
   public enum ResponseCode {
-    PROJECT_REGISTRATION_FAILED
+    PROJECT_REGISTRATION_FAILED,
+    PROJECT_ALREADY_EXISTS
```
KochTobi, line 78: You probably want to add a message to message.properties

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
```diff
@@ -97,10 +97,15 @@ private Project createProject(String code, String title, String objective,
 
     var registrationResult = projectDomainService.registerProject(intent, projectCode, projectManager, principalInvestigator, responsiblePerson, funding);
 
-    if (registrationResult.isError() && registrationResult.getError().equals(ResponseCode.PROJECT_REGISTRATION_FAILED)) {
-      throw new ApplicationException("Project registration failed.", ErrorCode.GENERAL, ErrorParameters.of(code));
-    }
-
+    registrationResult.onError(responseCode -> {
+      if (responseCode.equals(ResponseCode.PROJECT_ALREADY_EXISTS)) {
```
KochTobi, line 101: In Java you can now use swich expressions to ensure you handle every response code here.

KochTobi: https://docs.oracle.com/en/java/javase/17/language/switch-expressions.html#GUID-BA4F63E3-4823-43C6-A5F3-BAA4A2EF3ADC

sven1103 (dismissed):
@KochTobi 's comment about the message in message.property is the one crucial imo.

The other comments are valuable, but I would not block the PR because of it.

project-management/src/main/java/life/qbic/projectmanagement/domain/service/ProjectDomainService.java
Steffengreiner: As far as i understood it the message.properties only handle the ApplicationExceptions generated by the application services. For this i already used the implemented Errorcode: 
DUPLICATE_PROJECT_CODE=The project Code {0} already exists -> A valid project code must be unique

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectCreationService.java
Steffengreiner: Sadly the data-manager-application is running in java 17 and pattern matching in switch statements is only possible in java 21 :( 
Additionally since the domain service only provides 2 error responses (Registration failed or Already Exists) an if else statement is not more code 🤔

KochTobi: No for enums this exists from java 17 onwards :)

KochTobi (approved):
LGTM. Nice that you managed to reduce the code base by 25 lines!

PR #521, opened 2024-03-13
When granting project access only disclose username

KochTobi:
Only presents username to other people in the data manager as this is information provided explicitly to be seen by others.
This PR only shows the username. Fine-grained project access (reader, editor, admin) will be tackled in another PR.


Steffengreiner (approved):
As we discussed having the username be required would be awesome but overall nicely done 👍

Steffengreiner (approved):
Ship it! 🚢

PR #523, opened 2024-03-14
Allow distinct label information in proteomics measurements

sven1103:
We learned that in a pooled measurement, labeling information is unique for the sample in the pool.

This PR extends the data manager to preserve this association.


project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/ProteomicsMeasurement.java
```diff
@@ -185,9 +193,8 @@ public void setSamplePreparation(ProteomicsSamplePreparation samplePreparation)
     this.comment = samplePreparation.comment();
   }
 
-  public void setLabeling(ProteomicsLabeling labeling) {
-    this.labelingType = labeling.labelType();
-    this.label = labeling.label();
+  public void setLabeling(Collection<ProteomicsLabeling> labeling) {
+    this.labeling = new HashSet<>(labeling.stream().toList());
```
KochTobi, line 197: You can use `Set.copyOf(labeling)`. This creates an unmodifiable set that does not reflect any changes from the labeling collection.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -70,9 +70,23 @@ private static Optional<ProteomicsMeasurementMetadata> merge(
     }
     List<SampleCode> associatedSamples = measurementMetadataList.stream().map(
         ProteomicsMeasurementMetadata::sampleCodes).flatMap(Collection::stream).toList();
+    List<Labeling> labels = measurementMetadataList.stream().map(
+        metadata -> {
+          Labeling labeling;
+          try {
+            labeling = metadata.labeling().iterator().next();
+          } catch (Exception e) {
```
KochTobi, line 78: please only catch `RuntimeException` or NoSuchElementException

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -70,9 +70,23 @@ private static Optional<ProteomicsMeasurementMetadata> merge(
...
+            log.warn(
+                "Proteomics label: No label information found for pooling %s".formatted(
+                    metadata.sampleCodes().iterator().next().code()));
+            labeling = new Labeling(metadata.sampleCodes().iterator().next().code(), "", "");
```
KochTobi, line 82: the iterator of the sample codes and the iterator of the labeling do not return their elements in a specific order. According to the method signature we get a Collection here which does not provide ordered elements (might if it is actually a list but this is not guaranteed by the method signature). If you assume that only one sample and one labelling information is available, please check the assumption separately and fetch the information. 
It might be a good idea to extract a method for the lambda here since it is very complex to read.

KochTobi (changes requested):

KochTobi (approved):
LGTM

PR #527, opened 2024-03-15
Generalize upload measurement dialog to handle edit and register measurements 

Steffengreiner:
**What was changed** 
1.) To avoid duplication, the UploadMeasurementDialog now handles Validation via an executor interface, which can be provided for editing or registering measurement purposes. 
2.) Additionally, the fraction name is now shown in the grid and is searchable via the searchfield


project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/ValidationService.java
```diff
@@ -23,20 +23,20 @@ public class ValidationService {
 
   private final NGSValidator ngsValidator;
 
-  private final ProteomicsValidator pxValidator;
+  private final ProteomicsValidator pxpValidator;
```
KochTobi, line 26: why pxp?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -188,10 +190,19 @@ private void createProteomicsGrid() {
     proteomicsMeasurementGrid.addClassName("measurement-grid");
     proteomicsMeasurementGrid.addColumn(
             proteomicsMeasurement -> proteomicsMeasurement.measurementCode().value())
-        .setHeader("Measurement Code").setAutoWidth(true).setTooltipGenerator(proteomicsMeasurement -> proteomicsMeasurement.measurementCode().value());
+        .setHeader("Measurement Code").setAutoWidth(true).setTooltipGenerator(
+            proteomicsMeasurement -> proteomicsMeasurement.measurementCode().value()).setFlexGrow(0);
```
KochTobi, line 194: for my understanding; not a change request; what is the `setFlexGrow(0)` here? Why is it necessary?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -188,10 +190,19 @@ private void createProteomicsGrid() {
...
     //Todo Should the sampleCodes be retrieved via a service or from column?
-    proteomicsMeasurementGrid.addColumn(proteomicsMeasurement -> proteomicsMeasurement.label().orElse("")).setHeader("Measurement Label").setTooltipGenerator(proteomicsMeasurement -> proteomicsMeasurement.label().orElse("")).setAutoWidth(true);
-    proteomicsMeasurementGrid.addColumn(proteomicsMeasurement -> proteomicsMeasurement.labelingType().orElse("")).setHeader("Measurement Label Type").setTooltipGenerator(proteomicsMeasurement -> proteomicsMeasurement.labelingType().orElse("")).setAutoWidth(true);
+    proteomicsMeasurementGrid.addColumn(
+            proteomicsMeasurement -> proteomicsMeasurement.label().orElse(""))
+        .setHeader("Measurement Label")
+        .setTooltipGenerator(proteomicsMeasurement -> proteomicsMeasurement.label().orElse(""))
+        .setAutoWidth(true).setFlexGrow(1);
```
KochTobi, line 200: NIT: you can put `.setFlexGrow(1)` in another line so statements easily can be removed.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/RegistrationValidationExecutor.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * <b>Registration Validation Executor</b>
+ * <p>
+ * Implementation of the {@link ValidationExecutor} interface,
+ * handling the validation of newly registered {@link life.qbic.projectmanagement.application.measurement.MeasurementMetadata}
+ *
+ */
+@Component
+public class RegistrationValidationExecutor implements ValidationExecutor {
```
KochTobi, line 22: Weird naming. Why not `MeasurementMetadataValidator`?
I think this is not for every registration but only for measurement metadata? What is the difference between this executor and the validation service?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/ValidationExecutor.java
```diff
@@ -0,0 +1,29 @@
...
+import life.qbic.projectmanagement.application.measurement.validation.ValidationService.Domain;
+
+/**
+ * <b>Validation Executor</b>
+ * <p>
+ * An interface that enables injection of different configurations of
+ * {@link life.qbic.projectmanagement.application.measurement.validation.ValidationService} calls,
+ * based on the use case (create or edit).
+ * <p>
+ * To support more domains, just create more methods here.
```
KochTobi, line 17: This smells a bit. Why not have two different interfaces? Why not have a more general interface? Can you elaborate?
For example: `ValidationExecutor<T>` with `NGSValidationExecutor<NGSMEasurementMetadata>` and `ProteomicsValidationExecutor<ProteomicsMeasurementMetadata>` implementing 
`<T> ValidationResult validate(T metadata)`

KochTobi (commented):
Hey looks good and works fine. Some questions I would ask you to answer

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/ValidationService.java
Steffengreiner: Just for consistency reasons, we call registerPxP in the measurmentservice and check for "pxp" properties in the validator

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
Steffengreiner: The Flexgrow call ensures that the this column width doesn't grow or shrink if the grid size changes. This was done to ensure that the measurement code is always readable. For more detail see:
https://vaadin.com/docs/latest/components/grid#columns

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
Steffengreiner: I like the idea and adapted the call for all columns accordingly 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/RegistrationValidationExecutor.java
Steffengreiner: The naming is related to when this executor is used. This PR introduces different validation triggering executor for "Registration" and "Editing" measurement metadata. 
I'll happily adapt the naming scheme, but then this should also be done for ValidationResult Validation Report Validation Executor etc. to be consistent 🤔

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/ValidationExecutor.java
Steffengreiner: This is due to two reasons: 
1.) The dialog itself has to decide what kind of domain validation is necessary so we can't know during it's initialization what the domain will be.
2.) The idea of this interface was to seperate the edit and register cases for the different facilities because these cases have different validation needs
(Measurement Registration has no measurement codes, while editing measurements has to ensure the correct measurement codes were provided)

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
KochTobi: thank you for the clarification! :)

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/RegistrationValidationExecutor.java
KochTobi: Ah so the executor runs the validation service based on a trigger? Then `MeasurementRegistrationTriggeredValidator` would fit better. I do not see why this changes anything for `ValidationResult` and `ValidationReport`. The interface name can stay the same IMHO.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/ValidationExecutor.java
KochTobi: to 
1.) Why? What is the reasoning to delegate that to the dialog?
2.) Still, I do not see a reason for different interfaces?

Steffengreiner: To the provided questions
1.) The dialog will show the user immediately after uploading a file within the grey listbox of a file item the missing/wrong information for each uploaded file. 
Therefore the dialog has to know the domain of the uploaded file to trigger the correct validation. 

2.) The dialog has to know if the user wants to edit an existing measurement or add a new measurement, since: 
A) Some elements in the dialog will change (Save instead of register as the confirmbutton label e.g.). 

B) The provided information has to be validated differently dependent on if the user intends to edit or register a measurement. (Registrations will have no measurement codes,  User is not allowed to edit some of the provided columns in existing measurements. )

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementValidationResult.java
```diff
@@ -7,50 +7,52 @@
 import java.util.stream.Stream;
 
 /**
- * <b><class short description - 1 Line!></b>
+ * <b>Measurement Validation Result</b>
```
KochTobi, line 10: I do no think you have to split the validation result in two. You can keep the general validation result as it is only a means to transfer data about the result of a validation. This can be successful or failed and may additionally contain some warnings.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementValidationService.java
```diff
@@ -14,40 +14,42 @@
...
-    this.pxValidator = pxValidator;
+  public MeasurementValidationService(MeasurementNGSValidator measurementNgsValidator,
+      MeasurementProteomicsValidator pxpValidator) {
+    this.measurementNgsValidator = measurementNgsValidator;
+    this.pxpValidator = pxpValidator;
   }
 
-  public ValidationResult validateNGS(NGSMeasurementMetadata ngsMeasurementMetadata) {
-    return ngsValidator.validate(ngsMeasurementMetadata);
+  public MeasurementValidationResult validateNGS(NGSMeasurementMetadata ngsMeasurementMetadata) {
```
KochTobi, line 35: can return a simple ValidationResult

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementValidator.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * <p>Validation interface employed to check the provided user input for a measurement.
+ *    Implementations of this interface are supposed to be domain specific and will
+ *    return a ValidationResult dependent on the presence or absence of data
+ * </p>
+ *
+ */
+public interface MeasurementValidator<T extends MeasurementMetadata> {
+
+  MeasurementValidationResult validate(T measurementMetadata);
```
KochTobi, line 16: can return a simple ValidationResult

PR #529, opened 2024-03-21
Add Token Encoder Interface

KochTobi:
Adds an interface for the token encoder. Makes it easier to exchange this component.


Steffengreiner (approved):
Nicely done 👍

PR #532, opened 2024-03-27
Change token generation to use the same application salt for faster lookup

KochTobi:
Makes the salt for the randomly created tokens configurable. All tokens use the same salt, this allows for faster resolvement of raw tokens as the expected value can be pre-computed (provided the configured salt).


user-interface/src/main/java/life/qbic/datamanager/views/account/AddPersonalAccessTokenDialog.java
```diff
@@ -31,12 +28,11 @@
  */
 public class AddPersonalAccessTokenDialog extends DialogWindow {
 
-  private static final Logger log = logger(AddPersonalAccessTokenDialog.class);
   @Serial
   private static final long serialVersionUID = 6149070385320827888L;
   private final TextField tokenDescription = new TextField();
   private final Select<Duration> expirationDate = new Select<>();
-  private final Binder<PersonalAccessTokenDTO> personalAccessTokenDTOBinder;
+  private final Binder<PersonalAccessTokenFrontendBean> personalAccessTokenDTOBinder;
```
Steffengreiner, line 36: NIT: While i like the rename of the object i'd suggest to also rename the fieldName as well

user-interface/src/main/resources/application.properties
```diff
@@ -109,3 +109,6 @@ spring.mustache.check-template-location=false
 #EHCache used by spring security ACL to cache ACEs
 spring.cache.jcache.config=classpath:ehcache3.xml
 ###############################################################################
+################### Token Generation ##########################################
+qbic.access-token.salt=${ACCESS_TOKEN_SALT:}
```
Steffengreiner, line 113: What is the reason for the salt itself ending on a colon? 
```suggestion
qbic.access-token.salt=${ACCESS_TOKEN_SALT:}
```

user-interface/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -19,6 +19,16 @@
...
+    {
+      "name": "qbic.access-token.salt",
+      "type": "java.lang.String",
+      "description": "The salt used to encrypt personal access tokens."
+    },
+    {
+      "name": "qbic.access-token.iteration-count",
+      "type": "java.lang.String",
+      "description": "The number of iterations used in personal access token encryption."
     }
```
Steffengreiner, line 32: Nicely done! However should this information also be provided in the Readme markdown file?

identity/src/test/groovy/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoderTest.groovy
```diff
@@ -0,0 +1,116 @@
+package life.qbic.identity.domain.model.token
+
+import spock.lang.Specification
+
+class PersonalAccessTokenEncoderTest extends Specification {
```
Steffengreiner, line 5: NIT: Shouldn't the test be called PersonalAccessTokenEncoderSpec to adhere to our testing naming scheme?

Steffengreiner (changes requested):
Heya @KochTobi, nicely done! However i found some minor naming issues and am confused on why we would use the same salt for all the tokens. As far as i looked into it this makes the salting process [insecure and useless](https://auth0.com/blog/adding-salt-to-hashing-a-better-way-to-store-passwords/#Mitigating-Password-Attacks-with-Salt)

user-interface/src/main/resources/application.properties
KochTobi: The colon separates the value from the default. No default value exists here so the default will be empty (which is checked in the code)

user-interface/src/main/resources/META-INF/additional-spring-configuration-metadata.json
KochTobi: I think not :D the readme is out of date already and those descriptors are easily parseable.

identity/src/test/groovy/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoderTest.groovy
KochTobi: Yes

KochTobi:
I think we should probably change the download server instead. Thank you for your review @Steffengreiner

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,172 @@
...
+
+  public PersonalAccessTokenEncoder(
+      @Value("${qbic.access-token.salt}") String salt,
+      @Value("${qbic.access-token.iteration-count}") int iterationCount) {
+    this.salt = requireNonNull(salt, "salt must not be null");
+    if (this.salt.isBlank()) {
+      throw new IllegalArgumentException("salt must not be blank");
+    }
+    if (iterationCount <= 0) {
+      throw new IllegalArgumentException("iterationCount must be greater than 0");
```
sven1103, line 44: I think this exception is confusing, since the count actually must be greater or equal than the min count const.

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if (iterationCount <= EXPECTED_MIN_ITERATION_COUNT) {
+      log.warn(
+          "Low iteration count for token encryption detected n=%d. Expected minimal iteration count is n=%d".formatted(
+              iterationCount, EXPECTED_MIN_ITERATION_COUNT));
```
sven1103, line 49: Given the application and the power of our CPUs, the iteration count shall be as high as possible (see the NIST recommendation: https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf). I would rather not go under 100_000 iterations as we have it in the password encryption policy. 

The application shall treat it as exception, if the iteration count is too low to enforce minimal security constraints.,

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,172 @@
...
+    byte[] hash = pbe(token, saltBytes, iterationCountCopy);
+    return iterationCountCopy + ":" + toHex(saltBytes) + ":" + toHex(hash);
+  }
+
+  @Override
+  public boolean matches(char[] token, String encodedToken) {
+    byte[] readSalt = readSalt(encodedToken);
+    if (!Arrays.equals(readSalt, this.salt.getBytes(StandardCharsets.UTF_8))) {
+      log.warn("Personal access token has different salt than currently configured.");
+    }
```
sven1103, line 81: that is not an issue. The salt can be stored with the encrypted token, changes in the salt must not affect existing tokens.

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,172 @@
...
+    int readIterationCount = readIterationCount(encodedToken);
+    if (readIterationCount != this.iterationCount) {
+      log.warn("Personal access token has different iterations than currently configured.");
+    }
```
sven1103, line 85: This is not an issue, since we store the iteration count with the encrypted token

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,172 @@
...
+
+  }
+
+  private static final EncryptionSettings ENCRYPTION_SETTINGS = new EncryptionSettings("AES", 256);
+
+
+  @Override
+  public String encode(char[] token) {
+    String saltyString = requireNonNull(this.salt, "this.salt must not be null");
+    if (saltyString.isBlank()) {
```
sven1103, line 64: This is not enough ("All or a portion of the salt shall be generated using an approved Random Bit Generator
(e.g., see [5]). The length of the randomly-generated portion of the salt shall be at least
128 bits", NIST recommendation: https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf),

Please ensure that the byte array of the encoded value is of size 16 or longer.

identity/src/test/groovy/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoderTest.groovy
```diff
@@ -0,0 +1,116 @@
...
+        !tokenEncoderUnderTest.matches(tokenEncoderUnderTest.encode(secret).toCharArray(), encodedToken)
+
+        where:
+        encodedToken << [
+                "900:31323334:7a85f3f3c352af5964bb75750fbb98f7e3febd5d00b1793c5d0382c9d59d644e",
+                "10000:31323000:7a85f3f3c352af5964bb75750fbb98f7e3febd5d00b1793c5d0382c9d59d644e",
+                "10000:31323334:0005f3f3c352af5964bb75750fbb98f7e3febd5d00b1793c5d0382c9d59d6000"
+        ]
+    }
+
```
sven1103, line 93: Please add tests for too short iteration counts leading to an expection, same for too short salt length

sven1103 (changes requested):
I have made some comments that address some recommendations I would kindly ask to please implement them.

identity/src/main/java/life/qbic/identity/domain/model/token/PersonalAccessTokenEncoder.java
```diff
@@ -0,0 +1,168 @@
...
+@Service
+public class PersonalAccessTokenEncoder implements TokenEncoder {
+
+  private static final Logger log = logger(PersonalAccessTokenEncoder.class);
+
+  private final byte[] salt;
+  private final int iterationCount;
+
+  private static final int EXPECTED_MIN_ITERATION_COUNT = 100_000;
+  private static final int EXPECTED_MIN_SALT_BITS = 126;
```
sven1103, line 30: why 126?

sven1103 (approved):
LGTM!

PR #533, opened 2024-03-28
Hide contact dropdown from everyone but admins

wow-such-code:
* Tests authorization of user when fetching list of contacts
* Only admins are allowed to see contacts
* Hides contact dropdown menu in project creation and edit when it is empty


user-interface/src/main/java/life/qbic/datamanager/views/general/contact/AutocompleteContactField.java
```diff
@@ -143,6 +143,7 @@ public void setContact(Contact contact) {
 
   public void setItems(List<Contact> contacts) {
     contactSelection.setItems(contacts);
+    contactSelection.setVisible(!contacts.isEmpty());
   }
```
Steffengreiner, line 147: I'd move this logic into the catch block of the access denied exception if possible. Our fields should not decide on their visibility but the layout containing the field should imo 🤔

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -80,8 +82,11 @@ public AddProjectDialog(FinanceService financeService,
     this.experimentalInformationLayout = new ExperimentalInformationLayout(
         ontologyLookupService);
 
-    List<Contact> knownContacts = contactRepository.findAll().stream()
-        .map(contact -> new Contact(contact.fullName(), contact.emailAddress())).toList();
+    List<Contact> knownContacts = new ArrayList<>();
+    try {
+      knownContacts = contactRepository.findAll().stream()
+          .map(contact -> new Contact(contact.fullName(), contact.emailAddress())).toList();
+    } catch (AccessDeniedException e) {}
```
Steffengreiner, line 89: I fear this can not be changed to return a result object, which is then handled dependent on the viewers permissions? Since this is not an exceptional case, having an exception with try catch block seems dangerous

Steffengreiner (changes requested):
Heya @wow-such-code nicely done! However i have some minor improvement suggestions, and would ask for feedback if this is the final design we want to go for for the normal user? Maybe we can have the button behind the fields or add some spacing?  
userexperience

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/CollaboratorsLayout.java
```diff
@@ -118,6 +118,18 @@ public ProjectCollaborators getProjectCollaborators() {
...
+    principalInvestigatorField.setItems(knownContacts);
+    responsiblePersonField.setItems(knownContacts);
+    projectManagerField.setItems(knownContacts);
+  }
+
+  public void hideContactBox() {
+    principalInvestigatorField.hideContactBox();
+    responsiblePersonField.hideContactBox();
+    projectManagerField.hideContactBox();
+  }
```
Steffengreiner, line 131: Even fancier would be a method like formlayout.showContactBox(knownContacs.isEmpty()) that sets the visibility dependent on the boolean provided

Steffengreiner (dismissed):
Heya @wow-such-code! Awesome work, I have some minor improvement points but they are more NIT than functionality changing. Therefore this is good to go 🥇

project-management/src/main/java/life/qbic/projectmanagement/application/ContactRepository.java
```diff
@@ -22,7 +26,7 @@ private static List<Contact> dummyContacts() {
         new Contact("Anna Bell", "anna.bell@qbic.uni-tuebingen.de"),
         new Contact("Sophia Turner", "sophia.turner@qbic.uni-tuebingen.de"),
         new Contact("Tylor Smith", "tylor.smith@qbic.uni-tuebingen.de")
-    );
+    ));
   }
```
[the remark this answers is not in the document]
wow-such-code: 🤷

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/CollaboratorsLayout.java
wow-such-code: I think this is unnecessary, at least at the moment, because there is no possibility that the visibility will be set to true.

Steffengreiner (approved):
LGMT 🚀

PR #535, opened 2024-04-02
Adjust proteomics columns in sheet and UI

wow-such-code:
* LC column is now shown before Instrument, Labeling type and Label after Instrument, both in the Excel sheet, as well as the UI
* Enrichment Method and Cycle/Fraction Name are now optional columns


sven1103 (approved):
LGTM!

PR #537, opened 2024-04-09
Add artemis broker configuration

KochTobi:
**Please note: This will disable the embedded broker**


Steffengreiner (approved):
LGTM, however some explanation on what these properties are in the readme or metadata properties would be helpful imo🥇

PR #539, opened 2024-04-09
Increase ontology column length

KochTobi:


wow-such-code:
converting to draft to make other similar changes, e.g.
alter table sample modify species text;
alter table sample modify specimen text;
alter table sample modify analyte text;

KochTobi:
@wow-such-code what was the plan here? The length in the jpa annotation helps hibernate determine the length in the column.

wow-such-code:
> @wow-such-code what was the plan here? The length in the jpa annotation helps hibernate determine the length in the column.

been a whilte, but afaik these are json columns and when the two of us worked on fixing this on the server we changed them to text so we could guarantee long ontology entries to work.

KochTobi:
solved by https://github.com/qbicsoftware/data-manager-app/pull/605

PR #541, opened 2024-04-11
Fix typo and naming description in columns of template sheets

Steffengreiner:
**What was changed** 

1.) Adjust sample ids column to sample id column since samples are now pooled dependent on the value provided in the the sample pooling column
2.) Remove outdated information from both sheets in allowed values and property information


sven1103 (approved):
LGTM

PR #543, opened 2024-04-12
Store registration instant for new user registrations

sven1103:
For business reports, we need to resolve user registrations to actual datetimes.

### DIFF
--- a/identity/src/main/java/life/qbic/identity/domain/model/User.java
+++ b/identity/src/main/java/life/qbic/identity/domain/model/User.java
@@ -7,6 +7,7 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.Table;
 import java.io.Serial;
 import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


Steffengreiner (changes requested):
Heya @sven1103, i tested the code which sadly led to an exception during login for the current users and registration of new users:

`2024-04-12 11:05:58,047 ERROR [http-nio-8080-exec-1] o.s.s.w.a.UsernamePasswordAuthenticationFilter: An internal error occurred while trying to authenticate the user.
org.springframework.security.authentication.InternalAuthenticationServiceException: Could not extract column [6] from JDBC ResultSet [Zero date value prohibited] [n/a]`

I suspect that the current users need to be also provided with a registration date in the database.

Steffengreiner (approved):
I added the registration date for the current users and it works as expected! Nicely done 👍

PR #545, opened 2024-04-12
Enable User to download and edit NGS measurement data 

Steffengreiner:
**What was changed**

1.) User is able to edit NGS measurement data similiar to how the proteomics measurement data can be edited 

2.) User is able to download the measurement data dependent on the currently selected tab


sven1103 (changes requested):
Thank you for you contribution!

I observered, that the sample pooling information is not displayed when the user downloads the metadata NGS sheet. 

Moreover, the update / registration feedback message shows a different number of successfull entries to the one showed during validation. This is because pooled samples are merged and count as one measurement. Thus you need to display the number on the client side and not infer the number based on the service response.

NIT: During the in process view of the dialog, the case of the mode is wrong, it should be in small letters.

sven1103 (approved):
Great work!

PR #547, opened 2024-04-15
Enable async service calls for measurements

sven1103:
Enables async service method calls for measurement updates and registrations.

The current implementation also shows a intermediate progress bar to the user, until the task is finished.

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
@@ -17,6 +17,7 @@ import life.qbic.projectmanagement.domain.model.project.ProjectObjective;
 import life.qbic.projectmanagement.domain.model.project.ProjectTitle;
 import life.qbic.projectmanagement.domain.repository.ProjectRepository;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Async;
 import org.springframework.security.access.prepost.PostFilter;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.stereotype.Service;
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
@@ -7,8 +7,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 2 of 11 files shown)


Steffengreiner (changes requested):
Heya @sven1103! Thanks for the awesome contribution!
 I tested it with a proteomics sheet with invalid measurement Ids(Correct ID format but from another project) and noticed three issues.

1.) The user is bombarded with emails notifying him for each failing registration/update process of a measurement.
2.) The MeasurementRegistrationException (`ErrorCode.SAMPLECODE_NOT_FROM_PROJECT`) is swallowed so the user is not informed that this is the origin of the failure. 
3.) It is recommended to provide a custom async exception handler by implementing [AsyncUncaughtExceptionHandler interface](https://www.baeldung.com/spring-async), which is able to handle any uncaught exceptions(e.g. if a thread returns void) 

As a final NIT: Ideally we could parse the registration status every other second or so and provide the user with an actual count of the update process.

sven1103:
Dear @Steffengreiner, thank your for your comments, i will look into it. Regarding the registration status: it is either open or done. A quantitative progress feedback is currently not possible and would require substantial refactorings in the measurement registration logic. I would like to postpone that, since we are in an exceptionable registration time in the multiple seconds range.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementProteomicsValidator.java
```diff
@@ -161,6 +175,28 @@ ValidationResult validateSampleIds(Collection<SampleCode> sampleCodes) {
...
+        log.error("No project information found for projectId: " + projectId);
+        throw new ValidationException("This should not happen, please try again.");
+      }
+      var experimentIds = projectQuery.get().experiments();
+      var sampleQuery = sampleInformationService.findSampleId(sampleCode).flatMap(
+          sampleIdCodeEntry -> sampleInformationService.findSample(sampleIdCodeEntry.sampleId()));
+      if (sampleQuery.isEmpty()) {
+        log.error("No sample information found for sample id: " + sampleCode);
+        return ValidationResult.withFailures(1,
+            List.of("No sample information found for sample id: %s".formatted(sampleCode)));
```
Steffengreiner, line 191: Screenshot 2024-04-17 at 18 26 34
I think it's missing the string representation of the code here  😅

Steffengreiner (changes requested):
Heya @sven1103 Nicely done I love that we can now show the samplecodes which failed during edit and registration. I have only one minor request to be changed before it can be approved!

Steffengreiner (approved):
Nicely done 🥇

PR #549, opened 2024-04-18
User is able to see all his accessible projects in a paginated fashion in the projectOverview

Steffengreiner:
**What was changed**
Fixes the bug reported in [DM-1198](https://qbicsoftware.atlassian.net/browse/DM-1198).
Sadly the `PostFilter()` annotation is not possible for ACL secured objects within a Paginated context, which is a known spring [issue](https://github.com/spring-projects/spring-security/issues/2629) for over 10 years now. 

**Context**

Postfilter filters the projects after they are returned, leading to the issue that the original query number of projects is not fulfilled(since all inaccessible projects are removed before they are handed to the UI) leading to reported behaviour

**Solution**

1.) First load all projectIds to which the user and his role has access to within the ACL context. 
2.) Filter the paginated query via our already established specification framework to only retrieve the projects which are accessible to the user 

**Possible Improvements** 
Implement JPA based repositories out of the ACL columns to avoid having to manually define a query. 

[DM-1198]: https://qbicsoftware.atlassian.net/browse/DM-1198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/project/ProjectPreviewLookupImplementation.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public List<ProjectPreview> query(int offset, int limit) {
+    return projectPreviewRepository.findAll(new OffsetBasedRequest(offset, limit)).getContent();
+  }
+
+  @Override
+  public List<ProjectPreview> query(String filter, int offset, int limit,
+      List<SortOrder> sortOrders, Collection<ProjectId> projectIds) {
```
wow-such-code, line 39: NIT: Have you tried this using PreFilter annotation with the projectIds? It sounds easier to me, so you don't have to manually filter.

wow-such-code (approved):

PR #550, opened 2024-04-19
Allow deletion of measurements without data

wow-such-code:


sven1103 (changes requested):
Dear @wow-such-code, 

thank you for your pull-request!

I have some remarks to the feature, before I go into the code:

1.) Deletion does not require confirmation.
On delete action, we agreed on to provide the user with a confirm dialog, to avoid accidental delete actions.

2. There is no possibility to delete all measurements

3. There is no possibility to select multiple easily, lets say a block of mutiple of 20 or so. We already got the feedback from the platform coordinators, that this is quite a common use case.

Maybe we can discuss that with @Shraddha and get a better understanding of the feature requirements.

shraddha:
Kindly remove me from your mailing list. I am not the intended recipient.

user-interface/frontend/themes/datamanager/components/page-area.css
```diff
@@ -148,6 +148,9 @@
   width: 100%;
 }
 
+.selected-checkbox-grid * vaadin-checkbox {
+}
```
Steffengreiner, line 152: This css doesn't seem to do anything?

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * <class short description - One Line!>
+ * <p>
+ * <More detailed description - When to use, what it solves, etc.>
+ *
+ * @since <version tag>
+ */
```
Steffengreiner, line 20: Missing description (failure on my part though) 😓

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private final Set<T> selectedItems = new HashSet<>();
+
+  private final List<ComponentEventListener<CheckBoxSelectedEvent>> listeners = new ArrayList<>();
+
+  private final Checkbox selectAllCheckBox = new Checkbox();
+
+  /**
+   * Creates a new instance, with page size of 50.
+   */
```
Steffengreiner, line 33: This default description can be removed

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -0,0 +1,136 @@
...
+  private void addSelectAllItemsListener() {
+    selectAllCheckBox.addValueChangeListener(event -> {
+      if(event.isFromClient()) {
+        if (event.getValue()) {
+          selectedItems.addAll(getLazyDataView().getItems().toList());
+        } else {
+          getLazyDataView().getItems().toList().forEach(selectedItems::remove);
+        }
+      }
+      //Todo Should this be treated differently?
```
Steffengreiner:
```suggestion
```

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
```diff
@@ -281,8 +281,7 @@ private void onEditBatchClicked(EditBatchEvent editBatchEvent) {
   }
 
   private boolean isSampleRemovable(SampleId sampleId) {
-    ProjectId projectId = context.projectId().orElseThrow();
-    return deletionService.isSampleRemovable(sampleId, projectId);
+    return deletionService.isSampleRemovable(sampleId);
```
Steffengreiner, line 284: Why was this touched in this PR? Removing a measurement should not remove a sample 😕

Steffengreiner (changes requested):
Heya @wow-such-code! Nice work, i found some nitpicks and have some questions:
1.) The selected measurement info box (and maybe measurement list is not reset) if i switch between projects.
2.) The selected measurement info box (and maybe measurement list is not reset) if i switch between the genomic and proteomic tabs within a project
3.) Why were the sampledeletion checks adapted in this PR, deleting a measurement should not delete its associated sample AFAIK
4.) The measurement info box should be initialized with "0 measurement selected" or manually set to be visible if more then 0 measurements are selected to avoid having the grey square initialization
5.) Should the select all measurements be dependent on the filter? This i'm unsure of but wanted to bring it up again. 
6.) This is a NIT: It would be nice if the user could be informed which measurements have data attached to avoid selecting it again

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
wow-such-code: Sorry about that. It turns out openBIS does not need the projectID for a lot of operations and I removed the clutter.

wow-such-code:
Thank you for the thorough review!

> 1.) The selected measurement info box (and maybe measurement list is not reset) if i switch between projects.
> 2.) The selected measurement info box (and maybe measurement list is not reset) if i switch between the genomic and proteomic tabs within a project

Both issues have now been addressed.

> 3.) Why were the sampledeletion checks adapted in this PR, deleting a measurement should not delete its associated sample AFAIK

These changes were the removal of unused/unnecessary parameters for openBIS which I found while implementing deletion. Had to pull on that thread, sorry.

> 4.) The measurement info box should be initialized with "0 measurement selected" or manually set to be visible if more then 0 measurements are selected to avoid having the grey square initialization

Done!

> 5.) Should the select all measurements be dependent on the filter? This i'm unsure of but wanted to bring it up again.

I think they should (select all you can see in the table, like in Gmail) and now they are.

> 6.) This is a NIT: It would be nice if the user could be informed which measurements have data attached to avoid selecting it again

I think that is a good idea, but a bit complicated and it should be addressed in a future PR.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMain.java
```diff
@@ -119,6 +126,10 @@ public MeasurementMain(
     add(measurementTemplateListComponent);
     add(measurementTemplateDownload);
     add(measurementDetailsComponent);
+
+    measurementDetailsComponent.addListener(
+        selectionChangedEvent -> updateSelectedMeasurementInfo(selectionChangedEvent.getSource().getNumberOfSelectedMeasurements()));
```
Steffengreiner, line 131: This will lead to a null exception if the user switches between the measurement views within a project or between projects:

https://github.com/qbicsoftware/data-manager-app/assets/29627977/609bcf4d-83db-4c2e-af66-b19f7bd7e33d

`java.lang.NullPointerException: Cannot invoke "com.vaadin.flow.component.tabs.Tab.getLabel()" because the return value of "com.vaadin.flow.component.tabs.TabSheet.getSelectedTab()" is null
	at life.qbic.datamanager.views.projects.project.measurements.MeasurementDetailsComponent.getSelectedTabName(MeasurementDetailsComponent.java:369)
	at life.qbic.datamanager.views.projects.project.measurements.MeasurementDetailsComponent.getNumberOfSelectedMeasurements(MeasurementDetailsComponent.java:315)
	at life.qbic.datamanager.views.projects.project.measurements.MeasurementMain.lambda$new$e391be34$1(MeasurementMain.java:131)
	at life.qbic.datamanager.views.projects.project.measurements.MeasurementDetailsComponent.lambda$updateSelectedMeasurementsInfo$3(MeasurementDetailsComponent.java:273)`

Steffengreiner (changes requested):
Heya @wow-such-code, Nicely done, only one small bugfix left before approval 🙌

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMain.java
```diff
@@ -423,4 +518,13 @@ private void routeToRawData(ComponentEvent<?> componentEvent) {
...
     }
   }
+
+  private void selectionChanged(int selectedMeasurements, boolean updateUI) {
+    if(updateUI) {
+      String text = "%s measurements are currently selected.".formatted(
+          String.valueOf(selectedMeasurements));
+      measurementsSelectedInfoBox.setText(text);
+    }
+  }
```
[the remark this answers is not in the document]
wow-such-code: The reason for this was to only update the UI once, when the "select all" box is selected. Good point about the tab switch! That needs to be implemented then - unless this is not the right way to do it?

Steffengreiner (approved):
Nicely done

PR #553, opened 2024-04-23
Security overhaul and small bug fixes

sven1103:
Annotates public service methods to wire Spring Security with them.

Also fixes a bug during batch deletion, that triggers the sample deletion for the batch.

The sample deletion throws an event though, and a policy removes the sample reference in the referenced batch. In the case of a batch deletion event, there is no batch existing anymore with a given ID, and all background jobs failed.

This PR introduced a special error response code, that will only lead to a warning that is logged and is not treated as exception anymore. 

For all other error codes an exception is still thrown, since the system needs to recover from it.


Steffengreiner (changes requested):
Heya @sven1103 awesome work, thanks for securing our application! 
I have one question concerning the context of checking user permissions in our system context in terms of events. 🤔

Additionally the following edit processes are still possible with a user without write access, which i think should be protected as well:

1.) Upload Offer/Delete Offer
2.) Upload Sample QC/Delete Sample QC
3.) Create/Delete/Edit Experimental Variables
4.) Create/Delete/Edit Experimental Groups
5.) Edit Measurement throw the access denied exception after triggering the upload process, while Register Measurement throws the access denied exception before the file has finished uploading.

Steffengreiner (changes requested):
Awesome work! 🥳 
However i was still able to edit experiment information such as 
name,
species, 
specimen
analytes 
as a user without write permissions. This should be fixed before the PR can be approved

Steffengreiner (approved):
Well done 👍

PR #556, opened 2024-04-26
Test datasources on startup

wow-such-code:
* Adds a connection testing interface for openBIS
* Tests connection to application and datastore server on startup
* Logs successful connection attempts, otherwise startup fails with openBIS errors


Steffengreiner (approved):
LGTM! Thanks for the update

PR #560, opened 2024-05-02
Allow to add new groups even if samples are registered

wow-such-code:
* Fixes issue where adding experimental groups would delete and recreate old groups
* Allows users to add (**not edit**) experiment groups, even if samples are already registered


Steffengreiner (approved):
Nice catch 👀

PR #561, opened 2024-05-02
Project are shown as card designs on the "home page" 

Steffengreiner:
**What was changed** 

Redesign of the project list style grid into individual cards for each project, showing additional information such as: 
1.) The measurement types associated with the project
2.) The collaborators associated with each project
3.) The project responsible party if present and the project manager

**Missing features**
User Avatars are not colorized, since we currently don't store the avatars for each user. Therefore this feature was delayed after discussion with SP until the user profile page is ready

**Technical issues**
1.) Due to how the information is loaded within the grid, only the information within the projectpreview bean can be filtered via the searchfield. (ProjectCode, Modification Date, Project Name)
This has been documented as a seperate issue and should be tackled seperately, 
since we encounter a similiar issue([DM-1230](https://qbicsoftware.atlassian.net/browse/DM-1230)) in the measurement and raw data grid. 

2.) The project list will show a deleted user associated with a project as anonymous instead of throwing an exception 
This can happen if there is a mismatch between the entries defined in the ACL list and the user table (E.g. by deleting the user manually in the database) 

[DM-1230]: https://qbicsoftware.atlassian.net/browse/DM-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementLookupService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementLookupService.java
@@ -80,8 +80,12 @@ public class MeasurementLookupService {
         + measurementLookup.countProteomicsMeasurementsBySampleIds(sampleIds);
   }
 
-  public List<ProteomicsMeasurement> queryAllProteomicsMeasurement(List<SampleId> sampleIds) {
+  public List<ProteomicsMeasurement> queryAllProteomicsMeasurements(List<SampleId> sampleIds) {
     return measurementLookup.findProteomicsMeasurementsBySampleIds(sampleIds);
+  }
+
+  public List<NGSMeasurement> queryAllNGSMeasurements(List<SampleId> sampleIds) {
+    return measurementLookup.findNGSMeasurementsBySampleIds(sampleIds);
   }
 
   public Optional<ProteomicsMeasurement> findProteomicsMeasurement(String measurementId) {
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementLookupService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementLookupService.java
@@ -68,6 +68,14 @@ public class MeasurementLookupService {
     return new ArrayList<>(termList);
   }
 
+  public long countNGSMeasurementsBySampleIds(Collection<SampleId> sampleIds) {
+    return measurementLookup.countNgsMeasurementsBySampleIds(sampleIds);
+  }
+
+  public long countProteomicsMeasurementsBySampleIds(Collection<SampleId> sampleIds) {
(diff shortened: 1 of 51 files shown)


project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementLookupService.java
```diff
@@ -68,6 +68,14 @@ public List<NGSMeasurement> queryNGSMeasurementsBySampleIds(String termFilter,
     return new ArrayList<>(termList);
   }
 
+  public long countNGSMeasurementsBySampleIds(Collection<SampleId> sampleIds) {
```
wow-such-code, line 71: This seems not to be used by the project view, instead each experiment is used to count sample ids - might this be faster? also: why was it changed in this PR?

wow-such-code: ah, nevermind

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectDetailsComponent.java
```diff
@@ -158,7 +158,11 @@ private static List<Entry> extractContactInfo(Project project) {
     var projectResponsible = new Div();
     project.getResponsiblePerson()
         .ifPresentOrElse(person -> projectResponsible.add(generateContactContainer(person)),
-            () -> projectResponsible.add(createNoPersonAssignedSpan()));
+            () -> {
+              Span noProjectResponsibleSelected = new Span("No Project Responsible Selected");
```
wow-such-code, line 162: nice

user-interface/src/main/java/life/qbic/datamanager/views/general/Tag.java
```diff
@@ -6,11 +6,41 @@ public class Tag extends Span {
...
   }
 
   public Tag(String text, String toolTip) {
     this(text);
     setTitle(toolTip);
   }
+
+  public void setTagColor(TagColor tagColor) {
+    setClassName("tag");
+    addClassName(tagColor.getColor());
```
wow-such-code, line 20: "tag" class should already be set in the constructor?

wow-such-code (approved):
Looks good to me! Only a NIT comment.

user-interface/src/main/java/life/qbic/datamanager/views/general/Tag.java
Steffengreiner: This is a sneaky trick on my part. `setclassName` replaces all current css classes set within the tag with the basic "tag" class. This allows us to remove any previously set color for this tag :)

PR #562, opened 2024-05-08
Only allow creation of complete experimental groups

wow-such-code:
Fixes an issue where the test for incomplete groups - those without all defined variables - would not be checked upon edit or creation of groups


Steffengreiner (approved):
Well done 🙌

PR #563, opened 2024-05-08
Provide profile page in which user can set their user name

Steffengreiner:
**What was changed** 
Provide dedicated page showing the user profile with the ability to change the user name of the user.

**Technical difficulties**
1.) Spring uses the "username" within the `UserDetails` interface to validate the user access in the ACL context. 
This clashes however with our current definition of UserName in the user sphere. 
Currently the `user` entity has a column username, which is not returned via the getUserName() method of the `QbicUserDetails` class. 

`
  @Override
  public String getUsername() {
    // This is what is returned by the authentication module and used by ACL classes
    // We want the id instead of email address
    return user.id();
  }
`

There are 2 possibilities to fix this:

A) We replace the `UserName` property with a similiar word such as `alias` in our data manager app.
 This PR currently does this for the frontend beans
 
B) We override the spring ACL username context with our own solution.

2.) Additionally, the `UserDetails` provided by the sessioncontext is not updated until the session context is reloaded. Could this be an issue?


project-management/src/main/java/life/qbic/projectmanagement/application/authorization/QbicUserDetails.java
```diff
@@ -57,6 +57,17 @@ public String getUsername() {
...
   }
 
+  /**
+   * Returns the unique user display name displayed to the user. Has to be distinguished from the
+   * Spring default method getUserName() which returns the userId to which the ACL permissions are
+   * linked
+   *
+   * @return the unique user display name defined to the user
+   */
+  public String getUserDisplayName() {
```
sven1103, line 67: scheme is #property()

sven1103: maybe we go for #platformUserName, display is a little bit confusing i have to admin although I have suggested it (shame on me)

project-management/src/main/java/life/qbic/projectmanagement/application/authorization/User.java
```diff
@@ -5,7 +5,8 @@
  *
  * @since 1.0.0
  */
-public record User(String id, String fullName, String emailAddress, String encryptedPassword,
+public record User(String id, String fullName, String displayUserName, String emailAddress,
```
sven1103, line 8: platformUserName

sven1103 (changes requested):
Sorry, some very minor things, i know I am picky with the naming of things....

sven1103 (approved):
Nice work!

PR #565, opened 2024-05-10
Validate digestion method when updating proteomics measurements

wow-such-code:
* Fix typos in the Digestion Method dropdown of the template
* Specifically validate Digestion Method inputs when updating measurements, as the template validation cannot be guaranteed


Steffengreiner (approved):
Well done 🤚

PR #567, opened 2024-05-16
Enable ACL cache

sven1103:
Previously, the cache was not active and therefore we did not make actual use of the perfomance boost of cached ACL queries, which is especially helpful when users are navigating within the same project.

This PR enables the cache and also introduces a cache logger for debugging purposes.

### DIFF
--- a/user-interface/pom.xml
+++ b/user-interface/pom.xml
@@ -123,6 +123,11 @@
       <classifier>jakarta</classifier>
     </dependency>
     <dependency>
+      <groupId>javax.cache</groupId>
+      <artifactId>cache-api</artifactId>
+      <version>1.1.0</version>
+    </dependency>
+    <dependency>
       <groupId>com.mysql</groupId>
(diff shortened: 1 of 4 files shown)


user-interface/pom.xml
```diff
@@ -122,6 +122,11 @@
       <artifactId>ehcache</artifactId>
       <classifier>jakarta</classifier>
     </dependency>
+    <dependency>
+      <groupId>javax.cache</groupId>
+      <artifactId>cache-api</artifactId>
+      <version>1.1.0</version>
+    </dependency>
```
Steffengreiner, line 129: NIT: Is this dependency really necessary? I don't see it used in the changed code segment. 
If it's necessary why not use the "newest" [1.1.1 ](https://mvnrepository.com/artifact/javax.cache/cache-api/1.1.1)version?

Steffengreiner (approved):
Nicely done! I have one question concerning the introduced dependency however 👀

PR #569, opened 2024-05-17
Provide Filtering for all shown properties in the projectcollectioncomponent

Steffengreiner:
**What was changed**

Load the data provided in the newly introduced SQL View(#568 ) into the projectcollection and provide filtering for all shown properties.  

**Open Questions**

The last modification date is stored as an instant in the database but we show it dependent on the localdate time.
 To enable the proper filtering, the specification would need to have the clientdetailsprovider with the correct datetime formatter. 
 This has not been implemented yet and the component shows the lastmodification instant without formatting.


sven1103 (approved):
Great, well done!

PR #570, opened 2024-05-17
Introduces local domain event dispatcher

sven1103:


wow-such-code (changes requested):
Thank you for the contribution, all of this makes sense. But it would be very useful to provide an example of how the original local event in the aggregate is created in the update use case.

sven1103:
Hi @wow-such-code , but there is an example implemented already, for the Measurement Update in the PXP case :/

wow-such-code (dismissed):

wow-such-code (approved):

PR #573, opened 2024-05-24
Enable pool metadata property variation among samples

sven1103:
So far pooled measurement metadata have been merged, which did not account for unique sample related properties, such as fraction and label.

This PR introduces the measurement notion of measurement specifict metadata, which is going to reflect that and enables an low-pain effort for adjustments in the future.


project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementProteomicsValidator.java
```diff
@@ -95,7 +95,7 @@ public ValidationResult validate(ProteomicsMeasurementMetadata measurementMetada
       return mandatoryValidationResult;
     }
     //If all fields were filled then we can validate the entries individually
-    return validationPolicy.validateSampleIds(measurementMetadata.sampleCodes())
+    return validationPolicy.validateSampleId(measurementMetadata.sampleCode())
```
KochTobi, line 98: I don't understand this change. As far as I understand it, this method ('validate') ist called by the ValidationService to validate one sample. Previously this would have validated all of them? 
Can you help me understand why this was validating multiple IDs and why this changed?

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementProteomicsValidator.java
```diff
@@ -112,15 +112,14 @@ public ValidationResult validate(ProteomicsMeasurementMetadata measurementMetada
...
-            ValidationResult::combine).combine(validationPolicy.validateMeasurementId(
-                metadata.measurementIdentifier().orElse(""))
+    return validationPolicy.validateSampleId(metadata.associatedSample())
+        .combine(validationPolicy.validationProjectRelation(metadata.associatedSample(), projectId))
+        .combine(validationPolicy.validateMeasurementId(metadata.measurementIdentifier().orElse(""))
             .combine(validationPolicy.validateMandatoryDataForUpdate(metadata))
             .combine(validationPolicy.validateOrganisation(metadata.organisationId())
                 .combine(validationPolicy.validateInstrument(metadata.instrumentCURI())
-                    .combine(validationPolicy.validateDigestionMethod(metadata.digestionMethod())))));
+                    .combine(
```
KochTobi, line 121: Is this nesting of combine methods intentional?

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementProteomicsValidator.java
KochTobi: I just noticed the switch from multiple sample ids in the metadata to only one. Please consider this comment solved. This makes the code more clear.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/Labeling.java
```diff
@@ -7,6 +7,6 @@
  *
  * @since 1.0.0
  */
-public record Labeling(String sampleCode, String labelType, String label) {
+public record Labeling(String labelType, String label) {
```
KochTobi, line 10: Please adapt the javadoc to reflect the change. This record no longer associates sample code and labelling but only represents labelling.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -342,7 +236,8 @@ public Result<MeasurementId, ErrorCode> update(ProjectId projectId,
   }
 
   private Result<MeasurementId, ErrorCode> updatePxP(ProteomicsMeasurementMetadata metadata) {
-    List<DomainEvent> domainEventsCache = new ArrayList<>();
+
+    /*List<DomainEvent> domainEventsCache = new ArrayList<>();
```
KochTobi, line 240: Is this code commented out intentionally? Can it be removed? Where did the functionality of updating PxP go?

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -568,7 +440,7 @@ private Map<NGSMeasurement, Collection<SampleIdCodeEntry>> prepareNGSMeasurement
 
     metadata.assignedSamplePoolGroup()
         .ifPresent(measurement::setSamplePoolGroup);
-    ngsMeasurements.put(measurement, sampleIdCodeEntries);
+    ngsMeasurements.put(measurement, List.of(sampleIdCodeEntries.get()));
```
KochTobi, line 443: NIT: consider replacing Optional:get with Optional:orElseThrow

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -582,78 +454,126 @@ private List<MeasurementId> performRegistrationPxp(
...
+   *
+   * @param groupedMetadata already grouped measurement metadata by pool
+   * @param projectId       the project the measurement belongs to
+   * @return
+   * @since 1.0.0
+   */
+  private Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> mergeByPool(
+      Map<String, List<ProteomicsMeasurementMetadata>> groupedMetadata, ProjectId projectId) {
+    Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> metadataMap = new HashMap<>();
+    for (String poolLabel : groupedMetadata.keySet()) {
```
KochTobi, line 488: As you are using both the key and value, please iterate over the entrySet instead and access entry.key() and entry.value().

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -582,78 +454,126 @@ private List<MeasurementId> performRegistrationPxp(
...
+   * @param projectId    the project the measurement belongs to
+   * @return
+   * @since 1.0.0
+   */
+  private Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> build(
+      List<ProteomicsMeasurementMetadata> metadataList, ProjectId projectId) {
+    Map<SampleCode, SampleIdCodeEntry> sampleIdLookupTable = buildSampleIdLookupTable(metadataList);
+    var sampleCodes = sampleIdLookupTable.keySet();
+    var specificMetadata = createSpecificMetadata(metadataList, sampleIdLookupTable);
+    var assignedMeasurementCode = MeasurementCode.createMS(sampleCodes.iterator().next().code());
```
KochTobi, line 525: Accoding to the JavaDoc of the Set#iterator():
> Returns an iterator over the elements in this set. The elements are returned in no particular order (unless this set is an instance of some class that provides a guarantee)

I am unsure what the assigned measurementCode needs in regards of sample codes and it might not matter which of the sample codes it gets. But be aware that this is the behaviour. You can leave it as is if this is what you expect.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -582,78 +454,126 @@ private List<MeasurementId> performRegistrationPxp(
...
-    measurement.setFraction(metadata.fractionName());
-    proteomicsMeasurements.put(measurement, sampleIdCodeEntries);
-    return proteomicsMeasurements;
+  private List<ProteomicsSpecificMeasurementMetadata> createSpecificMetadata(
+      List<ProteomicsMeasurementMetadata> metadata,
+      Map<SampleCode, SampleIdCodeEntry> sampleIdCodeLookupTable) {
+    return metadata.stream().map(metadataEntry -> ProteomicsSpecificMeasurementMetadata.create(
+        sampleIdCodeLookupTable.get(metadataEntry.associatedSample()).sampleId(),
+        metadataEntry.labeling().label(), metadataEntry.fractionName(),
+        metadataEntry.comment())).toList();
```
KochTobi, line 560: Did you decide to keep the comment in? I thought you mentioned you wanted to remove the comment field.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -582,78 +454,126 @@ private List<MeasurementId> performRegistrationPxp(
...
   }
 
+
   private void runPreRegistrationChecks(
-      List<? extends MeasurementMetadata> mergedSamplePoolGroups, ProjectId projectId)
+      List<? extends MeasurementMetadata> measurements, ProjectId projectId)
```
KochTobi, line 565: nice method change :+1:

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -665,6 +585,23 @@ public CompletableFuture<List<Result<MeasurementId, ErrorCode>>> updateAll(
       List<MeasurementMetadata> measurementMetadataList, ProjectId projectId) {
     List<Result<MeasurementId, ErrorCode>> results;
 
+    if (!measurementMetadataList.isEmpty() && measurementMetadataList.get(
```
KochTobi, line 588: as this method became more compley, it would be helpful to add some javadoc to the method

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -665,6 +585,23 @@ public CompletableFuture<List<Result<MeasurementId, ErrorCode>>> updateAll(
...
+      for (MeasurementMetadata measurementMetadata : measurementMetadataList) {
+        metadata.add((ProteomicsMeasurementMetadata) measurementMetadata);
+      }
+      try {
+        results = updateAllPxp(metadata, projectId);
+      } catch (MeasurementRegistrationException e) {
+        log.error("Measurement update failed.", e);
+        return CompletableFuture.completedFuture(List.of(Result.fromError(e.reason)));
+      }
+      return CompletableFuture.completedFuture(results);
```
KochTobi, line 600: Why do we need to work with completable future here? o.O

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -674,6 +611,124 @@ public CompletableFuture<List<Result<MeasurementId, ErrorCode>>> updateAll(
...
+      throw new MeasurementRegistrationException(ErrorCode.MISSING_MEASUREMENT_ID);
+    }
+    if (!allMeasurementCodesExist(
+        metadata.stream().map(ProteomicsMeasurementMetadata::measurementId).toList())) {
+      throw new MeasurementRegistrationException(ErrorCode.UNKNOWN_MEASUREMENT);
+    }
+
+    var singleSampleMeasurements = metadata.stream()
+        .filter(measurement -> measurement.assignedSamplePoolGroup().isEmpty()).toList();
+    var pooledMeasurements = metadata.stream()
```
KochTobi, line 649: what happens with measurements where a pool group is provided but no other samples are included in the given pool group? Does it matter that you categorize here?
It looks to me as if some code is duplicated here.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -695,7 +750,7 @@ protected List<Result<MeasurementId, ErrorCode>> performUpdate(
     }
     try {
       if (measurementMetadataList.get(0) instanceof ProteomicsMeasurementMetadata) {
-        return performUpdatePxp(measurementMetadataList, projectId);
+        return Collections.emptyList(); // TODO Fix for functionality
```
KochTobi, line 753: What does this comment mean? Can you elaborate? I thought the update case is tackled by updateAll ?

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/ProteomicsMeasurement.java
```diff
@@ -162,51 +158,43 @@ private static void evaluateMandatoryMetadata(ProteomicsMethodMetadata method)
...
+      MeasurementCode measurementCode, Organisation organisation, ProteomicsMethodMetadata method,
+      Collection<ProteomicsSpecificMeasurementMetadata> proteomicsSpecificMeasurementMetadata)
       throws IllegalArgumentException {
-    if (sampleIds.isEmpty()) {
-      throw new IllegalArgumentException(
-          "No sample ids provided. At least one sample id must provided for a measurement.");
-    }
     requireNonNull(method.instrument());
     requireNonNull(measurementCode);
+    requireNonNull(proteomicsSpecificMeasurementMetadata);
```
KochTobi, line 171:
```suggestion
    requireNonNull(proteomicsSpecificMeasurementMetadata, "proteomicsSpecificMeasurementMetadata must not be null");
```
As Java introduced helpful NullPointer messages this is NIT!
This is low effort and can help in debugging.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -349,11 +343,20 @@ public void addListener(ComponentEventListener<MeasurementSelectionChangedEvent>
     listeners.add(listener);
   }
 
+  //TODO introduce custom tab with label and updateable count
```
KochTobi, line 346: Is this still a TODO? How does this belong to your PR?

KochTobi (changes requested):
Hi Sven
A lot of changes but overall I think you managed to simplify the implementation a lot. Thank you for your hard work on this!
There are some points I would like to hear your thoughts.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: Async service API

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: Then the pool is of size one, which is up to the user to define if it is meaningful or not.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/ProteomicsMeasurement.java
sven1103: any references for reading up?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
sven1103: not at all :/

sven1103: might come from @Steffengreiner s contribution

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: Can you explain, what the issue is with the current implementation and what your's does improve? Otherwise I consider it as different flavor and thus as not relevant for the PR

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: which sample code is taken in a pool is irrelevant, since the semantic encoding is already broken due to the 1:n relationship of measurement and sample it derives. 
A solution would be to refactor the measurement code completely and not include the sample code at all. Out of scope of this PR though

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: We need to go over in a separate PR when I talked to the stakeholders again. Need to check the mandatory fields, the templates, etc. too much for this PR

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
KochTobi: Here you loop over a map. When looping over the keys and getting the value in every iteration, you access the map 2 times. When iterating over the entries and extracting key and value from that you only loop once. My comment is in response to a Sonarlint recommendation for performance. As I did not think that this needs a big change I thought let's take all the performance we can get 😄 
Feel free to leave it like this as I do not think that the map will be to big and I do not have clear data on performance changes.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
KochTobi: thank you for the clarification.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
KochTobi: Yes but why do we need to work asynchronously here? Can the validation not be done in one sequential way? This way we could fail earlier as we do not have to wait for all the validations to go through. Did I overlook something?

As this was this way before your PR and the PR is already large and addressing another problem I am fine with leaving it for now.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
KochTobi: Yes but the code does not need to differentiate. You have a lot of duplicate logic in the two if branches.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/ProteomicsMeasurement.java
KochTobi: Sure:
The require non null method accepts a helpful message. Prior to Java 14, there was no information on what was null but only a generic nullpointer message. 

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-
https://openjdk.org/jeps/358

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: The API is of async, non-blocking, nature to enable the UI to continue and not block the user.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
sven1103: Apologies, I dont understand the issue with the referenced code line

KochTobi (dismissed):

PR #576, opened 2024-05-27
Separate Spring Datasources

KochTobi:
Adds different datasources for 

- `life.qbic.finance`
- `life.qbic.identity`
- `life.qbic.projectmanagement`

With `life.qbic.projectmanagement` being the primary datasource


KochTobi:
tables are created and application works BUT: ProjectOverviewLookup does not find any project anymore. need to investigate here.

KochTobi:
> tables are created and application works BUT: ProjectOverviewLookup does not find any projecst anymore. need to investigate here.

A database view was missing as it turned out.

user-interface/src/main/java/life/qbic/datamanager/Application.java
```diff
@@ -42,8 +41,7 @@
     shortName = "Data Manager",
     offlineResources = {"images/logo.png"})
 @NpmPackage(value = "line-awesome", version = "1.3.0")
-@ComponentScan({"life.qbic"})
-@EntityScan(basePackages = "life.qbic")
```
sven1103, line 46: so this was preventing the data sources from being found?

sven1103 (approved):
Really good work @KochTobi !

user-interface/src/main/java/life/qbic/datamanager/Application.java
KochTobi: Yes. It created the datasources and injected them prior to the specific configurations taking place.

PR #577, opened 2024-05-28
chore/proteomics-metadata-requirements

sven1103:


KochTobi (changes requested):
Hi @sven1103 
There is only one point: In the spreadsheet the column name `Organisation id` and the description of `Organisation` differ in name.

KochTobi (approved):

PR #581, opened 2024-05-31
Listing collaborators only requires READ role and more expressive access denied message

sven1103:
Listing collaborators is now allowed with READ permission in a project.

Also, I adjusted the exception handling, such that for ACCESS DENIED, the user no gets a more expressive message instead of a general error message.


KochTobi (approved):
LGTM, nice job!

PR #582, opened 2024-05-31
Remove biological replicate from concept

sven1103:
We came to the conclusion together with some stakeholders, that the concept of the biological replicate
is too complex and also technically hard to solve. It leads to a drop in UX, since it makes the sample
registration process more complex and it is hard to find human readable representations of biological id references.

The user is able to provide an organism ID for a sample, so we leave it to the user to use custom labels to identify true biological replicates for
any given condition.


KochTobi:
Dear @sven1103 👋  
Some tests still fail.

KochTobi (changes requested):
Dear @sven1103 
When defining experimental groups, a size can be set. This stays the same? If so, we should remove the phrasing of `biological replicate` from the code as well and change it to `group size` or similar to remove the concept of a `biological replicate`.

Please remove classes that are not required anymore `life.qbic.projectmanagement.domain.model.experiment.BiologicalReplicateId` and `life.qbic.projectmanagement.domain.model.experiment.BiologicalReplicate`.

sven1103:
The statistical concept is not affected by our refactor. The sample size represents the number of biological replicates.

KochTobi (changes requested):
Looks good. Please adjust the tests.

sven1103:
Argh, Groovy imports ...

KochTobi (approved):

PR #584, opened 2024-05-31
Show pooling specific measurement information in dedicated dialog

Steffengreiner:
**What was changed**
1.) Moved pooling information into a distinct dialog, openable via extendable icon in sample ID column in measurmentgrid 
2.) Remove GridItemDetails within measurement grid and move information into columns within the grid

**Open Questions**
1.) What should be shown in NGS pooling dialog? 
(Currently there is nothing implemented as sample specific metadata information for NGS)
2.) Are all fields still relevant as columns within NGS and proteomics? The grid is horizontally scrollable, but this is cumbersome. 

**Technical Issues**
1.) There is a lot of duplicate code which could be moved into a distinct outer class. 
2.) A dedicated SQL View might be a lot faster and host the frontend information directly, making the grid filterable as well. Currently the database is queried a lot since e.g. the sample information has to be gathered seperately

### DIFF
--- a/user-interface/frontend/themes/datamanager/components/dialog.css
+++ b/user-interface/frontend/themes/datamanager/components/dialog.css
@@ -11,7 +11,7 @@ vaadin-dialog-overlay::part(footer) {
 }
 
 vaadin-dialog-overlay::part(header) {
-  padding: 2rem 1rem 1rem 4rem;
+  padding: 2rem 4rem 1rem 4rem;
 }
 
 vaadin-dialog-overlay::part(title) {
@@ -432,6 +432,32 @@ vaadin-dialog-overlay::part(title) {
 
 .change-user-details-dialog .change-user-name {
   width: 100%;
+}
+
+.measurement-pooled-samples-dialog::part(overlay) {
+  min-width: 50vw;
+}
+
+.measurement-pooled-samples-dialog::part(content) {
+  display: flex;
+  flex-direction: column;
+  gap: var(--lumo-space-l);
+}
+
+.measurement-pooled-samples-dialog .pooled-measurement-details {
+  gap: var(--lumo-space-xs);
+  display: flex;
+  flex-direction: column;
+}
+
+.measurement-pooled-samples-dialog .pooled-measurement-details .pooled-detail {
+  gap: var(--lumo-space-s);
+  display: inline-flex;
+}
+
+.measurement-pooled-samples-dialog .pooled-measurement-details .pooled-detail .label {
+  color: var(--lumo-secondary-text-color);
+  font-weight: bold;
 }
 
 .measurement-upload-dialog::part(overlay) {
--- a/user-interface/frontend/themes/datamanager/components/page-area.css
+++ b/user-interface/frontend/themes/datamanager/components/page-area.css
@@ -145,13 +145,19 @@
 
 .measurement-details-component .measurement-grid {
   height: 100%;
-  width: 100%;
 }
 
-.measurement-details-component .measurement-grid .sample-code-column {
-  display: flex;
-  flex-direction: column;
-  row-gap: var(--lumo-space-s);
+.measurement-details-component .measurement-grid .sample-column-cell {
+  display: inline-flex;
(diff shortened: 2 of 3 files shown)


sven1103 (dismissed):
Nice 1, lets show it to the users!

user-interface/src/main/java/life/qbic/datamanager/views/general/CopyToClipBoardComponent.java
```diff
@@ -0,0 +1,145 @@
...
+      }
+      ui.access(() -> {
+        copyIcon.setVisible(true);
+        copySuccessIcon.setVisible(false);
+        ui.push();
+      });
+    }).start();
+  }
+
+  public void setSuccessIconTimer(long miliseconds) {
```
sven1103, line 86: Method that is not in use. Still required?

user-interface/src/main/java/life/qbic/datamanager/views/general/CopyToClipBoardComponent.java
```diff
@@ -0,0 +1,145 @@
...
+      } catch (InterruptedException ex) {
+        Thread.currentThread().interrupt();
+        log.debug("CopyToClipboard thread interrupted due to{}", ex.getMessage());
+      }
+      ui.access(() -> {
+        copyIcon.setVisible(true);
+        copySuccessIcon.setVisible(false);
+        ui.push();
+      });
+    }).start();
```
sven1103, line 83: The creation of a thread is a very resource expensive task and we should try to avoid it, since we also risk to overload our system with to many open threads. The simplest way is to wrap in a CompletableFuture object, and call its join method. Completable Future use the ForkJoinPool. commonPool(), which we are happy with for the moment.

sven1103 (changes requested):

user-interface/src/main/java/life/qbic/datamanager/views/general/CopyToClipBoardComponent.java
Steffengreiner: Ah good catch! I'll remove it since there is currently no need for it 🤔

user-interface/src/main/java/life/qbic/datamanager/views/general/CopyToClipBoardComponent.java
```diff
@@ -0,0 +1,134 @@
...
+    fireEvent(new SwitchToSuccessFullCopyIconEvent(this, componentEvent.isFromClient()));
+    UI.getCurrent().getPage().executeJs("window.copyToClipboard($0)", textToBeCopied);
+    copyIcon.setVisible(false);
+    copySuccessIcon.setVisible(true);
+    // reset copy view after a specific time
+    UI ui = UI.getCurrent();
+    ui.getPushConfiguration().setPushMode(PushMode.MANUAL);
+    Executor delayedExecutor = CompletableFuture.delayedExecutor(showSuccessTime,
+        TimeUnit.MILLISECONDS);
+    CompletableFuture.runAsync(() -> ui.access(() -> {
```
sven1103, line 70: nice, I like it !

user-interface/src/main/java/life/qbic/datamanager/views/general/CopyToClipBoardComponent.java
```diff
@@ -0,0 +1,133 @@
...
+ * successTimer to an icon indicating the success of the copy process. The switch between the copy
+ * icon and the success icon can be listened to via the {@link SwitchToCopyIconEvent} and
+ * {@link SwitchToSuccessFullCopyIconEvent}
+ */
+@JsModule("./javascript/copytoclipboard.js")
+public class CopyToClipBoardComponent extends Span {
+
+  private final Icon copyIcon;
+  private final Icon copySuccessIcon;
+  private final long showSuccessTime = 1000;
```
sven1103, line 31: please convert it to a constant

sven1103 (changes requested):

sven1103 (approved):
Great feature!

PR #585, opened 2024-06-04
Enable download of URL file

sven1103:
The wrong ID type has been passed to the ACL protected service method (experiment ID instead of project ID).

Thus the authorization always failed.

This PR passes the project ID to enable the correct access policy.


KochTobi (approved):
😮

PR #588, opened 2024-06-05
Provide UTF-16 support for measurement metadata

sven1103:
Microsoft Excel currently only supports exports to CSV UTF-8 format (which we do not support), TSV in an unknown format (thank you Microsoft) and Unicode UTF-16.

The latter uses the tab character as separator, which is exactly what we want. 

This PR reads in any file content for measurement metadata assuming UTF-16, which contains UTF-8 and thus should work for the most symbols used currently in production environments.


KochTobi (approved):
LGTM

PR #592, opened 2024-06-05
Enable pooled sample measurement for NGS

sven1103:


project-management/src/main/java/life/qbic/projectmanagement/application/dataset/RawDataService.java
```diff
@@ -114,7 +115,8 @@ public Collection<RawData> findNGSRawData(String filter,
                       .equals(datasetInformation.measurementCode()))
               .findFirst().orElseThrow();
           var sampleInformation = sampleInformationService.retrieveSamplesByIds(
-                  measurement.measuredSamples()).stream()
+                  measurement.specificMeasurementMetadata().stream().map(
+                      NGSSpecificMeasurementMetadata::measuredSample).toList()).stream()
```
KochTobi, line 119:
```suggestion
                  measurement.specificMeasurementMetadata().stream()
                  .map(NGSSpecificMeasurementMetadata::measuredSample)
                  .toList())
                  .stream()
```
When you put the consecutive stream calls in a new line each it is easy to see that here, a stream is collected into a list only to be streamed directly after again.
You can extract a variable here to store the list with a meaningful name or you can remove the `.toList().stream()` chaining.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -317,54 +238,31 @@ protected List<MeasurementId> performRegistration(
...
+        ngsMeasurementMetadata.add((NGSMeasurementMetadata) measurementMetadata);
       }
     }
-    Map<NGSMeasurement, Collection<SampleIdCodeEntry>> ngsMeasurementsMapping = new HashMap<>();
-    for (NGSMeasurementMetadata metadata : ngsMeasurements) {
-      ngsMeasurementsMapping.putAll(prepareNGSMeasurement(projectId, metadata));
-    }
-    return measurementDomainService.addNGSAll(ngsMeasurementsMapping);
-  }
+    Map<NGSMeasurement, Collection<SampleIdCodeEntry>> NGSmeasurementsMapping = new HashMap<>();
```
KochTobi, line 247: variable names should start with a lower case letter. 
```suggestion
    Map<NGSMeasurement, Collection<SampleIdCodeEntry>> ngsMeasurementsMapping = new HashMap<>();
```

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -317,54 +238,31 @@ protected List<MeasurementId> performRegistration(
   private List<MeasurementId> performRegistrationNGS(
       List<MeasurementMetadata> measurementMetadataList,
       ProjectId projectId) {
-    List<NGSMeasurementMetadata> ngsMeasurements = new ArrayList<>();
+    List<NGSMeasurementMetadata> ngsMeasurementMetadata = new ArrayList<>();
     for (MeasurementMetadata measurementMetadata : measurementMetadataList) {
       if (measurementMetadata instanceof NGSMeasurementMetadata) {
-        ngsMeasurements.add((NGSMeasurementMetadata) measurementMetadata);
+        ngsMeasurementMetadata.add((NGSMeasurementMetadata) measurementMetadata);
```
KochTobi, line 244: As of Java 16 you do not need to cast here. You can directly declare your variable of type in the if statement.
```java
if (measurementMetadata instanceof NGSMeasurementMetadata ngsMetadata) {
  ngsMeasurementMetadata.add(ngsMetadata);
```
https://openjdk.org/jeps/394

KochTobi: The same can be done for the proteomics case ;)

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -414,10 +312,55 @@ private Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> mergeByPool(
     return metadataMap;
   }
 
+  private Map<NGSMeasurement, Collection<SampleIdCodeEntry>> mergeByPoolNGS(
```
KochTobi, line 315: I think you should rename the proteomics method accordingly. `mergeByPoolPxP` instead of `mergeByPool`.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -414,10 +312,55 @@ private Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> mergeByPool(
...
+      Map<String, List<NGSMeasurementMetadata>> groupedMetadata, ProjectId projectId) {
+    Map<NGSMeasurement, Collection<SampleIdCodeEntry>> metadataMap = new HashMap<>();
+    for (var metadataGroup : groupedMetadata.entrySet()) {
+      metadataMap.putAll(buildNGS(metadataGroup.getValue(), projectId));
+    }
+    return metadataMap;
+  }
+
+  private Map<NGSMeasurement, Collection<SampleIdCodeEntry>> buildNGS(
```
KochTobi, line 324: Please rename the proteomics method accordingly. `buildPxp` instead of `build`.

As you are using the same method structure except for the type of grouped metadata and the call to a different build method, maybe there is some abstraction waiting here? For now keeping it separate makes sense :+1:

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -414,10 +312,55 @@ private Map<ProteomicsMeasurement, Collection<SampleIdCodeEntry>> mergeByPool(
...
+
+    NGSMeasurement measurement;
+    if (firstMetadataEntry.assignedSamplePoolGroup().isPresent()) {
+      measurement = NGSMeasurement.createWithPool(projectId,
+          firstMetadataEntry.assignedSamplePoolGroup().get(), assignedMeasurementCode,
+          organisationQuery.get(), method, specificMetadata);
+    } else {
+      measurement = NGSMeasurement.createSingleMeasurement(projectId, assignedMeasurementCode,
+          organisationQuery.get(), method, specificMetadata.get(0));
+    }
```
KochTobi, line 355: For proteomics there is no difference here. In both cases the `create` method is called. What is the difference to the NGS case?

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSIndex.java
```diff
@@ -0,0 +1,19 @@
+package life.qbic.projectmanagement.domain.model.measurement;
+
+/**
+ * <b><record short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
```
KochTobi, line 9: What is an `i5` or `i7` index? Providing some description here might help developers in the future if they are not aware of it.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
```diff
@@ -114,35 +114,58 @@ private static void evaluateMandatoryMetadata(NGSMethodMetadata method)
...
+  public static NGSMeasurement createWithPool(ProjectId projectId, String samplePool,
       MeasurementCode measurementCode, Organisation organisation, NGSMethodMetadata method,
-      String comment) throws IllegalArgumentException {
-    if (sampleIds.isEmpty()) {
-      throw new IllegalArgumentException(
-          "No sample ids provided. At least one sample id must provided for a measurement.");
+      Collection<NGSSpecificMeasurementMetadata> specificMeasurementMetadata)
+      throws IllegalArgumentException {
+    requireNonNull(measurementCode);
+    requireNonNull(method);
```
KochTobi, line 134:
```suggestion
    requireNonNull(measurementCode, "measurement Code must not be null");
    requireNonNull(method, "method must not be null");
```

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
```diff
@@ -114,35 +114,58 @@ private static void evaluateMandatoryMetadata(NGSMethodMetadata method)
...
+    requireNonNull(method.instrument());
```
KochTobi, line 135:
```suggestion
    requireNonNull(method.instrument(), "instrument must not be null");
```

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
```diff
@@ -114,35 +114,58 @@ private static void evaluateMandatoryMetadata(NGSMethodMetadata method)
...
+   * @param specificMeasurementMetadata sample specific metadata
+   * @return an instance of an {@link NGSMeasurement}
+   * @since 1.0.0
+   */
+  public static NGSMeasurement createSingleMeasurement(ProjectId projectId,
+      MeasurementCode measurementCode, Organisation organisation, NGSMethodMetadata method,
+      NGSSpecificMeasurementMetadata specificMeasurementMetadata) {
+    requireNonNull(measurementCode);
+    requireNonNull(method);
+    requireNonNull(method.instrument());
```
KochTobi, line 166: NIT: use the error message in the `Objects#requireNonNull`

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
```diff
@@ -114,35 +114,58 @@ private static void evaluateMandatoryMetadata(NGSMethodMetadata method)
...
+    return new NGSMeasurement(MeasurementId.create(), projectId, "", measurementCode, organisation,
```
KochTobi, line 167: Why use an empty string to signal no pool present instead of null?

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
```diff
@@ -0,0 +1,103 @@
+package life.qbic.projectmanagement.domain.model.measurement;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Embeddable;
+import java.util.Objects;
+import java.util.Optional;
+import life.qbic.projectmanagement.domain.model.sample.SampleId;
+
+/**
```
KochTobi, line 9: Please add some javadoc here

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+    this.indexI7 = indexI7;
+    this.comment = comment;
+  }
+
+  public static NGSSpecificMeasurementMetadata create(SampleId sampleId, String indexI5,
+      String indexI7, String comment) {
+    Objects.requireNonNull(sampleId);
+    Objects.requireNonNull(indexI5);
+    Objects.requireNonNull(indexI7);
+    Objects.requireNonNull(comment);
```
KochTobi, line 47: Why do we require non null for the fields?

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+    // - no indexing (if no multiplexing is done)
+    if (!indexI5.isBlank() && !indexI7.isBlank()) {
+      // Double indexing
+      return new NGSSpecificMeasurementMetadata(sampleId, indexI5, indexI7, comment);
+    }
+    if (!indexI7.isBlank()) {
+      // Single indexing
+      return new NGSSpecificMeasurementMetadata(sampleId, indexI5, indexI7, comment);
+    }
+    if (!indexI5.isBlank()) {
```
KochTobi, line 60: There is temporal coupling here. I suggest adding a check that i7 is not present. Otherwise, if the first condition is removed at some point, you might throw an erronous exception. You can also get rid of the first check by combining this check with `indexI7.isBlank()`

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+    // No indexing provided
+    return new NGSSpecificMeasurementMetadata(sampleId, indexI5, indexI7, comment);
+  }
+
+  public SampleId measuredSample() {
+    return measuredSample;
+  }
+
+  public Optional<String> comment() {
+    return comment.isBlank() ? Optional.empty() : Optional.of(comment);
```
KochTobi, line 72: If you would allow null values for the comment this could be simply. `Optional.ofNullable(comment)`

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public Optional<NGSIndex> index() {
+    if (indexI7.isBlank()) {
```
KochTobi, line 76: You can use the same logic in the constructor. Makes the if statements easier to understand.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -535,28 +532,58 @@ private void setPooledNgsMeasurementDetails(
...
-          .setAutoWidth(true);
-      sampleDetailsGrid.addColumn(sampleId -> ngsMeasurement.indexI5().orElse(""))
-          .setHeader("Index I5")
-          .setTooltipGenerator(sampleId -> ngsMeasurement.indexI5().orElse(""))
+              metadata -> retrieveSampleById(metadata.measuredSample()).orElseThrow().sampleCode()
+                  .code())
           .setAutoWidth(true);
-      sampleDetailsGrid.addColumn(sampleId -> ngsMeasurement.indexI7().orElse(""))
+      sampleDetailsGrid.addColumn(metadata -> {
+            if (metadata.index().isEmpty()) {
```
KochTobi, line 553: you can use Optional#map here. This is the advantage of Optional. 
```java
metadata.index().map(index -> index.indexI7()).orElse("");
```

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -535,28 +532,58 @@ private void setPooledNgsMeasurementDetails(
...
+              return "";
+            } else {
+              return metadata.index().get().indexI7();
+            }
+          })
           .setHeader("Index I7")
-          .setTooltipGenerator(sampleId -> ngsMeasurement.indexI7().orElse(""))
+          .setTooltipGenerator(metadata -> {
```
KochTobi, line 560: This is a great place to use `Optional#map`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -535,28 +532,58 @@ private void setPooledNgsMeasurementDetails(
...
+            if (metadata.index().isEmpty()) {
+              return "";
+            } else {
+              return metadata.index().get().indexI7();
+            }
+          }).setAutoWidth(true);
+      sampleDetailsGrid.addColumn(metadata -> {
+            if (metadata.index().isEmpty()) {
```
KochTobi, line 568: `Optional.map`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -535,28 +532,58 @@ private void setPooledNgsMeasurementDetails(
...
+              return "";
+            } else {
+              return metadata.index().get().indexI5();
+            }
+          })
+          .setHeader("Index I5")
+          .setTooltipGenerator(metadata -> {
+            if (metadata.index().isEmpty()) {
```
KochTobi, line 576: `Optional.map`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementPresenter.java
```diff
@@ -42,16 +44,16 @@ private static ProteomicsMeasurementEntry convertProteomicsMeasurement(
...
     return new NGSMeasurementEntry(measurement.measurementCode().value(),
         sampleInfo, measurement.organisation().IRI(), measurement.organisation().label(),
         measurement.instrument().getName().replace("_", ":"),
         measurement.instrument().getLabel(),
         measurement.samplePoolGroup().orElse(""), measurement.facility(),
         measurement.sequencingReadType(),
         measurement.libraryKit().orElse(""), measurement.flowCell().orElse(""),
-        measurement.sequencingRunProtocol().orElse(""), measurement.indexI7().orElse(""),
-        measurement.indexI5().orElse(""), measurement.comment().orElse(""));
+        measurement.sequencingRunProtocol().orElse(""), specificMeasurementMetadata.index().orElse(new NGSIndex("", "")).indexI7(),
```
KochTobi, line 55: Optional#map to the rescue, no need to create an empty index here.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementPresenter.java
```diff
@@ -42,16 +44,16 @@ private static ProteomicsMeasurementEntry convertProteomicsMeasurement(
...
+        specificMeasurementMetadata.index().orElse(new NGSIndex("", "")).indexI5(), specificMeasurementMetadata.comment().orElse(""));
```
KochTobi, line 56: Optional#map to the rescue, no need to create an empty index here.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
KochTobi: The way you did it in the accessor `index()` is way easier to read.

KochTobi (changes requested):
Hi great changes to the code!
I have some suggestions to make the code a little easier to read.
I did not run it and tested it yet but plan to do so. This is one reason why I put my review as changes requested.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSIndex.java
```diff
@@ -0,0 +1,19 @@
...
+public record NGSIndex(String indexI5, String indexI7) {
+
+  public static NGSIndex doubleIndexing(String indexI5, String indexI7) {
```
KochTobi, line 12: It might make sense to be a litte bit more explicit here and create subclasses `record NGSDoubleIndex` and `record NGSSingleIndex`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
KochTobi: I just noticed that this might not be the case as empty index neither is an Optional nor null. I initially read the `isEmpty()` as the `isEmpty()` method of the `Optional` class. It turned out it is the `isEmpty()` method of the `String` class.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
KochTobi: I just noticed that this might not be the case as empty index neither is an Optional nor null. I initially read the `isEmpty()` as the `isEmpty()` method of the `Optional` class. It turned out it is the `isEmpty()` method of the `String` class.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
KochTobi: I just noticed that this might not be the case as empty index neither is an Optional nor null. I initially read the `isEmpty()` as the `isEmpty()` method of the `Optional` class. It turned out it is the `isEmpty()` method of the `String` class.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementPresenter.java
KochTobi: I just noticed that this might not be the case as empty index neither is an Optional nor null. I initially read the `isEmpty()` as the `isEmpty()` method of the `Optional` class. It turned out it is the `isEmpty()` method of the `String` class.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementPresenter.java
KochTobi: I just noticed that this might not be the case as empty index neither is an Optional nor null. I initially read the `isEmpty()` as the `isEmpty()` method of the `Optional` class. It turned out it is the `isEmpty()` method of the `String` class.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
Steffengreiner: Good point!

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
Steffengreiner: Valid!

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
Steffengreiner: I heavily agree that we need to abstract the measurement itself instead of having duplicated code everywhere. 👍

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
Steffengreiner: That's a very good question! I asked SF about this, the difference is that during the NGS case we realised that a more stringent differentiation between pooling and singular measurement makes sense due to a difference in mandatory properties. 
He suggested to avoid bloating this PR with changes to Proteomics however so i won't change this in this PR.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSIndex.java
Steffengreiner: Good point 👍

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
Steffengreiner: Nice catch 👀

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
Steffengreiner: Good idea 🧠

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
Steffengreiner: It's so much cleaner Thanks 🧼

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
Steffengreiner: SampleId cannot be null since each measurement has to have at least one associated sampleId.
The other properties are dependent on what the user provides in the spreadsheet, either nothing (ergo an empty string) or content to be stored sample specific in a measurement.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSMeasurement.java
Steffengreiner: I think the logic here is that we expect the property to be there, however the user does not have to provide a String within the spreadsheet.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSIndex.java
Steffengreiner: While i like the idea, i would defer the splitting up into singular indices to once we better understand on the different use cases of indices in the NGS domain 🤔

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/NGSSpecificMeasurementMetadata.java
Steffengreiner: I think the issue here is that we want to ensure that we don't create an NGS Measurement with only an I5 index. To do it as we did it in the accessor, we would have to adapt the database and the entity to expect null values for the indices.

KochTobi (approved):

PR #595, opened 2024-06-06
Add item count for grids

wow-such-code:
* Places item count badges in the header of grids with tabs (measurements and raw data)
* Fixes issue with lazy loading returning the wrong counts in these cases (thanks to @Steffengreiner )
* Fixes issue where pooled samples would lead to measurements being counted more than once
* Fixes issue where selected measurement count would be wrong for filtered measurements
* Removes tab labeled with experiment name and count from sample grid
* Add all-select logic to Raw Data Tab
* Adds Span above the sample table displaying the correct sample count of samples in the current experiment

Known issue: when adding (/removing?) samples, the sample count above the grid is not updated instantly. Since this page will receive a big rework soon, it might be a waste of time to fix this at this point? You decide.


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -233,6 +229,7 @@ private void updateSampleGridDataProvider(ExperimentId experimentId, String filt
    * @param context the context in which the user is.
    */
   public void setContext(Context context) {
+    setSampleCount(0);
```
KochTobi, line 232: Do samples depend on the experiment or the project? If they belong to the experiment this should go into setExperiment as they need to be reset everytime the experiment context changes.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -219,12 +218,9 @@ private void updateSampleGridDataProvider(ExperimentId experimentId, String filt
...
       return sampleInformationService.queryPreview(experimentId, query.getOffset(),
           query.getLimit(), List.copyOf(sortOrders), filter).stream();
-    }, query -> SampleDetailsComponent.this.sampleInformationService.countPreviews(
-        experimentId,
-        filter));
-    sampleGrid.getLazyDataView()
-        .addItemCountChangeListener(
-            countChangeEvent -> experimentTab.setSampleCount(countChangeEvent.getItemCount()));
+    });
+    sampleGrid.getLazyDataView().addItemCountChangeListener(
```
KochTobi, line 222: NIT: You are accessing the sample grid while it is modified. Consider assigning a variable rather than directly modifying the field.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -121,13 +118,15 @@ public SampleDetailsComponent(@Autowired SampleInformationService sampleInformat
     experimentTabContent.add(sampleGrid, noGroupsDefinedDisclaimer, noSamplesRegisteredDisclaimer);
 
     TabSheet sampleExperimentTabSheet = new TabSheet();
-    sampleExperimentTabSheet.add(experimentTab, experimentTabContent);
+    sampleExperimentTabSheet.add("", experimentTabContent);
```
KochTobi, line 121: Why have a tab sheet at all? There is only one tab without a name...

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -590,4 +613,49 @@ private Span pooledMeasurementEntry(String propertyLabel, String propertyValue)
...
+    public String getTabLabel() {
+      return technology;
+    }
+
+    /**
+     * Helper method for creating a badge.
+     */
+    private static Span createBadge() {
+      Tag tag = new Tag(String.valueOf(0));
+      tag.setTagColor(TagColor.CONTRAST);
```
KochTobi, line 642: Shouldn't the coloring be handled by CSS? What am I missing?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -425,8 +448,8 @@ public void addListener(ComponentEventListener<MeasurementSelectionChangedEvent>
 
   //TODO introduce custom tab with label and updateable count
```
KochTobi, line 449: What about this TODO?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -425,8 +448,8 @@ public void addListener(ComponentEventListener<MeasurementSelectionChangedEvent>
...
   public Optional<String> getSelectedTabName() {
-    if (registeredMeasurementsTabSheet.getSelectedTab() != null) {
-      return Optional.of(registeredMeasurementsTabSheet.getSelectedTab().getLabel());
+    if (registeredMeasurementsTabSheet.getSelectedTab() != null && !tabsInTabSheet.isEmpty()) {
+      return Optional.of(tabsInTabSheet.get(registeredMeasurementsTabSheet.getSelectedIndex()).getTabLabel());
```
KochTobi, line 452: Why so complex?
Doesn't this method produce the same as
```java
Optional.ofNullable(registeredMeasurementsTabSheet.getSelectedTab()).map(Tab::getLabel);
```

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -332,13 +350,18 @@ private void createProteomicsGrid() {
...
                       it.getDirection().equals(SortDirection.ASCENDING)))
               .collect(Collectors.toList());
-          // if no order is provided by the grid order by last modified (least priority)
-          sortOrders.add(SortOrder.of("measurementCode").ascending());
-          return measurementService.findProteomicsMeasurements(searchTerm,
-                  context.experimentId().orElseThrow(),
-                  query.getOffset(), query.getLimit(), sortOrders, context.projectId().orElseThrow())
-              .stream();
+            // if no order is provided by the grid order by last modified (least priority)
+            sortOrders.add(SortOrder.of("measurementCode").ascending());
```
KochTobi, line 354: There seems to be some disagreement between code comment and code.

KochTobi (changes requested):
Very nice idea with the class extraction of the measurement tabs.
Some open questions regarding the samples tab sheet and some code smells.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
wow-such-code: moved to setexperiment 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
wow-such-code: we always use TagColor like that, I'm not sure if that's necessary, though

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
wow-such-code: I tried it, but it doesn't work like that. Tabsheets are weird.

KochTobi: Did for me 😄 I tried it too

wow-such-code: It didn't correctly update the selections for me and threw Nullpointer exceptions at one point

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -51,16 +51,17 @@ private Checkbox renderCheckbox(T bean) {
 
       if (event.isFromClient()) {
```
KochTobi, line 52: This selection listener is extremely complex.
Not only is one selection listener added to the `selectAllCheckbox` for every rendered checkbox (including ones that are rendered when scrolling) and never removed. Further the logic is pretty complex and all in one lambda.
The listener on the checkbox should only concern itself with actions related to the selection. There should not be any outside value change listener be added here.
Please refactor this. The functionality of selecting all /none belongs to the `selectAllCheckBox` and not to the individual checkboxes.

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -51,16 +51,17 @@ private Checkbox renderCheckbox(T bean) {
...
-        } else if(Boolean.FALSE.equals(allBoxVal) && areAllSelected()) {
+        } else if (Boolean.FALSE.equals(allBoxVal) && areAllSelected()) {
           selectAllCheckBox.setValue(true);
         }
       }
     });
     /*Necessary to propagate selection of select all checkbox to individual checkbox*/
     selectAllCheckBox.addValueChangeListener(event -> {
-      if(event.isFromClient()) {
+      // needed so boxes that are filtered out are not selected and lead to miscount
```
KochTobi, line 63: As previously stated, this does not belong here. It is behaviour of the select all checkbox component and not of the individual checkboxes.
This makes it easier to only select attached checkboxes as well.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -144,12 +156,14 @@ private void resetTabsInTabsheet() {
 
   private void addMeasurementTab(GridLazyDataView<?> gridLazyDataView) {
     if (gridLazyDataView.getItem(0) instanceof ProteomicsMeasurement) {
```
KochTobi, line 158: What happens when the data view does not contain items? IndexOutOfBoundException?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/rawdata/RawDataDetailsComponent.java
```diff
@@ -76,10 +78,22 @@ public RawDataDetailsComponent(@Autowired RawDataService rawDataService,
    *                   be filtered for
    */
   public void setSearchedRawDataValue(String searchTerm) {
+    if (!this.searchTerm.equals(searchTerm)) {
+      refreshGrids();
+    }
```
KochTobi, line 83: So you only refresh in case the search term changed? If so, you only need to set the search term if it is different as well.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/rawdata/RawDataDetailsComponent.java
```diff
@@ -76,10 +78,22 @@ public RawDataDetailsComponent(@Autowired RawDataService rawDataService,
...
     this.searchTerm = searchTerm;
+  }
+
+  public void refreshGrids() {
+    resetSelectedMeasurements();
     rawDataGridDataViews.forEach(AbstractDataView::refreshAll);
```
KochTobi, line 89: Why extract a method here?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/rawdata/RawDataDetailsComponent.java
```diff
@@ -76,10 +78,22 @@ public RawDataDetailsComponent(@Autowired RawDataService rawDataService,
...
   }
 
+  private void resetSelectedMeasurements() {
+    proteomicsRawDataGrid.clearSelectedItems();
+    ngsRawDataGrid.clearSelectedItems();
+  }
```
KochTobi, line 95: Consider renaming it to `clearSelectedMeasurements` to match your terminology. Alternatively `clearMeasurementSelection`:wink:

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -119,15 +115,13 @@ public SampleDetailsComponent(@Autowired SampleInformationService sampleInformat
...
-    TabSheet sampleExperimentTabSheet = new TabSheet();
-    sampleExperimentTabSheet.add(experimentTab, experimentTabContent);
-    sampleExperimentTabSheet.setHeightFull();
-
+    
     Div content = new Div();
     content.addClassName("sample-details-content");
 
-    content.add(buttonAndFieldBar, metadataDownload, sampleExperimentTabSheet);
+    countSpan = new Span();
```
KochTobi, line 122: Your count span does not have any CSS class. This makes it hard in the future to style it.

KochTobi (changes requested):

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
wow-such-code: it's not the same logic, as the item index is needed to get the correct tab. I can test it again to pinpoint the problem, if you think it's worthwhile

It does also not check if tabsInTabSheet is empty, which lead to the NPE iirc

user-interface/vite.config.ts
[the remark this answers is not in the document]
wow-such-code: vaadin stuff that shouldn't be on github

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/MeasurementService.java
```diff
@@ -98,6 +98,18 @@ public boolean hasMeasurements(ExperimentId experimentId) {
...
+    var samplesInExperiment = result.getValue().stream().map(Sample::sampleId).toList();
+    return measurementLookupService.countProteomicsMeasurementsBySampleIds(samplesInExperiment);
+  }
+
+  public long countNGSMeasurements(ExperimentId experimentId) {
+    var result = sampleInformationService.retrieveSamplesForExperiment(experimentId);
+    var samplesInExperiment = result.getValue().stream().map(Sample::sampleId).toList();
+    return measurementLookupService.countNGSMeasurementsBySampleIds(samplesInExperiment);
+  }
+
```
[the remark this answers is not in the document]
wow-such-code: because these are from JPA, while the openBIS result.size() method returns int

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -332,13 +349,18 @@ private void createProteomicsGrid() {
...
                       it.getDirection().equals(SortDirection.ASCENDING)))
               .collect(Collectors.toList());
-          // if no order is provided by the grid order by last modified (least priority)
-          sortOrders.add(SortOrder.of("measurementCode").ascending());
-          return measurementService.findProteomicsMeasurements(searchTerm,
-                  context.experimentId().orElseThrow(),
-                  query.getOffset(), query.getLimit(), sortOrders, context.projectId().orElseThrow())
-              .stream();
+            // if no order is provided by the grid order by last modified (least priority)
+            sortOrders.add(SortOrder.of("measurementCode").ascending());
```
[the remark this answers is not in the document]
wow-such-code: weird 😓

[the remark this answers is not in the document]
wow-such-code: ah, different omics type

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/rawdata/RawDataDetailsComponent.java
```diff
@@ -182,7 +207,8 @@ private void createProteomicsRawDataGrid() {
                   query.getOffset(), query.getLimit(), sortOrders, context.projectId().orElseThrow())
               .stream();
         });
-    proteomicsRawDataGrid.setSelectionMode(SelectionMode.MULTI);
+    proteomicsRawDataGrid.getLazyDataView().addItemCountChangeListener(
```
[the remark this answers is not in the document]
wow-such-code: it's provided by the MultiSelectLazyLoadingGrid.java

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
wow-such-code: It's a surprisingly complex problem that took us about a week to implement the last time to get rid of all the edge cases.

* The detach listener is a great idea and works well! 👍 

* Uncoupling "select all" and the selection of the individual boxes does not work so easily, as far as I can tell. The selectAllCheckBox must also change its value when a user selects the last unselected row or when they deselect one row when all others are selected. I don't see an alternative solution that doesn't make the code even more complex.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
wow-such-code: it's only called once the views are initialized, but I added a check anyway

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
```diff
@@ -590,4 +612,49 @@ private Span pooledMeasurementEntry(String propertyLabel, String propertyValue)
     }
   }
 
+  public static class MeasurementTechnologyTab extends Tab {
```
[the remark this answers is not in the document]
wow-such-code: thought about it and i think it's more dangerous to add it: two views use this class and the fields can be the same, yet the tabs are not.

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
```diff
@@ -59,14 +59,20 @@ private Checkbox renderCheckbox(T bean) {
...
       }
     });
     /*Necessary to keep items selected even if grid pagination is reloaded*/
     box.setValue(selectedItems.contains(bean));
+
+    //if the box is removed, we remove the respective listeners
+    box.addDetachListener(detachEvent -> {
+      boxListener.remove();
+      allSelectBoxListener.remove();
+    });
```
KochTobi, line 74: :+1:

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
KochTobi: Why does the select all checkbox need to change the value when all rows are selected?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
KochTobi: When, how and what will be re-worked?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementDetailsComponent.java
wow-such-code: problem was the custom tab, which can't set the tab label. hope it's still better to read now, as I need to cast

user-interface/src/main/java/life/qbic/datamanager/views/general/MultiSelectLazyLoadingGrid.java
KochTobi: following scenario:

1. the user filters for ngs
2. the user selects all
3. the user filters for some property
4. the user has made a type
5. NOTHING is selected anymore (even if the user corrects the typo)

wow-such-code: But this would be a consequence of the filtering starting while the user is typing. And if I select everything in Google Mail and start the filter (typo or not), nothing is selected anymore, either.
If this is an issue, I suggest we create a new card for a new PR addressing it.

KochTobi: I still think this whole select all implementation is more complex than need be. However I will no longer block this PR for that as it only touches on the subject.

KochTobi (approved):

PR #598, opened 2024-06-07
update download information

wow-such-code:
* adapt curl code for downloading data
* increase icon size


user-interface/src/main/java/life/qbic/datamanager/views/general/CodeBlock.java
```diff
@@ -31,7 +31,6 @@ public CodeBlock(String command, String... parameters) {
 
   private static Icon generateCopyIcon(String copyContent) {
     Icon copyIcon = VaadinIcon.COPY_O.create();
-    copyIcon.addClassName(IconSize.SMALL);
     copyIcon.addClassNames("clickable", "copy-icon");
     copyIcon.addClickListener(
         event -> UI.getCurrent().getPage().executeJs("window.copyToClipboard($0)", copyContent));
```
Steffengreiner, line 36: I'd recommend to replace this outdated method with the new CopyToClipboard component. 

You can remove this method and the javascript annotation and just provide the new component in the codeblock constructor as follows:
` CopyToClipBoardComponent copyToClipBoardComponent = new CopyToClipBoardComponent(codeToBeCopied);`
 `copyToClipBoardComponent.setIconSize(IconSize.SMALL);`

Steffengreiner (changes requested):
I have one suggestion otherwise it's perfect 🆗

Steffengreiner (approved):
Well done 👍

PR #601, opened 2024-06-07
Show the University Tübingen Logo and a background image for all landing page layouts

Steffengreiner:
**What was changed**
The data manager application now shows a background image and the university Tübingen logo with a title and subtitle on all sites with which the user interacts without logging in.


sven1103 (changes requested):
What is the source of the background image, what is it license?

Missing: legal note entry for the image

sven1103 (approved):

PR #603, opened 2024-06-10
Fix Project Overview entity

KochTobi:
The project overview entity idicated to JPA a confusing setup with a column and join table for an element collection on the same field. This does always throw exceptions when hibernate is trying to set up the database. Further the wrong '@Immutable` annotation was used.

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
@@ -11,10 +11,9 @@ import jakarta.persistence.Table;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Objects;
 import life.qbic.projectmanagement.domain.model.project.Project;
 import life.qbic.projectmanagement.domain.model.project.ProjectId;
-import org.springframework.data.annotation.Immutable;
+import org.hibernate.annotations.Immutable;
 
 /**
  * A limited view of the more complex {@link Project}.
@@ -23,6 +22,7 @@ import org.springframework.data.annotation.Immutable;
  */
 //Views are aggregates which cannot be changed
 @Immutable
+@org.springframework.data.annotation.Immutable
 @Entity
 @Table(name = "project_overview")
 public class ProjectOverview {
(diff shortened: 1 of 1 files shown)


project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
```diff
@@ -23,6 +22,7 @@
  */
 //Views are aggregates which cannot be changed
 @Immutable
+@org.springframework.data.annotation.Immutable
```
Steffengreiner, line 25: Now the project overview has both immutable annotations? I'd expected the "wrong" immutable to be replaced?

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
```diff
@@ -63,26 +62,6 @@ protected ProjectOverview() {
...
-    this.principalInvestigatorName = Objects.requireNonNull(principalInvestigatorName,
-        "principal investigator name cannot be null");
-    this.projectManagerName = Objects.requireNonNull(projectManagerName,
-        "project manager name cannot be null");
-    this.projectResponsibleName = projectResponsibleName;
-    this.ngsMeasurementCount = ngsMeasurementCount;
-    this.pxpMeasurementCount = pxpMeasurementCount;
-    this.collaboratorUserNames = Objects.requireNonNull(usernames,
-        "The collaborator user names cannot be null");
-  }
```
Steffengreiner, line 84: I don't understand why this constructor was removed? Now the projectOverview only has a protected constructor? How are the fields initialized then 🤔

Steffengreiner (changes requested):
Heya @KochTobi Awesome work, however i'm a little confused about the changes not mentioned in the PR description

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
KochTobi: There is no way to construct it as it is only a view. Further the constructor was unused.

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
KochTobi: It might be good to have both in here so that both frameworks know of the immutability

Steffengreiner (approved):
Thanks for the explanation!

PR #609, opened 2024-06-10
Fix/collaborator user names filtering

Steffengreiner:
**What was changed**
Provide filtering in collaborator user names within the project overview ensuring that null values don't break the application within the converter


KochTobi (approved):
LGTM

PR #614, opened 2024-06-12
Provide a default profile picture

KochTobi:
Provides a default image for each user based on the users identifier (not the username as the username can change).
The goal of this PR is to make users easily identifiable by providing an identicon. See https://en.wikipedia.org/wiki/Identicon

image

image

image

image


KochTobi:
The icon still needs to show up in a couple of places

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
```diff
@@ -58,6 +58,10 @@ public class ProjectOverview {
   @Column(name = "usernames")
   private List<String> collaboratorUserNames = new ArrayList<>();
 
+  @Convert(converter = CollaboratorUserInfosConverter.class)
+  @Column(name = "userInfos")
+  private List<UserInfo> collaboratorUserInfos = new ArrayList<>();
```
Steffengreiner, line 63: This currently leads to an exception since the column does not exist in our database. 
Also the collaboratorUserNames field should be then removed here, from the specification and the sql view script to avoid redundant information imo.

Steffengreiner (changes requested):
Heya @KochTobi Awesome work! the identicons look awesome. 
Sadly i can't only check them out in the menu because neither the projectcollection nor the profile page are loading with the testdatabase. 😞

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectOverview.java
KochTobi: You have to use the data_management database 😉

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/access/UserAvatarWithNameComponent.java
```diff
@@ -0,0 +1,19 @@
...
+import com.vaadin.flow.component.html.Span;
+import life.qbic.datamanager.views.account.UserAvatar;
+
+/**
+ * A component displaying a user avatar and user name
+ */
+public class UserAvatarWithNameComponent extends Div {
+
+  public UserAvatarWithNameComponent(UserAvatar userAvatar, String userName) {
+    addClassName("avatar-with-name");
```
Steffengreiner, line 13: This assigns the same css name as the div containing the avatar and username in the userprofilecomponent which can lead to confusion. I recommend to adapt the css name to avoid confusion.
(Especially since there is currently no styling associated with this component). 

NIT: I'd sanity check the design with SP but IMHO a gap between the avatar and the user name and align the user name in the middle of the avatar would work wonders:

Before: 
Screenshot 2024-06-19 at 13 34 46
After:
Screenshot 2024-06-19 at 13 43 17

For this the following css rule can be employed: 

`your-awesome-new-css-name {
	display: inline-flex;
	justify-content: center;
	align-items: center;
	column-gap: var(--lumo-space-s);
} `

Steffengreiner (changes requested):
Heya @KochTobi, 

Awesome work, the identicons look awesome! However i have one last very minor gripe about the css which i'm happy to discuss!

Steffengreiner (approved):
Well done 👍

PR #630, opened 2024-06-14
Reference API and contact in the footer

sven1103:


user-interface/src/main/java/life/qbic/datamanager/views/general/footer/FooterComponent.java
```diff
@@ -23,13 +25,21 @@ public class FooterComponent extends Footer {
...
+      @Value("${qbic.communication.api.url}") String apiUrl,
+      @Value("${qbic.communication.contact.email}") String contactEmail,
+      @Value("${qbic.communication.contact.subject}") String contactSubject) {
     setId("data-manager-footer");
     add(new Anchor(dataPrivacyAgreement.getHref(), "Data Privacy Agreement", AnchorTarget.BLANK),
         new Anchor(legalNoticeLink.getHref(), "Legal Notice", AnchorTarget.BLANK),
-        new Anchor(sourceCodeLink.getHref(), "Source", AnchorTarget.BLANK));
+        new Anchor(documentationUrl, "Documentation", AnchorTarget.BLANK),
+        new Anchor(apiUrl, "API", AnchorTarget.BLANK),
+        new Anchor(sourceCodeUrl, "Source", AnchorTarget.BLANK),
```
Steffengreiner, line 41: Please remove the quotation mark from the default example in the application.properties. 
Otherwise the source link will be broken (Since vaadin cannot recognize the protocol and therefore assumes a relative link instead of an absolute link) 
Screenshot 2024-06-14 at 13 10 41

user-interface/src/main/java/life/qbic/datamanager/views/DataManagerLayout.java
```diff
@@ -18,9 +19,9 @@
 public class DataManagerLayout extends AppLayout implements RouterLayout {
 
   private final Div contentArea;
-  private final FooterComponent footerComponent = new FooterComponent();
 
-  protected DataManagerLayout() {
+  protected DataManagerLayout(FooterComponent footerComponent) {
+    Objects.requireNonNull(footerComponent);
```
Steffengreiner, line 24: What's the advantage of wiring the footer component in each inheriting layout instead of creating it once in the outmost parent layout?

Steffengreiner (changes requested):
Well done! Once the default setting for the source url is adapted this can be merged 👍

Steffengreiner (approved):
Well done!

PR #631, opened 2024-06-18
Remove duplicate error message

KochTobi:
Removes error text duplication.


wow-such-code (approved):
long awaited fix, nice work!

PR #633, opened 2024-06-19
Add a prefill row for every replicate in an experimental group.

KochTobi:


sven1103 (approved):
Thanks @KochTobi for the quick hotfix,

PR #643, opened 2024-06-24
Extending project code to avoid collisions with legacy QBiC project codes

sven1103:
The legacy project codes are of length 5, so the data manager should not create codes that might be confused with legacy codes.

The most simple solution also discussed with project management is to extend the prefix from "Q" to "Q2".


wow-such-code (commented):
What is the rationale behind fixing the second symbol (to a "2")?
We could increase the number of possible projects and give the users more freedom by allowing them to customize the second symbol.

Edit: nevermind, clarified in personal discussion.

wow-such-code (approved):
tested (until data upload) and approved

PR #647, opened 2024-06-26
Addresses registration side effects observed on simultanous clients login

sven1103:


KochTobi:
@sven1103 and I tested it. Looks good for registration.

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationLayout.java
```diff
@@ -55,19 +72,76 @@ public class UserRegistrationLayout extends VerticalLayout {
...
+  private void handleResponse(ApplicationResponse response) {
+    response.ifSuccessOrElse(this::onUserRegistrationSucceeded, this::handleRegistrationFailure);
+  }
+
+  private void handleRegistrationFailure(ApplicationResponse response) {
+    UserRegistrationException exception = build(response);
+    handleRegistrationFailure(exception);
+  }
+
+  private UserRegistrationException build(ApplicationResponse applicationResponse) {
```
KochTobi, line 123:
```suggestion
  private UserRegistrationException convertToRegistrationException(ApplicationResponse applicationResponse) {
```
The method name `build` does not give any information in this context.

KochTobi (dismissed):
Great catch! Makes the code easier to understand by far!

This PR removes the tests for the user registration. I feel like we should still test it somehow but I do not have an idea how, that is why I will approve this PR.

KochTobi (approved):

PR #655, opened 2024-06-28
Add cancel confirmation dialog to most dialog winows

wow-such-code:
* disable project/experiment dialogs closing when clicking outside of the dialog
* add notification with confirmation button to all cancel operations when creating or editing projects and experiments
* this includes pressing "Esc" to cancel the dialogs


wow-such-code:
https://github.com/qbicsoftware/data-manager-app/assets/10232219/f0eef7a0-607c-4d2c-8c42-ecab0722e9f1

Shraddha0903:
Looks good, I have few points
1. Discard button should be in red
2. The secondary button should be next to primary button and not on the extreme left (all dialogs follow this)
3. Would be better if we don't name the secondary button "Cancel" as it confuses the user with the cancellation task at hand. "Keep it" or "Continue" can be used.
4. I feel we don't need the cancel confirmation for Offer uploads, as the damage done is reversible. The dialog window could just close upon clicking on cancel or escape and not ask for a confirmation.

Shraddha0903 (changes requested):
Looks good, I have a few points

1. Discard button should be in red
2. The secondary button should be next to primary button and not on the extreme left (all dialogs follow this)
3. Would be better if we don't name the secondary button "Cancel" as it confuses the user with the cancellation task at hand. "Keep it" or "Continue" can be used.
4. I feel we don't need the cancel confirmation for Offer uploads, as the damage done is reversible. The dialog window could just close upon clicking on cancel or escape and not ask for a confirmation.

user-interface/src/main/java/life/qbic/datamanager/views/CancelConfirmationNotificationDialog.java
```diff
@@ -0,0 +1,47 @@
...
+    Button redButton = new Button("Discard");
+    redButton.addClassName("danger");
+    setConfirmButton(redButton);
+    Button cancelButton = new Button("Continue");
+    setCancelButton(cancelButton);
+  }
+
+  public CancelConfirmationNotificationDialog withBodyText(String mainText) {
+    content.add(new Span(
+        mainText));
```
Steffengreiner, line 28: I'd recommend to call content.removeall() before adding a new span here to keep the behaviour consistent.

user-interface/src/main/java/life/qbic/datamanager/views/general/DialogWindow.java
```diff
@@ -29,6 +33,13 @@ protected DialogWindow() {
     confirmButton.addClickListener(this::onConfirmClicked);
   }
 
+  protected void initCancelShortcuts(Command onCreationCanceled) {
```
Steffengreiner, line 36: NIT: Since you're changing the cancel shortcut you're theoretically not initializing them 🤓 
```suggestion
  protected void specifyCancelShortcuts(Command onCreationCanceled) {
```

user-interface/src/main/java/life/qbic/datamanager/views/general/DialogWindow.java
```diff
@@ -29,6 +33,13 @@ protected DialogWindow() {
...
+      setCloseOnOutsideClick(false);
+      setCloseOnEsc(false);
+      Shortcuts.addShortcutListener((Component) this,
```
Steffengreiner, line 39: This casting seems redundant
```suggestion
      Shortcuts.addShortcutListener( this,
```

user-interface/frontend/themes/datamanager/components/dialog.css
```diff
@@ -42,6 +42,13 @@ vaadin-dialog-overlay::part(title) {
   margin: 0;
 }
 
+/*Vaadin currently does not allow to style the flex items within the footer.
+Since we want to remove the spacing between the cancel and confirm button we replace the flex setting with a grid layout*/
+.notification-dialog::part(footer) {
+  display: grid;
+  grid-template-columns: 0 15% 15%;
+}
```
Steffengreiner, line 50: Did you ask SP if this button layout is defined for all notification dialogs or only for all cancelconfirmationnotification dialogs. If it's the latter then i'd recommend to make this css more specific

Steffengreiner (changes requested):
Only NITS and minor questions, Awesome work 👍

Steffengreiner (approved):
Well done 👍

PR #702, opened 2024-07-09
Refactor SamplePage UI to current Prototype

Steffengreiner:
**What was changed**

1.) search and download logic is now handled by the SampleMain component similiar to measurementmain to adhere to prototype page structure
2.) Disclaimers are shown if no samples and experimental groups are provided, and the sample and batch components are hidden accordingly. 
3.) User can now filter sampleGrid by values set in organismId
4.) Column widths for both grids are updated if information changes to ensure that the content is always readable. 
5.) User can move column order in samplegrid according to preferences
6.) User can resize columns 
7.) Sample Id column is frozen and batch action column is frozen so it's always usable
8.) Sample label column name was changed to sample name

**Preview**
Screenshot 2024-07-09 at 17 46 47

https://github.com/qbicsoftware/data-manager-app/assets/29627977/3545eddc-ac11-4fae-ad13-754f02dbbdf4


Shraddha0903:
Hey @Steffengreiner, awesomeee work! 😍 
Everything checks off. Just that it would be even better if the column rearranging and resizing settings are saved for the user.
But I had fun rearranging 😉  We can also extend the grid functionalities to the measurements and RAW data view.

Shraddha0903 (commented):
Hey @Steffengreiner, awesomeee work! 😍
Everything checks off. Just that it would be even better if the column rearranging and resizing settings are saved for the user.
But I had fun rearranging 😉 We can also extend the grid functionalities to the measurements and RAW data view.

wow-such-code (changes requested):
Looks beautiful, nice work!
One thing I noticed, however, is that the Filter field does not reset when changing projects (and at the same time the filter is then not applied to the table).

wow-such-code (approved):

Issue #683, opened 2024-07-03
Experiment Information is editable even if samples are registered 


Steffengreiner:
**Describe the bug**
Currently experiment details such as species, specimen and analyte are editable even though samples have been registered. 
This means that a user can remove the species even though it has been associated with a sample, which leads to an inconsistent state. 

**To Reproduce**
In Design Experiment, after samples are registered, click on "edit experiment" and remove the selected species and add another, click on save and it still saves the new species.

**Expected behavior**
Species, Specimen and Analyte are not editable after samples have been registered.

KochTobi:
Is this truely a bug? As far as I understood it, the information on species, specimen and analyte are additional metadata curated by the user. Why do we need to check them?

Steffengreiner:
While this informaition is indeed experiment metadata, it is used to enable the user to associate a specific sample with the selected species, specimen and analyte. 
If the user deletes this information  samples cannot be registered anymore since no values can be selected. Additionally, even only allowing an edit leads to a mismatch between the species/specimen and analyte information associated with a sample and the experiment information.

KochTobi:
Who owns the information on which species, specimen, analyte are used in the experiment? What changes can occur during the lifetime of an experiment regarding the species, specimen and analyte?

PR #706, opened 2024-07-11
Restrict sample origin editing, if origin is used

wow-such-code:
Tackles #683 

https://github.com/qbicsoftware/data-manager-app/assets/10232219/a40fbfa5-9e07-4a99-864c-7aa870e18087


Steffengreiner (approved):
I like the solution! Well done 👍

PR #708, opened 2024-07-12
Logout user after beeing idle or max time of a session has beed reached

sven1103:


user-interface/src/main/java/life/qbic/datamanager/MyVaadinSessionInitListener.java
```diff
@@ -34,12 +38,23 @@ public void serviceInit(ServiceInitEvent event) {
...
           ui.getPage().retrieveExtendedClientDetails(clientDetailsReceiver);
           ui.getSession().setErrorHandler(errorEvent -> uiExceptionHandler.error(errorEvent, ui));
         });
+
+
+  }
+
+  @Override
+  public void sessionDestroy(SessionDestroyEvent event) {
+    log.debug("Session destroyed.");
```
KochTobi, line 56: can we log some kind of session id here? I imagine this information is not very usable if multiple sessions exist at the same time and we cannot differentiate.

user-interface/src/main/resources/application.properties
```diff
@@ -115,7 +115,7 @@ spring.servlet.multipart.max-request-size=16MB
 server.servlet.session.timeout=${SESSION_TIMEOUT:30m}
 vaadin.closeIdleSessions=${SESSION_CLOSEIDLESESSIONS:true}
 # Heartbeat value is specified as integer time in seconds
-vaadin.heartbeatInterval=${SESSION_HEARTBEATINTERVAL:300}
+vaadin.heartbeatInterval=${SESSION_HEARTBEATINTERVAL:60}
```
KochTobi, line 118: I think we can leave this at 5min

KochTobi (dismissed):

KochTobi (approved):

PR #750, opened 2024-07-30
provide cookie destruction method

wow-such-code:
this needs to be tested and maybe added as well to the manual user logout


KochTobi:
I could not figure out if this solves the issue. Cookies are populated with correct information as far as I could tell. I could not detect a difference, however I tested locally and not with the proxy setup.

Issue #743, opened 2024-07-29
When having a lot of experimental variables in an experiment I have to scroll to the right


KochTobi:
**Describe the bug**
When adding a lot of experimental variables I have to scroll to the right to see all of them. 

**To Reproduce**
1. Add a lot (>15) experimental variables
2. You have to scroll to the right

**Expected behavior**
The variables are in multiple rows.

**Screenshots**
![Screenshot 2024-07-29 at 16 02 29](https://github.com/user-attachments/assets/af223a00-e51a-48e3-acac-723a27e4ccb7)

**Smartphone (please complete the following information):**
 - Version data manager 1.2.3

KochTobi:
does not work for 
Firefox 129.0.2

sven1103:
@Steffengreiner this issue does not reference a PR, but has the status "Needs Review". Please link the PR that addresses the issue.

Steffengreiner:
Dear @sven1103. my apologies, These cards are a remnant of the discussion i had with @Shraddha0903. 
The original issue was already addressed in an older PR, however the cards resizing themselves had to be fixed.  
I linked the relevant PR

PR #754, opened 2024-08-01
Experimental variables and groups are shown as cards with proper scrolling and size

Steffengreiner:
**What was changed**
Addressed #743, #744 

and provides a zoom out of the application to make it more accessible for smaller screens

Finally updates the data privacy statement to include full name user name and orcid since we show it in the project access component now 🤔


Shraddha0903 (changes requested):
The cards look good, and I assume that the scroll direction has also changed now. 
Just one point - is the scroll applied to the entire Design experiment view? Because the species, specimen analyte are hidden when the user scrolls down. I feel the scroll should only be applied to the experiment variables/groups tab and not the entire view. So that the species, specimen and analyte would still be visible.

Steffengreiner:
Changes are out of date due to development branch reset. See #815 for new PR

PR #757, opened 2024-08-02
Provide feature request form

sven1103:


.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,42 @@
...
+      description: Also tell us, what is missing?
+      placeholder: Tell us what you would need!
+      value: "I am currently missing ... \nI would need ..."
+    validations:
+      required: true
+  - type: textarea
+    id: use-case
+    attributes:
+      label: Why do you need a solution?
+      description: Please explain your scenario (use case), how it affects your business.
```
Steffengreiner, line 27: NIT: I'd argue we don't work with businesses that much but more with individual users.
```suggestion
      description: Please explain your scenario (use case) and how it affects your experience
```

.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,42 @@
+name: Feature Request
```
Steffengreiner, line 1: NIT: I'd recommend to keep the naming consistent between the yml file and the name specified in the template
Renaming the `issue.yml` file to `feature_request.yml` would be more explicit

Steffengreiner: Additionally, you can also specify in what order the issue templates are shown by adding a numbering before the issue file name, which could make sense here as well (e.g. `01-bug-report.yml` and `02-feature-request.yml`) 🤔 
See [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#changing-the-order-of-templates)

Steffengreiner (approved):
I have some minor nitpicks, which could improve the yml but nothing major. Well done 👍

Steffengreiner (approved):
Well done 👍

Steffengreiner (approved):
Nice catch 👀

PR #761, opened 2024-08-06
Release 1.3.0

KochTobi:


sven1103:
Unless https://github.com/qbicsoftware/data-manager-app/issues/763 is not addressed, I am against creating the release. We covered this last week in the Sprint review, and agreed to focus on the release finish in favor of ongoing tasks.

PR #790, opened 2024-08-16
Provide a notification system

### LINKED ISSUE #682: Provide a notification based feedback system
**Is your feature request related to a problem? Please describe.**
Currently the users are not informed in a harmonized way. Notifications are shown sometimes on top and on the bottom left. This concerns the following notifications (Thanks SP for collecting):

**Success notifications (Toasts)**

Usage: Triggered after successful completion of each task

**Warning notifications (Dialogs)**

Usage: To warn the user of potential risks that may arise during an interaction. 

**Information notifications (Dialogs)**

Confirmation msg on dialog cancellation 

**Error Notifications**

Usage: To inform users of critical issues and guide them to resolution.

**Describe the solution you'd like**
Notifications are shown as "toast" notifications with a clear warning/error/success indication instead of stacking on top of each other

**Describe alternatives you've considered**
There is no alternative! 

**Additional context**
To see an example: This is already done after experimental group registration to guide the user to the sample registration.

KochTobi:
resolve #682 

Moved code to new branch to avoid dealing with git history


sven1103:
as I understand this PR, it is the notification system, which is not sufficient to resolve #682 . Correct?

sven1103:
@KochTobi in order for @Shraddha0903 to verify the UI/UX of this feature, please also provide a screen record / screen shot

sven1103:
Something is wrong with the implementation, when I add an experimental variable, the confirmation behaviour is the one of the cancel button.

I also think, that the notification shall be submitted only when the application server returns a successful operation response, and not when the confirm button is clicked.

https://github.com/user-attachments/assets/7e246520-2b03-4bcf-9b73-b2dc2fc48ea3

user-interface/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Creates a routing component with correct css classes and layout.
+   *
+   * @param text
+   * @param navigationTarget
+   * @param routeParameters
+   * @return
+   */
+  public Component createRoutingComponent(String text,
```
sven1103, line 238: I dont understand the purpose of this API. Why is it necessary to return the component? This gives way to nasty side effects (buttons behaviour can be easily changed). I suggest to have a factory method for a toast with a routing component instead.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /**
+   * Creates a matching toast content containing routing components with the correct css classes.
+   *
+   * @param content
+   * @param routingComponent
+   * @return
+   */
+  public static Component createRoutingContent(Component content, Component routingComponent) {
```
sven1103, line 221: I think this API is way out of scope for this class. Moreover, not a Toast is created, but a component.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Sets whether toasts are closed after navigation. By default, Toasts no not stay open after
+   * navigation.
+   *
+   * @param close whether to close the toast notification on navigation or not. true - closes the
+   *              toast after navigation.
+   * @return the modified toast
+   */
+  public Toast closeOnNavigation(boolean close) {
```
sven1103, line 69: Why not using the builder pattern here, to create an immutable Toast object? It can still be configured upon creation, but I dont understand the use case of having its configuration mutable. I have more concern about opening the door fot side effects in applications.

Can you elaborate some strong arguments for mutability, and if it is needed for ALL behaviour related properties? If not, i suggest to make them mutable-

sven1103 (changes requested):
Thank you @KochTobi for your work,

I might not understand all the reasoning that lead so some design decisions, but have some remarks and also some concerns that I provided. 

Lets chat about it to resolve it :)

user-interface/src/main/java/life/qbic/datamanager/views/notifications/WithCloseListener.java
```diff
@@ -0,0 +1,101 @@
...
+   * when the dialog is requested to close. Then you can decide whether to close or to keep opened
+   * the dialog. It means that dialog won't close unless you call the
+   * {@link CloseableWithoutListeners#closeIgnoringListeners()} method explicitly in the listener
+   * implementation.
+   *
+   * @param listener the listener to add
+   * @return the resulting
+   * {@link life.qbic.datamanager.views.notifications.WithCloseListener.ListenerRegistration}
+   */
+  ListenerRegistration addCloseListener(DialogCloseListener<C, T> listener);
```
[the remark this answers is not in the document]
KochTobi: No it is not. You can add the same listener multiple times and listen from everywhere.

KochTobi:
Thank you @sven1103 for your review I will think about the points you made here. I think the aspect of immutability is important but I am not entirely sure whether we want to do this here.
I will get back to you

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+  private Component content;
+  private Level level;
+  private boolean closeable;
+
+  enum Level {
+    SUCCESS,
+    INFO;
+  }
+
+  static class ToastBuilder {
```
sven1103, line 53: Obsolete inner class, you can use a static factory method for creation

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    addClassName("toast-notification");
+    add(closeButton);
+    setCloseable(DEFAULT_CLOSEABLE);
+    setPosition(DEFAULT_POSITION);
+    setType(DEFAULT_LEVEL);
+    closeOnNavigation(DEFAULT_CLOSE_ON_NAVIGATION);
+    setDuration(DEFAULT_OPEN_DURATION);
+  }
+
+  private static Button closeButton(Toast toast) {
```
sven1103, line 72: I have a hunch why you did it like this, however this method does some secret action, which is not clear from the signature: it already adds a click listener to the button to close the toast.

I suggest to make that behavioural change more explicit in an own call in the constructor.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    var button = new Button(LumoIcon.CROSS.create());
+    button.addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE);
+    button.addClickListener(it -> toast.close());
+    button.addClassName("close-button");
+    button.setVisible(false);
+    return button;
+  }
+
+  /**
+   * Sets whether toasts are closed after navigation. By default, Toasts no not stay open after
```
sven1103, line 82: typo: By default, Toasts DO not stay open

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+          toast.close();
+        }
+      };
+      UI ui = attachEvent.getUI();
+      Registration beforeLeaveListenerRegistration = ui.addBeforeLeaveListener(beforeLeaveListener);
+      closeOnNavigationListeners.add(beforeLeaveListenerRegistration);
+    };
+  }
+
+  public void setDuration(Duration duration) {
```
sven1103, line 114: no need to make it mutable, it can be passed as parameter during construction.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    super.setDuration((int) duration.toMillis());
+  }
+
+  /**
+   * The toast contains a success message
+   *
+   * @return
+   */
+  public Toast success() {
```
sven1103, line 123: i favor an expressive factory method call, like Toast.success(Duration d, Component c, ...)

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    setType(Level.SUCCESS);
+    return this;
+  }
+
+  /**
+   * The toast contains an informational message only
+   *
+   * @return
+   */
+  public Toast info() {
```
sven1103, line 133: Please encapsulate that from the client, there is no need to manipulate a toast during usage

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    setType(Level.INFO);
+    return this;
+  }
+
+  @Override
+  public void open() {
+    var cssClass = switch (level) {
```
sven1103, line 140: this should be set during construction and immutable

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    removeAll();
+    add(this.content, this.closeButton);
+  }
+
+  /**
+   * Can this toast be closed by the user?
+   *
+   * @return
+   */
+  public boolean isCloseable() {
```
sven1103, line 209: Why is this an important API?

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
```diff
@@ -0,0 +1,259 @@
...
+    return closeable;
+  }
+
+  /**
+   * Specifies whether the toast includes a button for the user to close the toast.
+   *
+   * @param closeable
+   * @return
+   */
+  public Toast setCloseable(boolean closeable) {
```
sven1103, line 219: What are the use cases for making that mutable? Should we not always offer the possibility for a user to close the toast?

sven1103 (dismissed):
I just went over the Toast.class and found some things that I see are critical.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
KochTobi: I think that the behaviour (closing a toast) is inherent to the close button. The close button would not be a close button without the closing behaviour. So this needs to be together imo. I renamed the method to make it more clear.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
KochTobi: The duration is mutable by default. This is given by the Vaadin class. So I adhere to this behaviour by introducing a simple method that takes a duration object instead of milliseconds (int).

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
KochTobi: Toast construction is now handeled exclusively in the factory.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
KochTobi: Wasn't clear at the time. I aggree

user-interface/src/main/java/life/qbic/datamanager/views/notifications/toasts/Toast.java
KochTobi:
```suggestion
   * Sets whether toasts are closed after navigation. By default, Toasts do not stay open after
```

user-interface/src/main/java/life/qbic/datamanager/views/general/ComponentFunctions.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.vaadin.flow.component.Component;
+
+public class ComponentFunctions {
+
+  private ComponentFunctions() {
+  }
+
+  /**
+   * Navigates up through the tree and checks whether the potential parent is an actual parent of
```
wow-such-code, line 13:
```suggestion
   * Navigates up through the tree and checks whether the potential parent is an actual ancestor of
```
Since this iterates, I would rather call the relationship "ancestor", not parent

user-interface/src/main/java/life/qbic/datamanager/views/notifications/CloseableWithoutListeners.java
```diff
@@ -0,0 +1,16 @@
+package life.qbic.datamanager.views.notifications;
+
+/**
+ * Can close without triggering any close listeners.
```
wow-such-code, line 4: what is the benefit of this?

user-interface/src/main/java/life/qbic/datamanager/views/general/QbicDialog.java
```diff
@@ -0,0 +1,152 @@
...
+
+  private final List<ListenerRegistration> dialogCloseListeners = new ArrayList<>();
+  private ListenerRegistration requireCloseConfirmationListener = null;
+  private boolean ignoresModificationForCloseConfirmation = SupportsCloseConfirmation.super.ignoresCloseCheckIfUnmodified();
+  private boolean modified = false;
+
+  public QbicDialog() {
+    setCloseOnOutsideClick(false);
+    setCloseOnEsc(false);
+    Shortcuts.addShortcutListener(this, this::close, Key.ESCAPE);
```
wow-such-code, line 32: this allows closing via ESC for all QbicDialogs, even those not requiring confirmation. Do we want that behaviour?

user-interface/src/main/resources/messages/error-messages.properties
```diff
@@ -1,11 +1,11 @@
 GENERAL=Apologies, an error occurred! -> Looks like something went wrong! We track these errors, but if the problem persists feel free to contact us at support@qbic.zendesk.com. In the meantime, try refreshing.
-ACCESS_DENIED=Insufficient rights. -> You don't have sufficient rights for this operation, please contact the project owner.
+ACCESS_DENIED=Insufficient rights. -> You don''t have sufficient rights for this operation, please contact the project owner.
```
wow-such-code, line 2: why two ' ? is this an escape in this context?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -77,8 +74,8 @@ public AddProjectDialog(ProjectInformationService projectInformationService,
       SpeciesLookupService speciesLookupService,
       ContactRepository contactRepository, TerminologyService terminologyService) {
     super();
-
-    initCancelShortcuts();
+    requireCloseConfirmation();
+    setIgnoreCloseCheckIfUnmodified(false);
```
wow-such-code, line 78: this method sounds like a good idea, but why is it set to false?

KochTobi: it is set to false as we did not implement the modification tracking. So as to not change the behaviour it is set to false right now. I will add a comment.

user-interface/src/main/java/life/qbic/datamanager/views/notifications/CloseableWithoutListeners.java
KochTobi: Sometimes (for example after confirmation / successful operation). As calling `close` requires confirmation, I needed another method that provides instant closing without any user confirmation.

user-interface/src/main/java/life/qbic/datamanager/views/general/QbicDialog.java
KochTobi: Due to the interface introduces, the close confirmation does not care whether you close by pressing ESC or calling `close`. By adding the shortcut explicitly, the `close` method is always called (triggering close listeners that show the close confirmation).

user-interface/src/main/java/life/qbic/datamanager/views/general/QbicDialog.java
```diff
@@ -0,0 +1,21 @@
...
+import com.vaadin.flow.component.Shortcuts;
+
+/**
+ * A {@link com.vaadin.flow.component.dialog.Dialog} with additional functionality.
+ * <p>
+ *
+ * @see com.vaadin.flow.component.dialog.Dialog
+ * @since 1.4.0
+ */
+public class QbicDialog extends com.vaadin.flow.component.dialog.Dialog {
```
wow-such-code, line 13: NIT: could just use "Dialog" as extension here? I don't see a naming conflict

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/BatchDetailsComponent.java
```diff
@@ -276,15 +277,21 @@ public static class DeleteBatchEvent extends ComponentEvent<BatchDetailsComponen
     @Serial
     private static final long serialVersionUID = -5424056755722207848L;
     private final BatchId batchId;
+    private final String batchLabel;
 
     public DeleteBatchEvent(BatchDetailsComponent source, BatchId batchId,
-        boolean fromClient) {
+        String batchLabel, boolean fromClient) {
```
wow-such-code, line 283: why did this change in this PR?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectDetailsComponent.java
```diff
@@ -226,8 +226,6 @@ private void openProjectInformationDialog() {
     project.ifPresentOrElse(proj -> {
           EditProjectInformationDialog editProjectInformationDialog = generateEditProjectInformationDialog(
               proj);
-          editProjectInformationDialog.addCancelListener(
```
wow-such-code, line 229: why could this be removed without replacement?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/components/ExperimentalVariablesDialog.java
```diff
@@ -132,15 +116,6 @@ public void addCancelEventListener(final ComponentEventListener<CancelEvent> lis
     addListener(CancelEvent.class, listener);
   }
 
-  /**
-   * Closes the dialog, all entered information is lost.
```
wow-such-code, line 136: since this was removed, does it mean the entered information is not reset now?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMetadataUploadDialog.java
```diff
@@ -333,17 +332,22 @@ private void showErrorNotification(String title, String description) {
...
+    NotificationDialog dialog = NotificationDialog.warningDialog()
+        .withTitle("Discard uploaded information?")
+        .withContent(new Span("Uploaded information has not yet been saved."));
+    dialog.setCancelable(true);
+    dialog.setCancelText("Continue Editing");
+    Button redButton = new Button("Discard upload");
+    redButton.addClassName("danger");
+    dialog.setConfirmButton(redButton);
+    dialog.addConfirmListener(event -> {
+      event.getSource().close();
```
wow-such-code, line 345: shouldn't this be taken over by CancelConfirmationDialogFactory ?

wow-such-code (changes requested):
some questions and suggestions

user-interface/src/main/java/life/qbic/datamanager/views/general/QbicDialog.java
KochTobi: You are right. I originally called the class Dialog as well thus the concrete class name with packages and all.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/BatchDetailsComponent.java
KochTobi: Because the notification now also can contain the batch label

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectDetailsComponent.java
KochTobi: Because the closing is handled by the dialog already. No special method is needed anymore as the close method resets the dialog.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMetadataUploadDialog.java
KochTobi: did it

wow-such-code (approved):
nice work!

Issue #738, opened 2024-07-29
Add information on how to enter experimental variables


KochTobi:
**Is your feature request related to a problem? Please describe.**
As a person working in the lab, when I want to enter experimental variables, I do not know how enter multiple levels.

**Describe the solution you'd like**
Describe somewhere that each row is one level or help me out here.

**Describe alternatives you've considered**
Provide multiple fields where I can enter experimental variables.
Provide an option for me to decide how to split my input.

**Additional context**
Adding levels with a list like `one, two, three` leads only to one level in the variable.

Shraddha0903:
Suggestion: 
We provide example placeholders for each term in the Add Experimental Variables dialog and also provide helper text to the Levels text field so along with the example it is clear to the user. 


![Image](https://github.com/user-attachments/assets/a7c3b432-5131-4974-99fd-fac32136023f)

PR #802, opened 2024-08-22
Add helper text for experimental variable input

KochTobi:
Solves #738 
image


Shraddha0903:
@KochTobi The helper text looks good although can we adjust the text to 3 lines only? 
I cannot see the placeholder text in this screenshot. We provide placeholder text for examples of the values and how they should be written. Font colour for placeholder text is lighter than body text (Secondary text)
Each time the user adds an extra variable in the dialog, it should show the same example as placeholder text in the text fields.

Shraddha0903 (changes requested):
@KochTobi The helper text looks good although can we adjust the text to 3 lines only? 
I cannot see the placeholder text in this screenshot. We provide placeholder text for examples of the values and how they should be written. Font colour for placeholder text is lighter than body text (Secondary text)
Each time the user adds an extra variable in the dialog, it should show the same example as placeholder text in the text fields.

KochTobi:
> @KochTobi The helper text looks good although can we adjust the text to 3 lines only?

It is quite some effort to do this and increases complexity. To do this the input field would need to be wider which in turn makes the dialog wider.

> I cannot see the placeholder text in this screenshot. We provide placeholder text for examples of the values and how they should be written. Font colour for placeholder text is lighter than body text (Secondary text) Each time the user adds an extra variable in the dialog, it should show the same example as placeholder text in the text fields.

I did not notice them in the draft. They are added now.

Shraddha0903 (changes requested):
@KochTobi Awesome!
It's just that it takes up a lot of space for a single property, was trying to reduce that.

Just one other thing the blue cross at the end of the variable row is not aligned with the text fields, seems to be a little down (maybe coz its blue? 😛) Can we align it with the fields? 

Screenshot 2024-08-22 at 15 08 32

KochTobi:
> @KochTobi Awesome! It's just that it takes up a lot of space for a single property, was trying to reduce that.
> 
> Just one other thing the blue cross at the end of the variable row is not aligned with the text fields, seems to be a little down (maybe coz its blue? 😛) Can we align it with the fields?

Not easy no :/ The image is perfectly aligned. There is some transparent background inside the icon but changing that would be a lot of effort / I don't know how to easily do that with shipped icons.

> 
> Screenshot 2024-08-22 at 15 08 32

KochTobi:
> > @KochTobi Awesome! It's just that it takes up a lot of space for a single property, was trying to reduce that.
> > Just one other thing the blue cross at the end of the variable row is not aligned with the text fields, seems to be a little down (maybe coz its blue? 😛) Can we align it with the fields?
> 
> Not easy no :/ The image is perfectly aligned. There is some transparent background inside the icon but changing that would be a lot of effort / I don't know how to easily do that with shipped icons.
> 
> > Screenshot 2024-08-22 at 15 08 32

Ah wait. I understand what you mean. Let me check

Shraddha0903 (commented):
@KochTobi I feel we can wait for @Steffengreiner to have a look if it's possible, i.e. to top align the icon with respect to the text box only. If it's not possible or takes a lot of effort then we can move to another solution (for e.g. displaying the option to delete the row on the bottom of each row)

Shraddha0903 (commented):
@KochTobi Nicee job! I think it looks good now. 👍

Issue #677, opened 2024-07-02
Support XLSX upload for measurement metadata registration / update


sven1103:
**Is your feature request related to a problem? Please describe.**
The export of XLSX to UTF-16 is just another step the user gets confronted with alerts and warnings about the information loss, etc. Also, it offers a lot of room for errors, since the users are overwhelmed with export file format options.

**Describe the solution you'd like**
Support the upload of the downloaded template or the downloaded metadata for updates.

PR #803, opened 2024-08-22
Support xlsx upload for measurement metadata registration

sven1103:
Solves #677


user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/TSVParser.java
```diff
@@ -0,0 +1,101 @@
...
+   * code through the usage of Java's {@link Optional}.
+   *
+   * @param array the array to access an element from
+   * @param index the index of the element in the array to access
+   * @return the array element at position of the index wrapped in {@link Optional}, or
+   * {@link Optional#empty}, if the index is out of bounds.
+   * @since 1.4.0
+   */
+  private static Optional<String> safeAccess(String[] array, Integer index) {
+    if (index >= array.length) {
```
KochTobi, line 59:
```suggestion
    if (index >= array.length || index < 0) {
```

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/TSVParser.java
```diff
@@ -0,0 +1,101 @@
...
+    try (BufferedReader reader = new BufferedReader(
+        new InputStreamReader(inputStream, StandardCharsets.UTF_16))) {
+      content = reader.lines().toList();
+    } catch (IOException e) {
+      throw new ParsingException("Cannot read from input stream", e);
+    }
+    if (content.isEmpty()) {
+      throw new ParsingException("No content provided!");
+    }
+    var keyMap = new HashMap<String, Integer>();
```
KochTobi, line 77: Maybe rename to columnToIndex map to make it clearer what this map maps?

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/TSVParser.java
```diff
@@ -0,0 +1,101 @@
...
+      if (headerToLowerCase) {
+        keyMap.put(header[i].toLowerCase(), i);
+      } else {
+        keyMap.put(header[i], i);
+      }
+    }
+
+    var values = content.subList(1, content.size());
+    var iterator = values.iterator();
+    List<List<String>> rows = new ArrayList<>();
```
KochTobi, line 90: Maybe add a record Row here. So we end up with a List<Row> which is more readable than List<List<...>>

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/TSVParser.java
```diff
@@ -0,0 +1,101 @@
...
+    while (iterator.hasNext()) {
```
KochTobi, line 91: Why do you use an iterator here instead of a simple loop over the List?

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/TSVParser.java
```diff
@@ -0,0 +1,101 @@
...
+      var row = iterator.next().split(VALUE_SEPARATOR);
+      String[] rowData = new String[header.length];
```
KochTobi, line 93: Here the row class would make things easier to understand

user-interface/src/main/java/life/qbic/datamanager/parser/ParsingResult.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * <ul>
+ *   <li>Value A1, Value B1, Value C1</li>
+ *   <li>Value A2, Value B2, Value C2</li>
+ *   <li>...</li>
+ * </ul>
+ *
+ * @since 1.4.0
+ */
+public record ParsingResult(Map<String, Integer> keys, List<List<String>> values) {
```
KochTobi, line 60: Maybe think of introducing a small Row class.  List<List<String>> could as well not be read as Row<Column<String>> but the other way around.

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/XLSXParser.java
```diff
@@ -0,0 +1,109 @@
...
+    return new XLSXParser(false);
+  }
+
+  public static XLSXParser createWithHeaderToLowerCase() {
+    return new XLSXParser(true);
+  }
+
+  private static String readCellAsString(Cell cell) {
+    return switch (cell.getCellType()) {
+      case _NONE, ERROR, BOOLEAN, FORMULA -> null;
```
KochTobi, line 55: Shouldn't Boolean lead to `true` and `false` rather than null?

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/XLSXParser.java
```diff
@@ -0,0 +1,109 @@
...
+    XSSFSheet metadataSheet = workbook.getSheetAt(0);
+    XSSFRow headerRow = Optional.ofNullable(metadataSheet.getRow(0))
+        .orElseThrow(() -> new ParsingException("No header row found"));
+
+    if (isNull(headerRow)) {
+      throw new MetadataParser.UnknownPropertiesException(
+          "No properties have been found: did you provide a header row?");
+    }
+    List<List<String>> rows = new ArrayList<>();
+    Map<String, Integer> columns = new HashMap<>();
```
KochTobi, line 81: This is the mapping of column name to column index? I think it is confusing to have a `rows` variable holding the actual data and a `columns` variable only mapping indices. I suggest renaming `columns` to something like `columnNameToIndex`

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  @Override
+  public List<MeasurementMetadata> convert(ParsingResult parsingResult, boolean ignoreMeasurementId)
+      throws UnknownMetadataTypeException {
+    Objects.requireNonNull(parsingResult);
+    var properties = parsingResult.keys().keySet();
+    if (looksLikeNgsMeasurement(properties, ignoreMeasurementId)) {
+      return convertNGSMeasurement(parsingResult);
+    } else if (looksLikeProteomicsMeasurement(properties, ignoreMeasurementId)) {
```
KochTobi, line 107: nice method name :+1:

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+              String.join(", ", properties)));
+    }
+  }
+
+  private List<MeasurementMetadata> convertProteomicsMeasurement(ParsingResult parsingResult) {
+    var result = new ArrayList<MeasurementMetadata>();
+    var keyIndices = parsingResult.keys();
+    for (List<String> valueEntry : parsingResult.values()) {
+      var pxpMetaDatum = new ProteomicsMeasurementMetadata(
+          safeListAccess(valueEntry, keyIndices.getOrDefault(MEASUREMENT_ID.propertyName(), -1),
```
KochTobi, line 121: maybe add a little javadoc to indicate that the value `-1` is present for missing columns?

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+          Arrays.stream(NGSMeasurementProperty.values())
+              .map(NGSMeasurementProperty::propertyName).collect(
+                  Collectors.toSet()));
+    }
+    var missingProperties = new ArrayList<>();
+    for (Entry<String, Integer> entry : hitMap.entrySet()) {
+      if (entry.getValue() == 0) {
+        missingProperties.add(entry.getKey());
+      }
+    }
```
KochTobi, line 221: I think this is a good solution to find out what was missing :+1:

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+    LABEL("label"),
+    COMMENT("comment");
+
+    private final String name;
+
+    ProteomicsMeasurementProperty(String value) {
+      this.name = value;
+    }
+
+    static String trimValue(String value) {
```
KochTobi, line 285: This method confuses me. Why did you introduce it? why not call `trim().toLowerCase()`
instead?
It does not only trim the value but also puts it to lower case.
If you want a method to sanitize the input consider renaming the method to `prepareValue`, `sanitizeValue`, ...

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+    static Optional<ProteomicsMeasurementProperty> fromString(String value) {
+      var trimmed = trimValue(value);
+      return Arrays.stream(ProteomicsMeasurementProperty.values())
+          .filter(property -> property.propertyName().equals(trimmed)).findFirst();
+    }
+
+    static boolean valueMatchesAnyProperty(String value) {
+      var trimmedValue = trimValue(value);
+      return Arrays.stream(MeasurementProperty.values()).map(MeasurementProperty::name)
+          .anyMatch(trimmedValue::equalsIgnoreCase);
```
KochTobi, line 298: why to lower case if you compare ignoring case?

user-interface/src/main/java/life/qbic/datamanager/parser/MetadataConverter.java
```diff
@@ -0,0 +1,352 @@
...
+     * Tries to convert an input property value to a known {@link NGSMeasurementProperty}.
+     * <p>
+     * Trailing whitespace will be ignored.
+     *
+     * @param value the presumed value to convert to a known {@link NGSMeasurementProperty}
+     * @return the matching property, or {@link Optional#empty()}.
+     * @since 1.4.0
+     */
+    static Optional<NGSMeasurementProperty> fromStringTrailingIgnored(String value) {
+      var trimmedValue = value.trim();
```
KochTobi, line 337: `String:trim` will remove all trailing AND all leading whitespace. Is this intended here? From the method name I would expect the leading whitespace to not be removed and only the trailing to be handled.

user-interface/src/main/java/life/qbic/datamanager/parser/ParsingResult.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public Stream<List<String>> rows() {
```
KochTobi, line 62: how about `public Stream<Row> rows()`

user-interface/src/main/java/life/qbic/datamanager/parser/ParsingResult.java
```diff
@@ -0,0 +1,78 @@
...
+    return values.stream();
+  }
+
+  public Iterator<List<String>> iterator() {
```
KochTobi, line 66: how about `public Iterator<Row> iterator()`

user-interface/src/main/java/life/qbic/datamanager/parser/ParsingResult.java
```diff
@@ -0,0 +1,78 @@
...
+    return values.iterator();
+  }
+
+  public List<String> getRow(int row) {
```
KochTobi, line 70: How about `public Row getRow(int rowIndex)`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMetadataUploadDialog.java
```diff
@@ -315,19 +159,50 @@ private void onUploadFailed(FailedEvent failedEvent) {
...
+      displayError(succeededEvent.getFileName(),
+          "Unsupported file type. Please make sure to upload a TSV or XLSX file.");
+      return;
+    }
+    List<MeasurementMetadata> result;
+    try {
+      result = MetadataConverter.measurementConverter()
+          .convert(parsingResult, mode.equals(MODE.ADD));
+    } catch (
+        UnknownMetadataTypeException e) { // we want to display this in the dialog, not via the notification system
```
KochTobi, line 198: The comment is helpful here. I like how this turned out.

KochTobi (changes requested):
The code looks good :+1: Great job. 
Some minor remarks I would like you to address

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementValidationExecutor.java
```diff
@@ -1,9 +1,11 @@
 package life.qbic.datamanager.views.projects.project.measurements;
 
 import java.util.concurrent.CompletableFuture;
+import life.qbic.projectmanagement.application.measurement.MeasurementMetadata;
 import life.qbic.projectmanagement.application.measurement.validation.MeasurementValidationService;
 import life.qbic.projectmanagement.application.measurement.validation.ValidationResult;
 import life.qbic.projectmanagement.domain.model.project.ProjectId;
+import org.apache.poi.ss.formula.functions.T;
```
KochTobi, line 8: This sound wrong...

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/XLSXParser.java
sven1103: Sorry, not sure what you mean :/

KochTobi: For `BOOLEAN` fields we could use `Boolean.parseBoolean(String input)`. Und `Boolean.toString(boolean value)`

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementValidationExecutor.java
KochTobi: I think this import has to go. You wanted to use generics here and not some sort of Excel function?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMetadataUploadDialog.java
```diff
@@ -828,12 +526,12 @@ public UploadItemsDisplay(Upload upload) {
       uploadSectionTitle.addClassName("section-title");
 
       var saveYourFileInfo = new InfoBox().setInfoText(
-              "Please save your excel file as UTF-16 Unicode Text (*.txt) before uploading.")
+              "Please save your Excel file as UTF-16 Unicode Text (*.txt) before uploading.")
```
KochTobi, line 529: I would argue that this is no longer the only way as you can upload xlsx as well now.

user-interface/src/main/java/life/qbic/datamanager/parser/xlsx/XLSXParser.java
```diff
@@ -0,0 +1,117 @@
...
+   * @since 1.4.0
+   */
+  private static String readCellAsString(Cell cell) {
+    if (cell == null) {
+      return "";
+    }
+    return switch (cell.getCellType()) {
+      case _NONE, ERROR, BOOLEAN, FORMULA, BLANK -> "";
+      case NUMERIC -> String.valueOf(cell.getNumericCellValue());
+      case STRING -> cell.getStringCellValue();
```
KochTobi, line 67:
```suggestion
      case _NONE, ERROR, FORMULA, BLANK -> "";
      case BOOLEAN -> Boolean.toString(cell.getBooleanCellValue());
      case NUMERIC -> String.valueOf(cell.getNumericCellValue());
      case STRING -> cell.getStringCellValue();
```

KochTobi (changes requested):
Very nice adjustments. I like the switch to the Row record. Increases readability a lot.
One minor thing when readin cell values and one question regarding measurement editing.

Regarding measurement editing:
I followed the steps detailled in the documentation. However when I

1. Download the measurement metadata
2. Clear rows at the end by selecting them and pressing backspace on my keyboard
3. Save the XLSX file
4. Upload the edited XLSX file

Then I get the error that at least one sample id is missing. This happens for **both XLSX and TXT** (UTF-16 tab separated)

Screenshot 2024-08-27 at 17 45 38
Screenshot 2024-08-27 at 17 37 17

sven1103:
> Regarding measurement editing: I followed the steps detailled in the documentation. However when I
> 
>     1. Download the measurement metadata
>     2. Clear rows at the end by selecting them and pressing backspace on my keyboard
>     3. Save the XLSX file
>     4. Upload the edited XLSX file
> 

I was able to reproduce it and included another method for the `Sanitizer` to be able to filter out empty rows for both TSVParser and XLSXParser.

KochTobi (approved):
Great Job!

PR #809, opened 2024-08-27
Fix broken ProjectAccessComponent after dependency upgrade

KochTobi:
Due to the way the grid constructor initializes the grid, the grid is now generated with columns for each bean property. As we add the columns ourselves, this leads to every column already being present (duplicate column key) and the ProjectAccessComponent.java not being instantiated. Using the no-args constructor of the Grid class solves this issue.


sven1103 (approved):
Thank you for testing and fixing it right away!

PR #814, opened 2024-08-29
User is able to select species specific search via Togglebutton

Steffengreiner:
**What was changed**
The default checkbox for species specific search within the ontologylookupcomponent is replaced with a vaadin checkbox styled as a togglebutton as shown below

https://github.com/user-attachments/assets/08494805-7b84-4557-93ba-3b457a82e37c


sven1103:
Thank you @Steffengreiner for your work on the toggle button. I added some suggested refactorings to enhance usability of your component with an expressive API. We might re-use this component, so it think it is worth the effort.

Have a look at it and write me what you think. 

I kindly ask you to provide some JDs for the class itself, what is its purpose or refer to a definition of a toggle button if available.

I will not submit a review, because I dont want to block it during my absence :)

Love the button!

KochTobi (approved):
LGTM

PR #815, opened 2024-08-30
Experimental variables and groups are shown as cards with proper scrolling and size 

Steffengreiner:
**What was changed**

Introduces the same changes as the #754 PR but taking into account the force-push to development branch.
Therefore, addressed https://github.com/qbicsoftware/data-manager-app/issues/743, https://github.com/qbicsoftware/data-manager-app/issues/744

and provides a zoom out of the application to make it more accessible for smaller screens


wow-such-code (approved):
this is a big improvement, so I will approve.
please don't forget to discuss scroll behaviour of single components with @Shraddha0903 as mentioned in #754

PR #716, opened 2024-07-18
Help user to choose a valid password in the reset password input

Steffengreiner:
**What was changed**
Introduce current concept of main components and remove vaadin based components from rest password and new password sending components. 
Additionally provide binders checking if the provided password and email is valid

PR #819, opened 2024-09-02
Help user to choose a valid password in the reset password input

Steffengreiner:
Transfers the changes made #716 to the dev branch state after the reset

**What was changed**
Introduce current concept of main components and remove vaadin based components from rest password and new password sending components.
Additionally provide binders checking if the provided password and email is valid


KochTobi (changes requested):
The error messages should be the same.
Screenshot 2024-09-04 at 09 25 05
Screenshot 2024-09-04 at 09 24 44

user-interface/src/main/java/life/qbic/datamanager/views/layouts/BoxLayout.java
```diff
@@ -1,182 +0,0 @@
...
-/**
- * <b>Box Layout</b>
- *
- * <p>A box with a shadow containing a title, description, a layout for fields, a layout for buttons
- * and a span to add links.
- * Furthermore, the description text can be toggled visible or invisible
- *
- * @since 1.0.0
- */
-public class BoxLayout extends VerticalLayout {
```
KochTobi, line 21: How is this change related to the issue? What happened here and why?

user-interface/src/main/java/life/qbic/datamanager/views/general/CardLayout.java
```diff
@@ -0,0 +1,23 @@
...
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.html.Div;
+
+/**
+ * Card Layout
+ * <p>
+ * Card styled Layout employed within all {@link Main} in the
+ * {@link life.qbic.datamanager.views.login.LoginLayout}
+ */
+public class CardLayout extends Div {
```
KochTobi, line 12: What is a Card Layout? where is it used? How can it look. I am confused about this.

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetEmailSentComponent.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Reset Email Sent Component
+ * <p>
+ * Card Stylized component similar to {@link com.vaadin.flow.component.login.LoginOverlay}
+ * component. Informing the user that a reset password email was sent and directing her to the login
+ * layout
+ */
+@SpringComponent
+@UIScope
+public class ResetEmailSentComponent extends CardLayout {
```
KochTobi, line 22: why not make this a page instead and rout to it from the reset page?

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetPasswordComponent.java
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Reset Password Component
+ * <p>
+ * Card Stylized component similar to {@link com.vaadin.flow.component.login.LoginOverlay}
+ * component, Providing the input fields necessary to enable a user to specify his email for which
+ * the password should be reset
+ */
+@SpringComponent
+@UIScope
+public class ResetPasswordComponent extends CardLayout {
```
KochTobi, line 35: According to the javadoc this component and NewPasswordSetComponent have the same purpose?

user-interface/src/main/java/life/qbic/datamanager/views/register/RegistrationOrcIdMain.java
```diff
@@ -99,13 +99,13 @@ private void handleRegistrationFailure(List<RuntimeException> exceptionList) {
     if (exceptionList.isEmpty()) {
       return;
     }
-    if (exceptionList.contains(UserExistsException.class)) {
+    if (exceptionList.stream().anyMatch(e -> e instanceof UserExistsException)) {
```
KochTobi, line 102: why these changes?

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationComponent.java
```diff
@@ -35,7 +35,7 @@
  */
 @SpringComponent
 @UIScope
-public class UserRegistrationComponent extends Div {
+public class UserRegistrationComponent extends CardLayout {
```
KochTobi, line 38: Why again the card layout?

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationMain.java
```diff
@@ -81,12 +81,12 @@ private void handleRegistrationFailure(List<RuntimeException> exceptionList) {
     if (exceptionList.isEmpty()) {
       return;
     }
-    if (exceptionList.contains(UserExistsException.class)) {
+    if (exceptionList.stream().anyMatch(e -> e instanceof UserExistsException)) {
```
KochTobi, line 84: why the changes?

KochTobi (dismissed):

user-interface/src/main/java/life/qbic/datamanager/views/general/CardLayout.java
Steffengreiner: I updated the JD for more clarity. In a nutshell we currently employ the same card-like style for all components during password reset, registration and login. To avoid having to copy paste the css over and over again, this DIV based component contains the style necessary to make it look like a hovering card containing the input fields. 

Screenshot 2024-09-05 at 17 14 41
Screenshot 2024-09-05 at 17 14 36

user-interface/src/main/java/life/qbic/datamanager/views/layouts/BoxLayout.java
Steffengreiner: The BoxLayout was a remnant from the time in which we relied on the Vaadin specified styling(Vertical Layout). This change aims to harmonize on how the login, registration and password related components can be harmonized in our style format. (CSS based styling)

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetEmailSentComponent.java
Steffengreiner: That's a good question. 
Currently there is not a real benefit to seperate this functionality into a different page, since the user will only be shown the reset email sent component if he triggered the process during the password reset main. Do you think there is a need to seperate this functionality into different routes?

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetPasswordComponent.java
Steffengreiner: Nice catch 👁️. I updated the JD to better reflect their purpose.
(Reset Password --> Inform user about new password set)

user-interface/src/main/java/life/qbic/datamanager/views/register/RegistrationOrcIdMain.java
Steffengreiner: In a nutshell the contains method does not work for instance based comparison in generics. Since we're only interested in the type of exception thrown, this had to be adapted.

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationMain.java
Steffengreiner: In a nutshell the contains method does not work for instance based comparison in generics. Since we're only interested in the type of exception thrown, this had to be adapted.

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationComponent.java
Steffengreiner: The Cardlayout is a DIV Container with a specific style. This style is the same for all components related to password reset, user registration and login. Therefore it's easier to maintain if we have one java class which defines the style and reuse it in the different main classes

user-interface/src/main/java/life/qbic/datamanager/views/general/CardLayout.java
KochTobi: > I updated the JD for more clarity. In a nutshell we currently employ the same card-like style for all components during password reset, registration and login. To avoid having to copy paste the css over and over again, this DIV based component contains the style necessary to make it look like a hovering card containing the input fields.

You do not need to copy paste anything but the css class name if I am not mistaken? I would vote in favor of composition over inheritance here. Components that extend card cannot extend anything else. If you want it to look like a card, why not only add the css class name. Shouldn't this be sufficient?

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetEmailSentComponent.java
KochTobi: I think that this information has nothing to do with the reset form. No strong feelings just feels like an independent informational page.

user-interface/src/main/java/life/qbic/datamanager/views/register/UserRegistrationMain.java
KochTobi: Nice catch. However please convert it to a for loop. 
```java
if (myList.isEmpty()) {
  return;
}
for (RuntimeExceptione : myList) {
  if (e instanceof UserExistsException userExistsException) {
   //... do stuff with userExistsException
  } else if (e instanceof OtherException otherException) {
   //... do stuff with otherException
  } else if (...) {
   //... do stuff with ... exception
  }
}
return;

```

Steffengreiner: That's a good point! Thanks for the suggestion <3

user-interface/src/main/java/life/qbic/datamanager/views/general/CardLayout.java
Steffengreiner: Good point! For me it feels more like a feeling thing, we want to reuse the same styling for all the components in this context, so i'd like to have it as a general java class. But since it provides no functionality as of yet, we can also go for YAGNI and copy paste the css class. I adapted the PR accordingly and removed the cardlayout 🤔

user-interface/src/main/java/life/qbic/datamanager/views/login/passwordreset/ResetEmailSentComponent.java
Steffengreiner: I'd like to keep it like this for now if that's ok. Down the line if the need arises, it's easily changed to a seperate page 👍

Shraddha0903:
> The error messages should be the same. In the registration case, the box does not show a helper and shows the helper "xyz characters are required..." when the input is not valid. The helper disappears with valid input. In the password reset form, the helper text is always shown and `password to short` is shown when not enough characters are provided. Is this the intended design @Shraddha0903 ? Screenshot 2024-09-04 at 09 25 05 Screenshot 2024-09-04 at 09 24 44

@KochTobi Hmm i see the issue here, the text is redundant in the second case. So instead of the helper text can we just keep the message "Minimum 12 characters required" as placeholder text for the password and then we can have the error validation message, "Please provide a password with at least 12 characters". Same for both - registration form and password reset.

wow-such-code:
> > The error messages should be the same. In the registration case, the box does not show a helper and shows the helper "xyz characters are required..." when the input is not valid. The helper disappears with valid input. In the password reset form, the helper text is always shown and `password to short` is shown when not enough characters are provided. Is this the intended design @Shraddha0903 ? Screenshot 2024-09-04 at 09 25 05 Screenshot 2024-09-04 at 09 24 44
> 
> @KochTobi Hmm i see the issue here, the text is redundant in the first case and the second case provides no information beforehand. So instead of the helper text can we just keep the message "Minimum 12 characters required" as placeholder text for the password and then we can have the error validation message, "Please provide a password with at least 12 characters"? Same for both - registration form and password reset.

It seems that this was not addressed, yet? Otherwise the PR is good to go imo.

wow-such-code (approved):

Issue #826, opened 2024-09-05
Make the footer always visible without having to scroll on smaller screens  


Steffengreiner:
### What are you missing?

Currently, the footer is not always immediately accessible within the application but hidden behind a scrollbar if the content within the application takes too much space. I attached some examples from my macbook air screen:

<img width="1440" alt="Screenshot 2024-09-05 at 15 37 41" src="https://github.com/user-attachments/assets/b8532e74-1371-473e-8ef8-b7ca52bb0704">
<img width="1440" alt="Screenshot 2024-09-05 at 15 37 23" src="https://github.com/user-attachments/assets/60f4a47e-cec3-46fe-9fa0-785ec116593b">
<img width="721" alt="Screenshot 2024-09-05 at 15 36 43" src="https://github.com/user-attachments/assets/379b36a4-b69e-4846-ac15-09ff63b9107e">
<img width="1440" alt="Screenshot 2024-09-05 at 15 36 18" src="https://github.com/user-attachments/assets/845b5f4c-0ce7-4610-b2bf-91b1543b41c3">

### Why do you need a solution?

Having to scroll to see the footer is non-intuitive and the content scrolling should also be independent of the footer. (It's sad if i scroll down just to see the footer) 

### Ideas

This is actually not trivial! Vaadin does not provide a native footer solution, which is why we currently rely on having the footer inside the content-area of the application. 
Sadly, since we employ the vaadin app-layout, we don't have a way to redesign the applayout structure entirely from scratch.(The navbar is contained within the applayout and there is no footer slot) 
This means to properly extract the footer, we need to redesign how the parent layout is built.
An additional benefit here would be that we can finally tackle overflow properly for all our components in a harmonized manner. 
 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #827, opened 2024-09-05
Experimental Variable and Group Cards sizing stays consistent independent of screen size

Steffengreiner:
**What was changed**

In the current build, the cards try to resize dependent on the screen width. This can lead to content overflow.

https://github.com/user-attachments/assets/751628e1-50b4-4a88-8f5e-eb83779fc01c

This PR aims to fix this by allowing the css grid to easier adjust the card position if the screen width changes:

https://github.com/user-attachments/assets/46a6858d-9bc2-4ae2-ba89-b98cf625c1f2

**Open Question**

### Why is the overflow not in the card collection itself?
For this i opened a seperate issue #826 , since it relates to how the data manager application handles component sizing. 
Currently we don't enforce a screen width and height for the content container. Therefore the components can grow in size, leading to an overflow scrollbar on the content container. To change this an entire rewrite of the application structure would be necessary.

### DIFF
--- a/user-interface/frontend/themes/datamanager/components/card.css
+++ b/user-interface/frontend/themes/datamanager/components/card.css
@@ -50,13 +50,11 @@
 
 .card-collection {
   display: flex;
-  align-content: space-evenly;
-  flex-flow: column;
+  flex-direction: column;
   gap: 1rem;
   padding-left: 1.5rem;
   padding-top: 1.5rem;
 }
-
 .card-collection .collection-title {
   font-weight: bold;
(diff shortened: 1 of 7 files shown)


sven1103:
The PR solves the original issue, hence I will approve it.

However I do not understand, why we address an issue of many variables, which imo is an artificial problem. As users already told us, most of them will have probably 1-3 variables. 

This will then look like this on a 4k screen:

image

Maybe @Shraddha0903 , this is something worth to rethink?

sven1103 (approved):

PR #828, opened 2024-09-05
Fix NGS measurement template dropdowns

### LINKED ISSUE #823: I'm unable to select the "sequencing read type" while registering or editing NGS measurements
### What happened?

While registering NGS measurements i wanted to select the "Sequencing Read Type" for my samples from the preselected options. Sadly this was not possible, since the dropdown menu with possible options only appears after row 6

**No Dropdown** 
Screenshot 2024-09-03 at 17 33 52

**With Dropdown In Row 7**
Screenshot 2024-09-03 at 17 33 59

Even worse i'm completely unable to see the possible selection options within the cells of the "Sequencing Read Type" column while editing the downloaded NGS measurements independent of the selected row. 

Screenshot 2024-09-03 at 17 36 02

This occured in Excel and Numbers so i suspect it's on the application side. 

### What browsers are you seeing the problem on?

Excel and Numbers

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

KochTobi:
Fixes #823


Steffengreiner (changes requested):
Heya @KochTobi ! Thanks for the quick fix. 
However the dropdown menu is not shown when i want to edit downloaded NGS measurements, which was also documented in the issue?

KochTobi:
> Heya @KochTobi ! Thanks for the quick fix. However the dropdown menu is not shown when i want to edit downloaded NGS measurements, which was also documented in the issue?

Good spotting. But as far as I know the dopdown was never shown in the edit case. Regarding the issue I suggest an additional PR. This PR only tackles the template and not some generated excel sheet.

Steffengreiner (approved):
Thanks for clarifying the edit case in the daily! Well done 👍

Issue #740, opened 2024-07-29
Link ORCiD to existing account


KochTobi:
**Is your feature request related to a problem? Please describe.**
I already have an account that I created with my email and password. I also have an ORCiD. How do I link that ORCiD in my account so that other can see it and find me on ORCiD as well?

**Describe the solution you'd like**
I need a way to add the orcid to my profile.

**Describe alternatives you've considered**
None. I do not have any suggestion on how to implement it. I only want to have the opportunity to link the account and login with my ORCiD.

sven1103:
Suggestion: trigger the account linking in the user profile page, and do not worry about existing auth sessions, they will timeout anyway. 

I suggest to do the auth flow that triggers a callback URL again, which at this moment is just for linking an existing account. You might need to pass the user id for the callback, so you know for which user entity the update should be done.

Shraddha0903:
Suggestion:

Step 1: Link ORCiD account
![Image](https://github.com/user-attachments/assets/37e5554a-4f6b-4bb4-84f4-d01f0bb806e5)

Step 2: Show ORCiD account in profile
![Image](https://github.com/user-attachments/assets/17f39aaf-9518-4da0-850e-e3571693a239)

PR #833, opened 2024-09-12
Show orcId if present for users in their profile or show Anchor to link orcId 

Steffengreiner:
**What was changed**

Provides the UI to show the Orcid Linkage if a user has an orcid account: 

Screenshot 2024-09-12 at 17 01 40

Alternatively provide a non-functional anchor for users without orcid account:
Screenshot 2024-09-12 at 17 02 41

Linkage will be handled in a seperate PR (see #740)


wow-such-code (approved):

KochTobi (changes requested):
I am sorry but I think we cannot add it as is. As we do not have the functionality to `Link ORCID Account` we should not show this part. If you adapt this PR to only show linked ORCID accounts rather and not link new ones, I remove the change request again.

sven1103:
I think we can still render the linkage, if it is present. If the PR introduces the non-functional link, then I would ask to remove it and introduce it with the feature that supports it.

KochTobi (approved):

PR #852, opened 2024-10-16
Address Changes requested by @Shraddha0903

KochTobi:
* introduce spacing below the section headers
* decrease batch name field width
* add spacing in the download metadata section

PR #853, opened 2024-10-16
Address change requests

KochTobi:
Replaces #852 due to branch rename


Shraddha0903 (changes requested):
Looks good now, just have 2 more points

Screenshot 2024-10-16 at 13 17 27

1. Can you increase the spacing between the Uploaded sample data and Uploaded file, it could be the same as the space between other sections?
2. Also, now the error message design looks good, however the messages are not very clear to me.
The error message can say - No analysis method provided for **2 samples** / Missing species value (CURIE) for **1 sample**

Steffengreiner (changes requested):
Heya @Shraddha0903 and @KochTobi! I checked out the PR and like the changes! However i found some minor inconsistencies: 

1.) Why is there a disabled cancel button during the sample registration process?
2.) There is no error message shown below the batch name field nor in the validation section, if the batch name field is returned to an empty value state. 
3.) If the batch name field is empty, the register button will not trigger the register process and no error message appears. Furthermore the batch name field is not even visible immediately since the dialog is too big for my laptop screen
4.) If the batch name field is filled and an invalid spreadsheet is uploaded, the validation message appears as expected. 
However if i press the register button again the validation message will specify that i should upload a new spreadsheet which is incorrect, since my invalid spreadsheet is still in the uploaded files section. (I can't even upload a new file if i don't delete the current one within the section).

https://github.com/user-attachments/assets/d1155b17-dbd5-4021-90d7-23fa4da6de88

KochTobi:
> Heya @Shraddha0903 and @KochTobi! I checked out the PR and like the changes! However i found some minor inconsistencies:
> 
> 1.) Why is there a disabled cancel button during the sample registration process? 

Where did you observe that? I cannot find it. The only time the button is disabled is while the registration takes place. No possibility to cancel that so no cancel button you can press.

> 2.) There is no error message shown below the batch name field nor in the validation section, if the batch name field is returned to an empty value state. 

I was not aware there should be an error message. However the field is marked as required and is invalidated when no batch name is shown. 

> 3.) If the batch name field is empty, the register button will not trigger the register process and no error message appears. Furthermore the batch name field is not even visible immediately since the dialog is too big for my laptop screen

Depends on the hardware and is in accordance to the design. 

> 4.) If the batch name field is filled and an invalid spreadsheet is uploaded, the validation message appears as expected. However if i press the register button again the validation message will specify that i should upload a new spreadsheet which is incorrect, since my invalid spreadsheet is still in the uploaded files section. (I can't even upload a new file if i don't delete the current one within the section).

This is a check on whether there are valid measurement data present. What is the expected behavior here?

Shraddha0903 (changes requested):
@KochTobi i was able to successfully register my batch without any errors, the batch component and grid displays also display the info. 🎉
I reviewed your changes for the dialog :
1. I saw that you have set the row-gap as lumo-space-l , can we make it lumo-space-m? also i think we should keep this same spacing for the other two row-gaps in the dialog.

Screenshot 2024-10-17 at 13 21 28

2. for the error messages 
a) "n sample(s)" should be in bold
b) Is it necessary to say CURI is missing? I feel the user would not immediately understand what we are talking about. Saying "Missing value for species/specimen/analyte for **3 samples**" could be more easy to understand.
c) There are fullstops and colons in different messages, can we remove them altogether or add fullstops at the end of each sentence?

KochTobi:
> @KochTobi i was able to successfully register my batch without any errors, the batch component and grid displays also display the info. 🎉 I reviewed your changes for the dialog :
> 
>     1. I saw that you have set the row-gap as lumo-space-l , can we make it lumo-space-m? also i think we should keep this same spacing for the other two row-gaps in the dialog.
> 

What other row gaps are you talking about?

>     2. for the error messages
>        a) "n sample(s)" should be in bold

Can do :+1:

>        b) Is it necessary to say CURI is missing? I feel the user would not immediately understand what we are talking about. Saying "Missing value for species/specimen/analyte for **3 samples**" could be more easy to understand.

They could have entered something like `Homo sapiens` but we would not be able to parse that as we need the `Homo sapient [NCBITaxon:9606]` instead.

>        c) There are fullstops and colons in different messages, can we remove them altogether or add fullstops at the end of each sentence?

Yes. I have to find them in the code first.

Shraddha0903:
> What other row gaps are you talking about?

these ones
Screenshot 2024-10-17 at 14 24 07
 they are currently set to 1 em

> They could have entered something like `Homo sapiens` but we would not be able to parse that as we need the `Homo sapient [NCBITaxon:9606]` instead.
Are they allowed to enter text for those columns? I saw that the user can only select the values from the dropdown and those already come with an ontology CURI right?
If I try to enter my own values there, I get an excel error to select from the dropdown. Is this what you meant?

KochTobi:
> > What other row gaps are you talking about?
> 
> these ones 

Fixed.

> 
> > They could have entered something like `Homo sapiens` but we would not be able to parse that as we need the `Homo sapient [NCBITaxon:9606]` instead.
> 
> Are they allowed to enter text for those columns? I saw that the user can only select the values from the dropdown and those already come with an ontology CURI right? If I try to enter my own values there, I get an excel error to select from the dropdown. Is this what you meant?

Yes we cannot rely on the excel validation. They can still copy paste whatever they want and even if that would not be the case they could manipulate it in other ways. We need the ID in the square brackets. I did however catch the case when nothing was entered and this is now shown as missing.

Steffengreiner (approved):
Technically it's solved beautifully. I'd suggest hammering out the Validation/Error intricacies with SP. (These can also be tracked and tackled in a future PR, which is why i'll approve it)

KochTobi:
> @Steffengreiner @KochTobi
> 
> > 1.) Why is there a disabled cancel button during the sample registration process?
> 
> Can we remove the Cancel button from this step?

It is the same as for the measurement upload. I don't know why it is there. I can remove it from both but then the finish button should also be removed from both.
> 
> > 2.) There is no error message shown below the batch name field nor in the validation section, if the batch name field is returned to an empty value state.
> 
> I think we can provide an inline error message for this - "Please provide a batch name" All mandatory fields in our dialogs have the same behaviour.

Done.

> 
> > 4.) If the batch name field is filled and an invalid spreadsheet is uploaded, the validation message appears as expected. However if i press the register button again the validation message will specify that i should upload a new spreadsheet which is incorrect, since my invalid spreadsheet is still in the uploaded files section. (I can't even upload a new file if i don't delete the current one within the section).
> 
> Can we delete the file automatically if it's invalid and allow the user to upload the new file? Reading the instructions to re-upload and then seeing the disabled upload button could frustrate the user.
> 
> Also, clicking the Register button again after getting an invalid msg should not display this message, it's misleading. I would expect in this case that the Register button becomes disabled until the data is approved.

I changed the message to only show if nothing is uploaded. I thought we do not disable buttons in dialogs but let them press the buttons with nothing happening rather. I imagine it quite confusing if the button is disabled and the batch name was not filled. Then there is no error displayed and no possibility to press the button. 

I will not address anything beyond that for now as I think those small fixes can be done later on when the feature is in use.

Shraddha0903:
> It is the same as for the measurement upload. I don't know why it is there. I can remove it from both but then the finish button should also be removed from both.

The cancel button in this step was to stop the registration process (in case the process takes a lot of time), but if that cannot be done, we can remove the button. The finish button there is to show the next possible action to the user (it would have gotten active on successful registration), but now I feel doesn't serve its purpose.
I would still favour to have an active Cancel button there to allow the user to get out of the dialog if it takes a long time and the user doesn't want to perform the action at that time (for e.g. copy dialog in windows). But we can discuss this later, esp if we see this happening. For now you can remove both buttons from sample and measurement dialogs.

> I changed the message to only show if nothing is uploaded. I thought we do not disable buttons in dialogs but let them press the buttons with nothing happening rather. I imagine it quite confusing if the button is disabled and the batch name was not filled. Then there is no error displayed and no possibility to press the button.

Hmm, yes but eventually the user would upload the valid file and then if they click the Register button, would be prompted with the batch name error. But anyway, there are a mix of opinions here about the disable button. For me it would just indicate to the user that the process can't move ahead, that's why disabled.
Since we don't lose anything by keeping it active, we can do so. I also observed that the content shakes if you keep clicking on Register, so that's also good for feedback 👍🏼

> I will not address anything beyond that for now as I think those small fixes can be done later on when the feature is in use.
Okay, but I then later would want to address those fixes - especially regarding the disabled upload button and the error validation message structure.

After this change, the PR is good to go! Thanks for the fixes! 🙂

Shraddha0903 (changes requested):
<It is the same as for the measurement upload. I don't know why it is there. I can remove it from both but then the finish button should also be removed from both.

The cancel button in this step was to stop the registration process (in case the process takes a lot of time), but if that cannot be done, we can remove the button. The finish button there is to show the next possible action to the user (it would have gotten active on successful registration), but now I feel doesn't serve its purpose.
I would still favour to have an active Cancel button there to allow the user to get out of the dialog if it takes a long time and the user doesn't want to perform the action at that time (for e.g. copy dialog in windows). But we can discuss this later, esp if we see this happening. For now you can remove both buttons from sample and measurement dialogs.

<I changed the message to only show if nothing is uploaded. I thought we do not disable buttons in dialogs but let them press the buttons with nothing happening rather. I imagine it quite confusing if the button is disabled and the batch name was not filled. Then there is no error displayed and no possibility to press the button.

Hmm, yes but eventually the user would upload the valid file and then if they click the Register button, would be prompted with the batch name error. But anyway, there are a mix of opinions here about the disable button. For me it would just indicate to the user that the process can't move ahead, that's why disabled.
Since we don't lose anything by keeping it active, we can do so. I also observed that the content shakes if you keep clicking on Register, so that's also good for feedback 👍🏼

<I will not address anything beyond that for now as I think those small fixes can be done later on when the feature is in use.

Okay, but I then later would want to address those fixes - especially regarding the disabled upload button and the error validation message structure.

After this change, the PR is good to go! Thanks for the fixes! 🙂

KochTobi:
> > It is the same as for the measurement upload. I don't know why it is there. I can remove it from both but then the finish button should also be removed from both.
> 
> The cancel button in this step was to stop the registration process (in case the process takes a lot of time), but if that cannot be done, we can remove the button. The finish button there is to show the next possible action to the user (it would have gotten active on successful registration), but now I feel doesn't serve its purpose. I would still favour to have an active Cancel button there to allow the user to get out of the dialog if it takes a long time and the user doesn't want to perform the action at that time (for e.g. copy dialog in windows). But we can discuss this later, esp if we see this happening. For now you can remove both buttons from sample and measurement dialogs.

I removed both buttons now.
> 
> > I changed the message to only show if nothing is uploaded. I thought we do not disable buttons in dialogs but let them press the buttons with nothing happening rather. I imagine it quite confusing if the button is disabled and the batch name was not filled. Then there is no error displayed and no possibility to press the button.
> 
> Hmm, yes but eventually the user would upload the valid file and then if they click the Register button, would be prompted with the batch name error. But anyway, there are a mix of opinions here about the disable button. For me it would just indicate to the user that the process can't move ahead, that's why disabled. Since we don't lose anything by keeping it active, we can do so. I also observed that the content shakes if you keep clicking on Register, so that's also good for feedback 👍🏼
> 

So you would enable/disable the button based on the uploaded data and not on whether the dialog content is valid? 

> > I will not address anything beyond that for now as I think those small fixes can be done later on when the feature is in use.
> 
> Okay, but I then later would want those issues to be addressed - especially regarding the disabled upload button and the error validation message structure.
> 

What do you mean with disabled upload button? As we only enable the users to upload one file at a time this button is disabled by vaadin. Why should it stay enabled?

What do you mean with error validation message structure? I added the `reason for XYZ samples` to the end of each validation error reason. Is there another issue I am not aware of?

> After this change, the PR is good to go! Thanks for the fixes! 🙂

Shraddha0903 (approved):
Great job, Tobi! 🙂

PR #854, opened 2024-10-17
Fix measurement registration

KochTobi:
Due to the abstraction and centralization of column names, measurement registration was no longer possible (for ngs).
This PR addresses the measurement registration and edit case and fixes the column enums. Measurement registration and edition are separated now and the columns are easier to read in the code. The hope is that this PR will help in the future to understand which columns are read only, mandatory and at which position the columns are in the edit and register case. 
Further the separation of the columns allows to introduce registration template creation in a future PR.


Steffengreiner (changes requested):
Heya @KochTobi. Thanks for the fix, i think it helps a lot concerning the readability.

However i found two issues while testing it out within the following project/experiment:
`http://localhost:8080/projects/2935224a-8190-4d9c-891e-b8747c6947d3/experiments/a2bb8010-f991-4631-b875-53c233b00c39/samples`

1.) The Sample Main Page is not openable on this PR due to a Nullpointer Exception:

The exception is as follows: 
`java.lang.NullPointerException: Cannot invoke "life.qbic.projectmanagement.domain.model.sample.AnalysisMethod.label()" because the return value of "life.qbic.projectmanagement.application.sample.SamplePreview.analysisMethod()" is null
	at life.qbic.datamanager.views.projects.project.samples.SampleDetailsComponent.lambda$createSampleGrid$c0df629f$4(SampleDetailsComponent.java:144)`

2.) I was unable to register new Measurements for both proteomics and NGS. 

`Missing properties for NGS measurement: [measurement id, organisation name, instrument name]`

`Missing properties for proteomics measurement: [measurement id, organisation name, ms device name]`

KochTobi:
@Steffengreiner I addressed your comments. :)

Steffengreiner (approved):
Well done 👍

Issue #642, opened 2024-06-20
Editing registered metadata to add confounding variables


oquendoM:
**Is your feature request related to a problem? Please describe.**
I would like to be able to add confounding variables to already registered samples.

**Describe the solution you'd like**
After registration of a batch, it would be nice to be able to add confounding variables (ex. age, sex, BMI, cell culture medium, etc.) when editing the metadata of the batch.

**Describe alternatives you've considered**
At the moment the information can only be added as a comment, but it would be better to have it as a separate column. The new columns could be added through the browser editing function or, if not possible, by downloading the metadata as txt and uploading it with the additional columns.

<hr>

Based on personal discussions with @oquendoM @Shraddha0903 and @FraBarSol we conclude some facts and properties for confounding variables in the application:

- can be categorical or continuous
- are not always available beforehand
- can be provided at any point of the project
- are annotations of samples
- can be associated based on the sample name and sample id
- can be incomplete

In the case of an incomplete annotation of samples, warn the user but not block the annotation process. It might be, that there are missing values and the user is aware of it.

###  The use case

_"As a researcher, I want to annotate samples with confounding variables and their value for samples in a batch."_

#### Requirements 
 - If at least one sample misses an annotation for a provided confounding variable, warn the user before saving any change
 - The user is given the sample id and sample name during annotation
 - The user can update existing confounding variables

sven1103:
[DM-1115]

sven1103:
Dear @oquendoM , 

thank you for your request! The feature is actually already scheduled in the current development sprint and the first draft hopefully soon ready to show.

sven1103:
Depends on https://github.com/qbicsoftware/data-manager-app/issues/680, since it would enable us to let the user provide additional columns for confounding variables.

KochTobi:
We decided to move the concept of sample batches into the confounding variables.

KochTobi:
Samples within one experiment share experimental and confounding variables even if there is no value for some samples.

KochTobi:
The confounding variables may later be used for analysis. By filtering data for samples with certain criteria in confounding variables, effects of these variables on the results are investigated.

PR #877, opened 2024-10-30
Add support for confounding variables

KochTobi:
Addresses #642


KochTobi:
As confounding variables are valid for all samples in an experiment (including samples where a value might not yet been known), I decided to go for including empty variables in the samples.

KochTobi:
For this to work, [sql/complete-schema.sql](https://github.com/qbicsoftware/data-manager-app/blob/5b0befc6f99b394c0dd848de4afe01382d684e80/sql/complete-schema.sql) has to be executed on the database

sven1103:
@KochTobi there are some conflicts with the target branch `development`

sven1103:
@KochTobi can you elaborate more on the changes in the database? I also encourage to look into database versioning solutions like Flyway. I have some concerns with our current database schema development strategy: hoping for the best and no way back.

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleMetadata.java
```diff
@@ -49,12 +54,25 @@ public static SampleMetadata createUpdate(SampleId sampleId,
...
       String comment,
+      Map<ConfoundingVariableInformation, String> confoundingVariables,
       String experimentId) {
     return new SampleMetadata(sampleId, sampleCode, sampleName, analysisToBePerformed,
         biologicalReplicate, experimentalGroupId, species, specimen, analyte, comment,
+        confoundingVariables,
         experimentId);
   }
 
+  public static SampleMetadata fromNewToExisting(SampleId id, SampleMetadata sampleMetadata) {
```
sven1103, line 65: I am unsure about the naming here: the method just creates a new sample metadata record with the given sample ID. 
I suggest a naming that describes what it does, instead of assuming context. 

`addSampleID(SampleId id, SampleMetadata metadata)`

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
```diff
@@ -42,32 +51,46 @@ public class SampleRegistrationServiceV2 {
...
   @PreAuthorize("hasPermission(#projectId, 'life.qbic.projectmanagement.domain.model.project.Project', 'WRITE')")
   @Async
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
   public CompletableFuture<Void> registerSamples(Collection<SampleMetadata> sampleMetadata,
-      ProjectId projectId, String batchLabel, boolean batchIsPilot, ExperimentId experimentId)
+      ProjectId projectId, String batchLabel, boolean batchIsPilot, ExperimentReference experiment)
       throws RegistrationException {
     var result = batchRegistrationService.registerBatch(batchLabel, batchIsPilot, projectId,
-        experimentId);
+        ExperimentId.parse(experiment.id()));
```
sven1103, line 76: I think we overdid it here big time with the Identifier... Imo there is no need to parse, because it does not have any semantics or schema. It is just a string value generated from the UUID generator. You either find an object with the ID in the database, or you dont. 
Happy to see some simplification.

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
```diff
@@ -42,32 +51,46 @@ public class SampleRegistrationServiceV2 {
...
     this.sampleRepository = Objects.requireNonNull(sampleRepository);
     this.batchRepository = Objects.requireNonNull(batchRepository);
     this.sampleCodeService = Objects.requireNonNull(sampleCodeService);
     this.deletionService = Objects.requireNonNull(deletionService);
+    this.confoundingVariableService = confoundingVariableService;
   }
 
   @PreAuthorize("hasPermission(#projectId, 'life.qbic.projectmanagement.domain.model.project.Project', 'WRITE')")
   @Async
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
```
sven1103, line 71: What does this annotation do? Why is it needed?

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
```diff
@@ -42,32 +51,46 @@ public class SampleRegistrationServiceV2 {
...
     if (result.isError()) {
       throw new RegistrationException("Batch registration failed");
     }
     var batchId = result.getValue();
     try {
-      var sampleIds = registerSamples(sampleMetadata, batchId, projectId);
+      var registeredSamples = registerSamples(sampleMetadata, batchId, projectId);
+      for (Entry<Sample, SampleMetadata> registeredSample : registeredSamples.entrySet()) {
+        Map<VariableReference, String> levels = new HashMap<>();
+        registeredSample.getValue().confoundingVariables().forEach((k, v) -> levels.put(k.id(), v));
```
sven1103, line 85: suggestion: you can use `Collectors.toMap()` here, makes it more concise.

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
```diff
@@ -101,14 +126,25 @@ public CompletableFuture<Void> updateSamples(
     var sampleBySampleCode = samples.stream()
         .collect(Collectors.toMap(sample -> sample.sampleCode().code(), Function.identity()));
     var updatedSamples = updateSamples(sampleBySampleCode, sampleMetadata);
-    sampleRepository.updateAll(projectId, updatedSamples);
+    sampleRepository.updateAll(projectId, updatedSamples.keySet());
+
+    for (Entry<Sample, SampleMetadata> updatedSample : updatedSamples.entrySet()) {
+      Map<VariableReference, String> levels = new HashMap<>();
+
+      updatedSample.getValue().confoundingVariables().forEach((k, v) -> levels.put(k.id(), v));
```
sven1103, line 134: Again ``Collectors.toMap()``

sven1103 (changes requested):
Some minor change requests, but I have some concerns with how we deal with database changes.

KochTobi:
> @KochTobi there are some conflicts with the target branch `development`

Yes I resolved it.

KochTobi:
> @KochTobi can you elaborate more on the changes in the database? I also encourage to look into database versioning solutions like Flyway. I have some concerns with our current database schema development strategy: hoping for the best and no way back.

Hi Sven. I simply merged all the sql files to set up the database into one and added the tables for confounding variables. In this case the database schema is simply extended by the tables of confounding variables (samples, experiment and so forth are not touched).
I will happily look into Flyway or other database migration solutions. In the case of this PR I only recorded the current database scheme we have (plus the tables I added).

Even thou the topic of database migration and future proofing is of great importance, I would like to separate the investigation on database migration and future proofing from this PR.

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
KochTobi: I agree!

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
KochTobi: The transactional annotation is needed to indicate a transaction taking place. It starts a transaction in the database and commits changes when exiting the method. The propagation property tells spring that a new transaction is to be created for this operation. The default behaviour is that an existing active transaction is used and if there is none that one is created. I do not think we need to explicity create a new transaction here so I removed the propagation type.

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
KochTobi: To use `Collectors#toMap`, I need to create a stream first. So it does not make it more readable I feel. However I have no strong opinion on the matter. So if you think the lower example is better I will change it.
```
confoundingVariables().foreach((k, v) -> levels.put(k.id(), v));
```
```
registeredSample.getValue().confoundingVariables()
            .entrySet()
            .stream()
            .collect(Collectors.toMap(
                entry -> entry.getKey().id(),
                Entry::getValue));
```

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
KochTobi: see https://github.com/qbicsoftware/data-manager-app/pull/877#discussion_r1930180547

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
sven1103: Yes, just nit, but to be fair it is:

```java
Map<VariableReference, String> levels = new HashMap<>();
registeredSample.getValue().confoundingVariables().forEach((k, v) -> levels.put(k.id(), v));
```

vs 

```java
var levels = registeredSample.getValue().confoundingVariables()
                                .entrySet().stream()
                                .collect(Collectors.toMap(entry -> entry.getKey().id(), Entry::getValue));

The last one is side-effect free, vs the first is depending on an external map initialization :P

sven1103 (approved):
Great work @KochTobi, i am looking forward to the users feedback!

project-management/src/main/java/life/qbic/projectmanagement/application/sample/SampleRegistrationServiceV2.java
KochTobi: There are different scenarios:

1. no `@Transactional` annotation
2. a `@Transactional` annotation without `Propagation.REQUIRES_NEW`
3. a `@Transactional` annotation the default propagation

in 1. there will be an exception if the method is called from outside a transactional context. In our case no transactional context exists so the exception will be thrown. Without any transaction, no rollback is possible

in 2. at the start of the method a new transaction will be created, suspending an existing transaction if any already exists in this scope. The changes are made within the new transaction which is committed when the method exits (or rolled back in case of exceptions). After the newly created transaction finishes (commit or rollback) the previously suspended transaction continues. 

in 3. when called from a transactional scope (already existing transaction) the operations within the method are appended/added to the existing transaction. If the method throws an exception the complete transaction (including the changes made before this method was called) are rolled back. If no transaction existed when the method is called, a new transaction is created.

The difference between 2. and 3. is that in 2. a transaction is suspended. Rollbacks do not affect a scope outside of the method in 2. Suspending transactions bears a risk of deadlocking the system. If the amount of maximal transactions is limited and the creation of the transaction fails, there might be a scenario where the original transaction is kept in a suspended state further limiting the amount of available transactions. 
We also want to rollback an existing transaction in case of exceptions so only rolling back changes made in this method is not what we want I think.
Thus the propagation type of `REQUIRES_NEW` was wrong here.

KochTobi:
> > @KochTobi can you elaborate more on the changes in the database? I also encourage to look into database versioning solutions like Flyway. I have some concerns with our current database schema development strategy: hoping for the best and no way back.
> 
> Hi Sven. I simply merged all the sql files to set up the database into one and added the tables for confounding variables. In this case the database schema is simply extended by the tables of confounding variables (samples, experiment and so forth are not touched). I will happily look into Flyway or other database migration solutions. In the case of this PR I only recorded the current database scheme we have (plus the tables I added).
> 
> Even thou the topic of database migration and future proofing is of great importance, I would like to separate the investigation on database migration and future proofing from this PR.

https://github.com/qbicsoftware/data-manager-app/issues/1000 <- to track the need

Issue #894, opened 2024-11-13
Experiment Creation is not possible on the development branch


Steffengreiner:
### What happened?

I wanted to create an experiment within the development branch but got an error message instead of my cool new experiment: 

I think this broke during the overhaul of the notification system. 

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Relevant log output

```shell
Missing level info for experiment.created.success.
java.lang.RuntimeException: Missing level info for experiment.created.success.
	at life.qbic.datamanager.views.notifications.MessageSourceNotificationFactory.parseLevel(MessageSourceNotificationFactory.java:156)
	at life.qbic.datamanager.views.notifications.MessageSourceNotificationFactory.toast(MessageSourceNotificationFactory.java:70)
	at life.qbic.datamanager.views.projects.project.info.ProjectInformationMain.displayExperimentCreationSuccess(ProjectInformationMain.java:343)
	at life.qbic.datamanager.views.projects.project.info.ProjectInformationMain.onExperimentAddEvent(ProjectInformationMain.java:314)
	at com.vaadin.flow.component.ComponentEventBus.fireEventForListener(ComponentEventBus.java:239)
	at com.vaadin.flow.component.ComponentEventBus.fireEvent(ComponentEventBus.java:228)
	at com.vaadin.flow.component.Component.fireEvent(Component.java:411)
	at life.qbic.datamanager.views.projects.project.experiments.experiment.create.AddExperimentDialog.onConfirmClicked(AddExperimentDialog.java:128)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #897, opened 2024-11-13
Fix message level not found

KochTobi:
Solves #894 by adding the missing property.
Further logs rather than fail in the users face.
Needs to be refined with java enums for the future.


Steffengreiner (approved):
Well done 👍

Issue #904, opened 2024-11-14
Improve Visualisation for the Objective property in Project Summary


Shraddha0903:
### What are you missing?

The current visualisation of the Objective field is difficult to read, occupies excessive space, and creates a cluttered interface. A responsive layout is needed to enhance readability without taking up too much space.

### Why do you need a solution?

A responsive layout is needed to enhance readability without taking up too much space.

### Ideas

Suggestion:
Use the "Details" component to expand/collapse the Objective field. This will save space, improve readability, and provide a cleaner look.

Acceptance Criteria:

- Use "Details" component in filled variant, with background color #F9F9F9.
- Adjust height to fit full content.
- Set width to 60em.
- Increase line height to 1.7rem.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Shraddha0903:
Mockups:

![Image](https://github.com/user-attachments/assets/dfff6380-2d12-4049-9371-29e5f7862fe9)
![Image](https://github.com/user-attachments/assets/1cee3d5d-d92e-47b2-83cc-fcd544d6242e)

PR #905, opened 2024-11-15
Decrease measurement type badge gap

sven1103:
Fixes the space between measurement type badges in the summary and adds scroll-bars as overflow action for the objective field, until https://github.com/qbicsoftware/data-manager-app/issues/904 is ready.


user-interface/src/main/java/life/qbic/datamanager/templates/measurement/ProteomicsMeasurementRegisterTemplate.java
```diff
@@ -148,6 +148,8 @@ public byte[] getContent() {
       }
 
       setAutoWidth(sheet);
+      setAutoWidth(hiddenSheet);
```
KochTobi, line 151: why? It is hidden and not modifiable?

KochTobi:
```suggestion

```

KochTobi (changes requested):
Looks good!
Please remove the modification of the ProteomicsMeasurementTemplate from this PR then I approve :)

KochTobi (approved):

PR #927, opened 2024-11-26
Filter for distinct experiment information

sven1103:
During the aggregation of the project summary information, this contribution now also filters for distinct ontology terms to avoid redundant information.


KochTobi (approved):
LGTM

PR #931, opened 2024-11-27
Improve software quality by addressing some reliability issues

sven1103:
Improves the reliability and maintenance of the software by addressing various issues.


Steffengreiner (changes requested):
Heya @sven1103, 
Nicely done! I found 2 issues, however i'm unsure if they are related to this branch specifically. 

1.) Sonarcloud is not covering any of the new code:
https://github.com/qbicsoftware/data-manager-app/pull/931/checks?check_run_id=33709119877 

2.) During proteomic measurement registration and editing, the values input in the comment column are replaced with an empty string after registration. 
The reason for this seems to within one of the `convertProteomicsMeasurement` methods in the `MeasurementPresenter `class, which replaces the `measurement.comment()` value with "". 

I'll approve this PR once the sonarcloud report is working properly and i'm able to see if the code rating has improved.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/measurement/ProteomicsSpecificMeasurementMetadata.java
```diff
@@ -80,4 +80,8 @@ public String fractionName() {
     return fractionName;
   }
 
+  public String comment() {
+    return comment;
+  }
+
```
Steffengreiner, line 86: Why was the comment included in the `ProteomicsSpecificMeasurementMetadata`? 
Shouldn't it be in the `ProteomicsMeasurment` class? 

If not, the comments are currently not shown in the measurementgrid itself, since they are collected from the `ProteomicsMeasurement` class, which always returns an empty optional. 🤦

Issue #930, opened 2024-11-27
When creating an experiment, the specimen list shows empty entries


KochTobi:
### What happened?

During specimen selection multiple entries do not show a label. When selecting an entry without a corresponding label, multiple entries are selected. This is probably due to multiple `equal` entries contained in the combobox items.

The same issue exists for analytes.
https://github.com/user-attachments/assets/20e159fd-2e67-4875-b42e-241f5856307e

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #932, opened 2024-11-27
Fix redundant ontology terms that are actually the same

sven1103:
Due to the improvement of the OntologyTerm comparison by considering equal terms when they have the same IRI, the combobox showed empty rows for a term that was referenced in more than two ontologies.

This contribution filters them out.

Also, an unused ontology term list is removed. The NCBI Taxon filtering happens in the SpeciesSearchService and should also not be knowledgable by the client.


sven1103:
Fixes #930

PR #934, opened 2024-11-29
Introduce stepper dialog 

sven1103:
A stepper dialog (also known as wizard), contains at least two steps that are displayed consecutively to the user and ask them for input. They are usually applied when the user input is more complex, is conditionally chained, or can be separated contextually into logical steps.

This contribution introduces a stepper dialog component, that is easy to build and extend, and simplifies the stepper dialog creation.


user-interface/src/main/java/life/qbic/datamanager/views/general/dialog/stepper/StepperDialog.java
```diff
@@ -0,0 +1,117 @@
...
+  private final int numberOfSteps;
+  private int currentStep;
+
+  private StepperDialog(AppDialog dialog, List<Step> steps) {
+    this.dialog = Objects.requireNonNull(dialog);
+    this.steps = new ArrayList<>(Objects.requireNonNull(steps));
+    this.navigationListeners = new ArrayList<>();
+    if (steps.isEmpty()) {
+      throw new IllegalArgumentException("Steps cannot be empty");
+    }
```
Steffengreiner, line 32: Does it make sense to allow a list of steps with only one steps here? 🤔

Steffengreiner (dismissed):
Easy to understand, Works and looks nice! Well done 👍 (I have some minor questions as always 🧠 though)

Steffengreiner (approved):
Well done 👍

Issue #951, opened 2024-12-12
Remove hunting yellow box in Excel templates


sven1103:
### What are you missing?

Though with good intentions, the yellow info text box for every cell in the Excel template for sample registration are actually preventing to interact with some cells they hide, especially if users are filling up the sheet with a lot of information.

Also, the box gets really annoying quickly, since it appears everywhere you click in the spreadsheet

<img width="370" alt="image" src="https://github.com/user-attachments/assets/5d42ebe0-bc4c-4d4f-8942-441f9d8cbfd2" />

### Why do you need a solution?

We want to keep the mood of the users up.

### Ideas

Show it only in the header and maybe in the first row of values to add.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #960, opened 2024-12-19
Provide helper box only in the header and the first row of each template

Steffengreiner:
**What was changed**
Addresses #951 by only showing the helperbox in the header row and first row of each template.


KochTobi (approved):
👍 LGTM

PR #967, opened 2025-01-10
Add third party licenses

sven1103:
Adds the current transitive dependency licenses explicitly in `THIRD-PARTY.txt` and will include an up-to-date list in future builds.


Steffengreiner (approved):
Well deone

PR #974, opened 2025-01-14
Introduce toast factory and apply it on sample batch interaction

sven1103:
Introduces a toast factory to create toasts


sven1103:
@Steffengreiner I found the problem. Don't yet understand it though. The verified sample metadata referenced in the broadcastet event get somehow emptied, after i close the dialog via `event.getSource().close`. I made an actual copy of the sample metadata now from the consumed event before the processing starts. Did the same for registration, i dont want to get some nasty registration effects there as well.

Thank you for taking the time and 🦅 eye!

Steffengreiner (approved):
Well done 👍

Steffengreiner (dismissed):
Thanks for adding the JD 👍

Issue #975, opened 2025-01-14
Sample registration template messed up


sven1103:
### What happened?

We aim to assist the user by providing drop down lists of controlled vocabularies or information the user has already entered (e.g. condition).

However the sample template for sample registration shows no such drop down in the first row. Instead, the yellow info box is shown again, which must not except in the header.

<img width="512" alt="image" src="https://github.com/user-attachments/assets/443b92b3-5f35-4ab3-a792-e2b4b03b9f49" />

### What browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sven1103:
It is also true for species, specimen and analyte.

PR #976, opened 2025-01-14
Show helper box only in the header cells of all sample and measurement excel templates

Steffengreiner:
**What was changed**

Addresses #975, by restraining the helper box only to the header of the templates. This is done so users can easily drag and drop the values within the cells. 

**Open Questions**

1.) Should there be a highlight that there is a helper box hidden within the helper cell?
2.) There are header cells with links, which direct the user to their respective homepage (Organization to RoR Homepage, Instrument to Ontology). Is this an issue if the user is solely interested in the helper box content?


sven1103:
> **What was changed**
> 
> Addresses #975, by restraining the helper box only to the header of the templates. This is done so users can easily drag and drop the values within the cells.
> 
> **Open Questions**
> 
> 1.) Should there be a highlight that there is a helper box hidden within the helper cell? 2.) There are header cells with links, which direct the user to their respective homepage (Organization to RoR Homepage, Instrument to Ontology). Is this an issue if the user is solely interested in the helper box content?

@Steffengreiner thank you for the fix. I suggest to leave improvements out of the hotfix context, since it affects user to fill out the metadata properly atm.
I ping @Shraddha0903, she might give you an answer about the UX.

sven1103:
Please checkout from main and apply a hotfix, so we can deploy it.

Steffengreiner:
Closed since it's intended as a hotfix and should be applied to main directly. 
New PR can be found in #978

PR #986, opened 2025-01-16
Integrate the public Orcid API to simplify person lookup for project-affiliated contacts

sven1103:
Uses the public Orcid API to look up persons that the user might want to add as contact in a project.


Steffengreiner:
Just to give a quick update: This PR now works as intended, however the UI is basic and still needs to be confirmed with @Shraddha0903. 

Additionally the PR was tested with orcid Sandbox Instance so this needs to be adapted to the real orcid repository.
Finally make sure to set the environment variables to the sandbox instance as required:

`
ORCID_API_CLIENT_ID=Your_Awesome_Application_Client_Registered_With_Sandbox_Orcid;
ORCID_API_CLIENT_SECRET=Your_Awesome_Application_Secret_Registered_With_Sandbox;
ORCID_AUTHORIZATION_URL=https://sandbox.orcid.org/oauth/authorize; 
ORCID_JWK_SET_URI=https://sandbox.orcid.org/oauth/jwks;
ORCID_TOKEN_URI=https://sandbox.orcid.org/oauth/token;
ORCID_USERINFO_URI=https://sandbox.orcid.org/oauth/userinfo
`

Open Questions include:

1.) I had to extend the project information in the database to contain the oidc information for all 3 contacts (principal investigator, project manager, person responsible). I'd rather refactor this into a seperate table, maybe in a new PR? 

2.) Understanding and extending the BoundContactField is actually not trivial and we get a lot of complicated edge cases due to the Vaadin implementations. This could be refactored way easier without employing any of the Vaadin Customfield logic and just write it more basic.

3.) The Contact information was now extended to also include the oidc and oidcissuer information since some of our users have it and some don't. The question is then why we still employ 2 distinct qbicuserdetails. 

4.) I also had to account for a lot of edge cases due to the HTTP responses from the orcid responses such as:

a) Special Character usage like e.g. two dashes "--" leads to a 401 response from the repository
b) Checking if the connection to the database could be made and logging errors if it doesn't
c) Invalid Entries within the sandbox instance had to be filtered (e.g. no fullname or email), Unsure if this happens in the real repository.

I really appreciate a throughout review and feedback 👍

KochTobi:
> Open Questions include:
> 
> 1.) I had to extend the project information in the database to contain the oidc information for all 3 contacts (principal investigator, project manager, person responsible). I'd rather refactor this into a seperate table, maybe in a new PR?

Hi @Steffengreiner This will break backwards compatibility with existing projects. So this is a bigger refactor. Why does it hurt to have those columns in the same table?

> 3.) The Contact information was now extended to also include the oidc and oidcissuer information since some of our users have it and some don't. The question is then why we still employ 2 distinct qbicuserdetails.

Users can login differently in our application. Either they use the spring boot way of username password (resulting in the userdetails) or they use the oauth flow for which spring boot returns an DefaultOidcUser. If you find a way to make it so that the user details service and the oauth flow both use the same object, go ahead. Until then we need to merge the database information into the spring boot DefaultOidcUser which is done in the QbicOidcUser. Similarly we need to provide an implementation of UserDetails which is QbicUserDetails. 

> 
> 4.) I also had to account for a lot of edge cases due to the HTTP responses from the orcid responses such as:
> 
> a) Special Character usage like e.g. two dashes "--" leads to a 401 response from the repository 

Which cannot be parsed and thus leads to no entry found?

> b) Checking if the connection to the database could be made and logging errors if it doesn't

Which database are you referring to?

> c) Invalid Entries within the sandbox instance had to be filtered (e.g. no fullname or email), Unsure if this happens in the real repository.

Might be the case. Email Addresses do not need to be publicly accessible.

user-interface/src/main/resources/application.properties
```diff
@@ -186,3 +186,7 @@ spring.messages.basename=messages.error-messages, messages.toast-notifications,
 # you always have to escape single quotes by another single quote ' -> ''
 # https://www.mscharhag.com/java/resource-bundle-single-quote-escaping for reference
 spring.messages.always-use-message-format=true
+#################### Orcid API calls #########################
+qbic.orcid.api.client.id=${ORCID_API_CLIENT_ID:}
+qbic.orcid.api.client.secret=${ORCID_API_CLIENT_SECRET:}
+qbic.orcid.api.endpoint.token=${ORCID_TOKEN_URI:}
```
KochTobi, line 192:
```suggestion
#################### Orcid API calls #########################
qbic.orcid.api.client.id=${spring.security.oauth2.client.registration.orcid.client-id}
qbic.orcid.api.client.secret=${spring.security.oauth2.client.registration.orcid.client-secret}
qbic.orcid.api.endpoint.token=${spring.security.oauth2.client.provider.orcid.token-uri}
```
No need for new properties

Steffengreiner: You already fixed it thanks 👍

sql/complete-schema.sql
```diff
@@ -1,581 +1,1641 @@
...
-    `id`            bigint(20) NOT NULL AUTO_INCREMENT,
-    `experiment_id` varchar(255) DEFAULT NULL,
-    `file_content`  longblob     DEFAULT NULL,
-    `fileName`      varchar(255) DEFAULT NULL,
-    PRIMARY KEY (`id`)
-) ENGINE = InnoDB
-  AUTO_INCREMENT = 1
-  DEFAULT CHARSET = utf8mb4
-  COLLATE = utf8mb4_unicode_ci;
+    `id`
```
KochTobi, line 1093: could you please undo this formatting. This is not java and I think this formatting is way harder to read.

project-management/src/main/java/life/qbic/projectmanagement/application/authorization/QbicOidcUser.java
```diff
@@ -73,10 +69,19 @@ public String getName() {
...
   }
 
   public boolean isActive() {
     return qbicUserInfo.active();
   }
+
+  public record QbicUserInfo(String userId, String fullName, String email, boolean active,
+                             String oidc, String oidcIssuer) {
+
+  }
```
KochTobi, line 86: I am confused by this change.

project-management/src/main/java/life/qbic/projectmanagement/application/authorization/QbicOidcUser.java
```diff
@@ -73,10 +69,19 @@ public String getName() {
   }
 
   public String getOidcId() {
-    return originalAuthName;
+    return Optional.ofNullable(originalAuthName).orElse(qbicUserInfo.oidc());
```
KochTobi, line 72: I am not sure you want to do this... This might interfere with DefaultOidcUser. It is assumed that when the login did not work we can safely load the information from our database. This is not correct. Am I missing something?

project-management/src/main/java/life/qbic/projectmanagement/application/authorization/QbicOidcUser.java
```diff
@@ -73,10 +69,19 @@ public String getName() {
...
+  }
+
+  public String getOidcIssuer() {
+    return Optional.ofNullable(super.getIssuer().toString()).orElse(qbicUserInfo.oidcIssuer());
```
KochTobi, line 76: Is it safe to load it from the database? I say no as the information should come only from the DefaultOidcUser filled by the auth flow.

project-management/src/main/java/life/qbic/projectmanagement/application/measurement/validation/MeasurementValidationService.java
```diff
@@ -102,7 +103,8 @@ public ValidationResult validateNGS(MeasurementUpdateInformationNGS update, Proj
   }
 
   @PreAuthorize("hasPermission(#projectId, 'life.qbic.projectmanagement.domain.model.project.Project', 'WRITE')")
-  public ValidationResult validatePxp(MeasurementRegistrationInformationPxP registration, ProjectId projectId) {
```
KochTobi, line 105: Please format only files you touched. It is very cumbersome to review changes due to formatting in files that were not changed.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/Contact.java
```diff
@@ -17,9 +17,15 @@ public class Contact {
...
 
-  public Contact(String fullName, String emailAddress) {
+  private String oidc;
+
+  private String oidcIssuer;
+
+  public Contact(String fullName, String emailAddress, String oidc, String oidcIssuer) {
     requireNonNull(fullName, "fullName must not be null");
     requireNonNull(emailAddress, "emailAddress must not be null");
+    requireNonNull(oidc, "oidc must not be null");
```
KochTobi, line 27: Why is this required? Can't I have contacts without orcid or other oidc authentication ids?

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/Contact.java
```diff
@@ -42,6 +50,13 @@ public String emailAddress() {
     return emailAddress;
   }
 
+  public String oidc() {
```
KochTobi, line 53: Is this value always present? What is the empty value? A little javadoc would be appreciated.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/Project.java
```diff
@@ -64,21 +64,27 @@ public class Project {
...
+      @AttributeOverride(name = "oidcIssuer", column = @Column(name = "principalInvestigatorOidcIssuer"))
   })
   private Contact principalInvestigator;
 
   @Embedded
   @AttributeOverrides({
       @AttributeOverride(name = "fullName", column = @Column(name = "responsibePersonFullName")),
-      @AttributeOverride(name = "emailAddress", column = @Column(name = "responsibePersonEmailAddress"))
+      @AttributeOverride(name = "emailAddress", column = @Column(name = "responsibePersonEmailAddress")),
+      @AttributeOverride(name = "oidc", column = @Column(name = "responsiblePersonOidc")),
```
KochTobi, line 86: Please adapt the complete-schema.sql file accordingly.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/Project.java
```diff
@@ -64,21 +64,27 @@ public class Project {
...
+      @AttributeOverride(name = "oidcIssuer", column = @Column(name = "projectManagerOidcIssuer"))
   })
   private Contact projectManager;
 
   @Embedded
   @AttributeOverrides({
       @AttributeOverride(name = "fullName", column = @Column(name = "principalInvestigatorFullName")),
-      @AttributeOverride(name = "emailAddress", column = @Column(name = "principalInvestigatorEmailAddress"))
+      @AttributeOverride(name = "emailAddress", column = @Column(name = "principalInvestigatorEmailAddress")),
+      @AttributeOverride(name = "oidc", column = @Column(name = "principalInvestigatorOidc")),
```
KochTobi, line 77: Please adapt the complete-schema.sql file accordingly.

project-management/src/main/java/life/qbic/projectmanagement/domain/model/project/Project.java
```diff
@@ -64,21 +64,27 @@ public class Project {
   @Embedded
   @AttributeOverrides({
       @AttributeOverride(name = "fullName", column = @Column(name = "projectManagerFullName")),
-      @AttributeOverride(name = "emailAddress", column = @Column(name = "projectManagerEmailAddress"))
+      @AttributeOverride(name = "emailAddress", column = @Column(name = "projectManagerEmailAddress")),
+      @AttributeOverride(name = "oidc", column = @Column(name = "projectManagerOidc")),
+      @AttributeOverride(name = "oidcIssuer", column = @Column(name = "projectManagerOidcIssuer"))
```
KochTobi, line 69: Please adapt the complete-schema.sql file accordingly.

project-management/src/test/groovy/life/qbic/projectmanagement/application/AddExperimentToProjectServiceSpec.groovy
```diff
@@ -45,7 +45,7 @@ class AddExperimentToProjectServiceSpec extends Specification {
                 ProjectTitle.of("Oral microbiome study"),
                 ProjectObjective.create("Analysis if tooth paste has an impact oral health and the mouth microbiome"))
         ProjectCode projectCode = ProjectCode.random()
-        Contact personReference = new Contact("John Doe", "john@doe.abcdefg")
+        Contact personReference = new Contact("John Doe", "john@doe.abcdefg", "", "")
```
KochTobi, line 48: The issue with empty values vs non-existing values. (See my comments in the Contact class)

project-management/src/test/groovy/life/qbic/projectmanagement/application/ProjectCreationServiceSpec.groovy
```diff
@@ -47,7 +47,7 @@ class ProjectCreationServiceSpec extends Specification {
     given:
     projectRepositoryStub.add(_ as Project) >> {}
     addExperimentToProjectServiceStub.addExperimentToProject(_ as ProjectId, _ as String as String, _ as List<Species> as List<Species>, _ as List<Specimen> as List<Specimen>, _ as List<Analyte>) >> {}
-    def contact = new Contact("Mustermann", "some@notavailable.zxü")
+    def contact = new Contact("Mustermann", "some@notavailable.zxü", "", "")
```
KochTobi, line 50: The issue with empty values vs non-existing values. (See my comments in the Contact class)
Think about providing different means of object creation.

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,230 @@
...
+
+  @Override
+  public List<OrcidEntry> findAll(String query, int limit, int offset) {
+    String url = paginatedQuery;
+    url += "+AND+email:*"; //require an email to be present
+    url += "+AND+given-names:*"; //require a first name to be present
+    url += "+AND+family-name:*"; //require a family name to be present
+
+    //We want to enable the user to enter a query with spaces (e.g. searching for John Doe), so we need to sanitize the query string
+    String sanitizedQuery = query.trim().replaceAll("\\s+", "+AND+");
```
KochTobi, line 129: I am not sure that this works reliably. Why not use the URLEncoder designed for this purpose?

user-interface/frontend/themes/datamanager/components/all.css
```diff
@@ -363,11 +367,11 @@ Buttons
   background-color: var(--lumo-primary-color);
 }
 
-.button-size-medium-dialog {
+.button-size-dialog-medium {
```
KochTobi, line 370: Why are those classes required? Wouldn't a selector checking for the `.button` class in a dialog be more streight forward (e.g. `.dialog-medium .button`, `.dialog-large .button` (https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) ? Where is this selector used? I cannot find any occurrence.

KochTobi (changes requested):

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -306,4 +298,80 @@ public CancelEvent(AddProjectDialog source, boolean fromClient) {
...
     }
   }
+
+  /**
+   * <b>Project Creation Information</b>
+   *
+   * <p>Down to earth project info data container.</p>
+   *
+   */
+  public static final class ProjectCreationInformation implements Serializable {
```
KochTobi, line 308: Why not a record?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -210,6 +213,11 @@ private boolean isDialogContentValid() {
   }
 
   private boolean isDialogContentInvalid() {
+    //Since we don't employ one Binder for all fields within a layout anymore,
```
KochTobi, line 216: Why not work with binder validation here? You could have a simple record with three contacts?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -153,23 +157,22 @@ private void onCancelClicked() {
   }
 
   private void onConfirmClicked(ClickEvent<Button> event) {
-    if (projectDesignLayout.validate().isInvalid()) {
+    if(projectDesignLayout.validate().isInvalid()){
       return;
     }
-    if (fundingInformationLayout.validate().isInvalid()) {
+    projectCreationInformation.setProjectDesign(projectDesignLayout.getProjectDesign());
```
KochTobi, line 163: Why did you decide to go with a mutable object that is filled step by step instead of handing off the information once at the end? What are the advantages?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
```diff
@@ -153,23 +157,22 @@ private void onCancelClicked() {
...
     if (experimentalInformationLayout.validate().isInvalid()) {
-      return;
+        return;
     }
-    fireEvent(new ConfirmEvent(this, projectDesignLayout.getProjectDesign(),
-        fundingInformationLayout.getFundingInformation(),
-        collaboratorsLayout.getProjectCollaborators(),
-        experimentalInformationLayout.getExperimentalInformation(), true));
-
+    fireEvent(new ConfirmEvent(this, projectCreationInformation, experimentalInformationLayout.getExperimentalInformation(), true));
```
KochTobi, line 175: You could construct an immutable record here. I do not see the advantage of the mutable object you fill.

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/CollaboratorsLayout.java
```diff
@@ -2,111 +2,99 @@
...
-      collaboratorsBinder.writeBean(projectCollaborators);
+      projectCollaborators.setProjectManager(managerBinding.getValue());
+      projectCollaborators.setPrincipalInvestigator(investigatorBinding.getValue());
+      //Necessary since otherwise an empty contact will be generated, which will fail during project creation in the application service
+      if (responsibleBinding.getValue().hasMinimalInformation()) {
+        projectCollaborators.setResponsiblePerson(responsibleBinding.getValue());
+      }
     } catch (ValidationException e) {
-      throw new ApplicationException("Tried to access invalid project collaborator information.", e);
+      throw new RuntimeException("Tried to access invalid project collaborators.", e);
```
KochTobi, line 82: Why the switch to a RuntimeException?

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/CollaboratorsLayout.java
```diff
@@ -2,111 +2,99 @@
...
+        fieldProjectManager);
   }
 
-  @Override
-  public String getDefaultErrorMessage() {
-    return "Please complete the mandatory information. Some input seems to be invalid.";
+  private static ContactField createRequired(String label,
+      PersonLookupService personLookupService) {
+    var contact = ContactField.createSimple(label, personLookupService);
+    contact.setRequiredIndicatorVisible(true);
```
KochTobi, line 58: should be `setRequired(true)`. You are not in a composite field here but in some layouting class. What is the difference to the createMandatory of the bound contact field?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
```diff
@@ -447,6 +457,25 @@ private Div renderContactInfo(Contact contact) {
...
+      return contactInfo;
+    }
+    if (contact.oidcIssuer().isEmpty() || contact.oidc().isEmpty()) {
+      return contactInfo;
+    }
+    var oidcType = Arrays.stream(OidcType.values())
+        .filter(ot -> ot.getIssuer().equals(contact.oidcIssuer()))
+        .findFirst();
+    if (oidcType.isPresent()) {
+      String oidcUrl = String.format(oidcType.get().getUrl()) + contact.oidc();
```
KochTobi, line 471: As the format of the urls might differ based on the OIDC type, I suggest moving this to the enum instead. oidcType.getUrlFor(contact.oidc())

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
```diff
@@ -447,6 +457,25 @@ private Div renderContactInfo(Contact contact) {
...
+      Anchor oidcLink = new Anchor(oidcUrl, contact.oidc());
```
KochTobi, line 472: The display should always be `https://orcid.org/<id>` for orcid entries. https://info.orcid.org/brand-guidelines/#h-orcid-logos-and-icons

KochTobi (changes requested):

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,226 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Repository;
+
+/**
+ * <b><class short description - 1 Line!></b>
+ *
+ * <p><More detailed description - When to use, what it solves, etc.></p>
+ *
+ * @since <version tag>
+ */
```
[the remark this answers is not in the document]
Steffengreiner: Good Point i adapted it so it's similiar to the TerminologyService API

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,226 @@
...
+
+  @Autowired
+  public OrcidRepository(
+      @Value("${qbic.external-service.person-search.orcid.client-id}") String clientID,
+      @Value("${qbic.external-service.person-search.orcid.client-secret}") String clientSecret,
+      @Value("${qbic.external-service.person-search.orcid.token-uri}") String tokenEndpoint,
+      @Value("${qbic.external-service.person-search.orcid.extended-search-uri}") String paginatedQuery) {
+    this.paginatedQuery = paginatedQuery;
+    httpClient = createHttpClient();
+    var authResponse = authenticate(httpClient, clientID, clientSecret, tokenEndpoint);
```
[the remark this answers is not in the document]
Steffengreiner: I think this is fine due to the following reasons:
1.) The [refresh token](https://info.orcid.org/ufaqs/what-are-refresh-tokens/) is solely used to ask for a new access token from the service and is just stored not employed within the application.
2.) The [access token](https://info.orcid.org/ufaqs/how-long-do-access-tokens-last/) has a lifespan of 20 years

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,226 @@
...
+  private static HttpClient createHttpClient() {
+    return HttpClient.newBuilder().version(HttpClient.Version.HTTP_2)
+        .followRedirects(HttpClient.Redirect.NORMAL).connectTimeout(
+            Duration.ofSeconds(10)).build();
+  }
+
+  private static AuthResponse authenticate(HttpClient client, String clientID, String clientSecret,
+      String tokenEndpoint) {
+    var params = Map.of("client_id", clientID, "client_secret", clientSecret, "scope",
+        "/read-public", "grant_type", "client_credentials");
```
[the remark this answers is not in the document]
Steffengreiner: Good catch 👀 it's adapted 👍

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,226 @@
...
+    if (orcid.isBlank()) {
+      return null;
+    }
+    if (fullName.isBlank()) {
+      return null;
+    }
+    if (emailList.isEmpty()) {
+      return null;
+    }
+    var email = emailList.stream().findFirst().orElse("");
```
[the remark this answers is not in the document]
Steffengreiner: done 👍

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
```diff
@@ -0,0 +1,230 @@
...
+      log.error("Error sending orcid request", e);
+      throw new OrcidRepository.QueryException("Authentication failed", e);
+    }
+
+  }
+
+  private static OrcidEntry convert(OrcidRecord orcidRecord) {
+    var emailList = Arrays.stream(orcidRecord.email()).toList();
+    //We want to show the full name meaning first and last name in a human readable format, ensuring that no trailing whitespaces are added
+    var fullName = orcidRecord.givenName() + " " + orcidRecord.familyName().trim();
```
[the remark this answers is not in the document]
Steffengreiner: done 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/overview/ProjectOverviewMain.java
```diff
@@ -163,30 +171,35 @@ private boolean isOfferSearchAllowed() {
...
-    if (confirmEvent.getFundingEntry() != null && !confirmEvent.getFundingEntry()
-        .isEmpty()) {
-      funding = Funding.of(confirmEvent.getFundingEntry().getLabel(),
-          confirmEvent.getFundingEntry().getReferenceId());
+    ProjectCreationInformation projectCreationInformation = confirmEvent.projectCreationInformation();
+    if (projectCreationInformation.getFundingEntry().isPresent()
+        && projectCreationInformation.getFundingEntry()
+        .isPresent()) {
+      funding = Funding.of(projectCreationInformation.getFundingEntry().get().getLabel(),
+          projectCreationInformation.getFundingEntry().get().getReferenceId());
```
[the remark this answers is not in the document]
Steffengreiner: Good Point i adapted it

user-interface/src/main/java/life/qbic/datamanager/views/projects/edit/EditContactsComponent.java
```diff
@@ -24,36 +24,34 @@
...
+    //We need to load the current user into the dialog so we can compare which type of user was provided
     fieldProjectManager.setMyself(currentUser, "Set myself as project manager");
     fieldProjectResponsible.setMyself(currentUser, "Set myself as project responsible");
     fieldPrincipalInvestigator.setMyself(currentUser, "Set myself as principal investigator");
 
+    investigatorBinding.setValue(projectInformation.getPrincipalInvestigator());
+    managerBinding.setValue(projectInformation.getProjectManager());
+    projectInformation.getResponsiblePerson()
+        .ifPresent(contact -> projectResponsibleBinding.setValue(contact));
+    setResponsiveSteps(new ResponsiveStep("0", 1));
```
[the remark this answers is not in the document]
Steffengreiner: This is a UI adjustement to the Formlayout containing the contactfields uses the entire dialog width. 
See [here](https://vaadin.com/docs/latest/components/form-layout#columns)

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/contact/OrcidRepository.java
Steffengreiner: That's a valid point i adapted it 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
Steffengreiner: It doesn't have to be, the display can also be set to the [compact Orcid Id f](https://info.orcid.org/documentation/integration-guide/orcid-id-display-guidelines/)ormat to save space

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
Steffengreiner: That's a good point i adjusted it 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
Steffengreiner: Valid Point, i thought it's easier to understand since the object is filled step by step but it doesn't matter in the end

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
Steffengreiner: Valid Point, i thought it's easier to understand since the object is filled step by step but it doesn't matter in the end

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
Steffengreiner: Valid Point, i thought it's easier to understand since the object is filled step by step but it doesn't matter in the end

user-interface/frontend/themes/datamanager/components/all.css
Steffengreiner: I adjusted the css class to the already existing naming schema but forgot to update the string within the java class using the css class. Good point i adjusted it.

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/CollaboratorsLayout.java
Steffengreiner: While i agree that ideally there is the possiblity to just define this field as required, it's sadly not as simple since the validation was decoupled from the `ContactField` and moved into the` BoundContactField`. This means, that the `ContactField` itself does not provide a `setRequired()` method to be set to true but does have the UI indicator that the field has to filled. This is also the difference between this method and the createMandatory method of the `BoundContactField` which requires the `ContactField` as a parameter.

user-interface/src/main/java/life/qbic/datamanager/views/projects/create/AddProjectDialog.java
Steffengreiner: This method employs the binder validation on a `ContactField` specific level, since the binding is now done within each `BoundContactField` in the `CollaboratorLayout` and not the `CollaboratorLayout` itself, which is why this workaround is necessary.

user-interface/src/main/java/life/qbic/datamanager/views/general/dialog/ButtonFactory.java
```diff
@@ -19,7 +19,7 @@ private static Button createButton(String label, String[] classNames) {
 
   public Button createConfirmButton(String label) {
     return createButton(label,
-        new String[]{"button-text-primary", "button-color-primary", "button-size-medium-dialog"});
+        new String[]{"button-text-primary", "button-color-primary", "button-size-dialog-medium"});
```
KochTobi, line 22: Is it the button size of `medium` in a dialog or the button size of a button in a medium sized dialog? I think this is an important distinction where the css selectors could add more clarification `.dialog-medium .button` in the case of the latter (which is indicated by your renaming)

Steffengreiner: Valid Point, I was adhering to the established schema of 
#Dialog css
# Button css

KochTobi (approved):

KochTobi (approved):

KochTobi (approved):

PR #993, opened 2025-01-17
Close upload dialog on navigation change

sven1103:

PR #994, opened 2025-01-17
Close open dialogs

sven1103:


Steffengreiner (changes requested):
Heya @sven1103, awesome work with closing the dialogs. Sadly we have a lot of dialogs that have to be closed dependent on the state a project is in, so there are still dialogs that need your help!

Maybe it makes more sense, that the basic dialog class from which we extend the dialogs listens and closes itself if a BeforeLeaveEvent occurs to make sure we don't miss any of them in the future? 

I compiled a list of dialogs that have to be adapted with screenshots below:

1.) Cannot Edit Exp. Group Dialog
2.) Cannot Edit Exp. Variable Dialog
3.) Add Collaborator Dialog
4.) Edit Project Design Dialog
5.) Edit Funding Information Dialog
6.) Edit Project Contact Dialog
7.) Batch Deletion Dialog 


Screenshot 2025-01-17 at 17 00 44
Screenshot 2025-01-17 at 17 02 11

Screenshot 2025-01-17 at 17 04 22
Screenshot 2025-01-17 at 17 06 11
Screenshot 2025-01-17 at 17 06 45
Screenshot 2025-01-17 at 17 07 15
Screenshot 2025-01-17 at 17 17 25

sven1103:
Dear Steffen,  yes i see the problem as well. I am just not sure that this is the strategy we should aim for. The dialog is in context of its surrounding view, and immediate closing them on navigation might not be what we want (e.g. user input might get lost). Our design so far, also with the new AppDialog is to leave the dialog simple and not having any dependencies on surrounding events. I dont know what Shraddha has to say here, maybe we should also not overdo it with dialogs?

sven1103:
addressed in #993

PR #1001, opened 2025-02-03
Provide pooling name in measurement overview

KochTobi:


KochTobi:
The code works so far. However I will now attempt to move some old code to the new way we do dialogs

sven1103 (changes requested):
The sample pool name given by the user is now displayed in the sample column, however I also see it in the comment section, which i do not expect it to be there:

image

KochTobi:
Found it after looking for some time. Fixed

sven1103 (changes requested):
Unfortunatley, I don't see any pools anyomore now.

KochTobi:
> Unfortunatley, I don't see any pools anyomore now.

Sorry but I can't reproduce this. For me the pools show up as desired
image

Issue #1004, opened 2025-02-06
GOAWAY received in terminology search


KochTobi:
### What happened?

There are empty rows in the terminology list

![Image](https://github.com/user-attachments/assets/37bfde36-8a3d-424a-a8d9-9a471d7f9254)

### Relevant log output

```shell
life.qbic.application.commons.ApplicationException: Lookup for term 'blood' failed.
        at life.qbic.projectmanagement.application.ontology.TerminologyService.query(TerminologyService.java:55)
Caused by: life.qbic.projectmanagement.application.ontology.LookupException: Terminology service search failed. Service might not be reachable
        at life.qbic.projectmanagement.infrastructure.ontology.TIBTerminologyServiceIntegration.wrap(TIBTerminologyServiceIntegration.java:109)
Caused by: java.io.IOException: /<censored>: GOAWAY received
        at java.net.http/jdk.internal.net.http.HttpClientImpl.send(HttpClientImpl.java:966)
Caused by: java.io.IOException: /<censored>: GOAWAY received
        at java.net.http/jdk.internal.net.http.Http2Connection.handleGoAway(Http2Connection.java:1362)

Note I censored IP addresses with <censored>
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sven1103:
might be related to https://github.com/qbicsoftware/data-manager-app/issues/997

PR #1006, opened 2025-02-07
Update default values for the TIB environment variables

KochTobi:
The address of the service API changed.
See https://api.terminology.tib.eu/swagger-ui/index.html#/ for more information.


KochTobi:
fixes #1004 
closes #994

PR #1012, opened 2025-02-11
Fix issue with sample creation

KochTobi:
The openbis api only specifies a List but expects a mutable List.


Steffengreiner (approved):
Well done 👍

PR #1025, opened 2025-02-17
Batch update must be idempotent with respect to sample ids

sven1103:
Ensures that sample IDs already referenced in the sample batch are not added again, which led to duplicates.


KochTobi (approved):
I could no longer reproduce the issue

PR #1038, opened 2025-02-21
Add API objects for project information

sven1103:
Provides the API to handle more project information tasks.

### DIFF
--- a/user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
+++ b/user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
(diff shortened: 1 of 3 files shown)


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
```diff
@@ -530,14 +533,18 @@ private void handleUpdateEvent(ProjectDesignUpdateEvent event) {
     var request = new ProjectUpdateRequest(project,
         new ProjectDesign(info.getProjectTitle(), info.getProjectObjective()));
 
-    asyncProjectService.update(request)
+    submit(request, req -> asyncProjectService.update(req)
```
KochTobi, line 536:
```suggestion
    submit(request, requestConsumer -> asyncProjectService.update(requestConsumer)
```

KochTobi: where do you make use of the request itself here?
sumbit takes a request as first parameter and a consumer as second. Maybe I misunderstood something here?

KochTobi (approved):
LGTM

PR #1041, opened 2025-02-24
Support request cache

sven1103:


user-interface/src/main/java/life/qbic/datamanager/RequestCache.java
```diff
@@ -0,0 +1,61 @@
...
+   * @since 1.9.0
+   */
+  public Optional<CacheableRequest> get() throws CacheException {
+    VaadinSession session = VaadinSession.getCurrent();
+    if (session != null) {
+      return Optional.ofNullable(
+          (CacheableRequest) session.getAttribute(CacheableRequest.class.getName()));
+    }
+    throw new CacheException("No Vaadin session found");
+  }
```
KochTobi, line 52: Should there only be one request per session? When multiple async editions are performed, this might overwrite a request that was sent before it completed. Further the API does not allow the calling side to remove a request once successfully completed. Is this intentional?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/info/ProjectSummaryComponent.java
```diff
@@ -560,6 +570,9 @@ private void handleSuccess(ProjectUpdateResponse response) {
   private void handleRequestFailed(RequestFailedException error) {
```
KochTobi, line 570: We probably have to remove the request or timestamp it once it resulted in a successful response.

KochTobi:
Hi there
I am unsure whether one request per session is a viable solution. When is the cache cleared? What happens when a request is overwritten in the cache due to processing still going on?
What do you think about these questions?

KochTobi (approved):
LGTM

PR #1042, opened 2025-02-24
Provide one button app dialog

sven1103:
You can now call build a `DialogFooter` with only a confirm button:

```java
DialogFooter.withConfirmOnly(AppDialog dialog, String confirmText)
```
This will show only one button (the primary button, that executes the confirm action).

### DIFF
--- a/user-interface/src/main/java/life/qbic/datamanager/views/demo/ComponentDemo.java
+++ b/user-interface/src/main/java/life/qbic/datamanager/views/demo/ComponentDemo.java
@@ -220,6 +220,39 @@ public class ComponentDemo extends Div {
     return steps;
   }
 
+  private static Div dialogWithOneButton(AppDialog dialog, String dialogType) {
+    Div content = new Div();
+    Button showDialog = new Button("Show Dialog");
+    // Dialog set-up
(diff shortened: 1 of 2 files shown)


user-interface/src/main/java/life/qbic/datamanager/views/demo/ComponentDemo.java
```diff
@@ -220,6 +220,39 @@ public UserInput userInput() {
...
+    });
+
+    dialog.registerCancelAction(() -> {
+      dialog.close();
+      if (dialog.hasChanges()) {
+        confirmBox.setText("Cancelled the dialog although there where changes made!");
+      } else {
+        confirmBox.setText("Cancelled the dialog. No changes.");
+      }
+    });
```
KochTobi, line 245: Is this even possible when the footer only has a confirm button?

KochTobi (approved):

user-interface/src/main/java/life/qbic/datamanager/views/demo/ComponentDemo.java
sven1103: you can set it, but it cant be executed :)

PR #1050, opened 2025-03-04
Provide ontology term in memory cache

sven1103:
In order to mitigate network interrupts or shorter remote service down-times, this PR
introduces a very simple in-memory cache to recover in the case of term validation.

### DIFF
--- a/project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/TIBTerminologyServiceIntegration.java
+++ b/project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/TIBTerminologyServiceIntegration.java
@@ -15,9 +15,13 @@ import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/ontology/TIBTerminologyServiceIntegration.java
```diff
@@ -174,9 +180,12 @@ public List<OntologyClass> query(String searchTerm, int offset, int limit)
...
   public Optional<OntologyClass> searchByCurie(String curie) throws LookupException {
     try {
-      return searchByOboIdExact(curie).map(TIBTerminologyServiceIntegration::convert);
+      return searchByOboIdExact(curie).map(this::updateCache)
+          .map(TIBTerminologyServiceIntegration::convert);
     } catch (IOException e) {
-      throw wrapIO(e);
+      // this happens on network interrupts or if the remote service is down
+      // we try to recover from the cache
+      return cache.findByCurie(curie).map(TIBTerminologyServiceIntegration::convert);
```
KochTobi, line 188: You handle the exception here. It would be good to know that it occurred otherwise we are not informed of it. Maybe add a log entry here (warn or error)

KochTobi (changes requested):
Only one minor suggestion. Otherwise looks good

KochTobi (approved):

PR #1060, opened 2025-03-07
Optimise imports

sven1103:


KochTobi (commented):
The change in import order is caused by IntelliJ?

KochTobi (changes requested):
This PR contains a commit not related to the PR
https://github.com/qbicsoftware/data-manager-app/pull/1060/commits/0d510ef5b3f889677f93bbe30039652942aec02a
Is this intentional?

PR #1063, opened 2025-03-10
Implement API for sample management

sven1103:
Provides the async service implementation for the sample management related requests.

### DIFF
--- a/user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
+++ b/user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
@@ -17,9 +17,9 @@ import life.qbic.application.commons.SortOrder;
 import life.qbic.datamanager.views.Context;
 import life.qbic.datamanager.views.general.PageArea;
 import life.qbic.datamanager.views.general.Tag;
+import life.qbic.datamanager.views.notifications.MessageSourceNotificationFactory;
 import life.qbic.projectmanagement.application.api.AsyncProjectService;
 import life.qbic.projectmanagement.application.api.AsyncProjectService.RequestFailedException;
-import life.qbic.projectmanagement.application.sample.SampleInformationService;
 import life.qbic.projectmanagement.application.sample.SamplePreview;
 import life.qbic.projectmanagement.domain.model.batch.Batch;
 import life.qbic.projectmanagement.domain.model.experiment.Experiment;
@@ -49,10 +49,13 @@ public class SampleDetailsComponent extends PageArea implements Serializable {
   private final Span countSpan;
   private final Grid<SamplePreview> sampleGrid;
   private final transient AsyncProjectService asyncProjectService;
+  private final MessageSourceNotificationFactory messageFactory;
   private Context context;
 
   @Autowired
-  public SampleDetailsComponent(AsyncProjectService asyncProjectService) {
+  public SampleDetailsComponent(AsyncProjectService asyncProjectService,
+      MessageSourceNotificationFactory messageFactory) {
+    this.messageFactory = Objects.requireNonNull(messageFactory);
     this.asyncProjectService = Objects.requireNonNull(asyncProjectService);
     addClassName("sample-details-component");
     sampleGrid = createSampleGrid();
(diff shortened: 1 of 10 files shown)


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -155,24 +160,32 @@ private static Grid<SamplePreview> createSampleGrid() {
...
           .collect(Collectors.toList());
       // if no order is provided by the grid order by last modified (least priority)
       sortOrders.add(SortOrder.of("sampleCode").ascending());
-      return sampleInformationService.queryPreview(experimentId, query.getOffset(),
-          query.getLimit(), List.copyOf(sortOrders), filter).stream();
+      return Objects.requireNonNull(
+          asyncProjectService.getSamplePreviews(projectId.value(), experimentId.value(),
+                  query.getOffset(), query.getLimit(), List.copyOf(sortOrders), filter).collectList()
+              .doOnError(RequestFailedException.class, this::handleRequestFailed)
+              .block()).stream();
```
KochTobi, line 178: How about https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#toStream-- instead?
Maybe you do not even need to block then (but I don't know this).
https://projectreactor.io/docs/core/release/reference/apdx-operatorChoice.html#which.blocking
Seems to me that a stream is returned (which is expected by vaadin) so the whole collection into a list and streaming afterwards might not be needed.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
```diff
@@ -155,24 +160,32 @@ private static Grid<SamplePreview> createSampleGrid() {
...
     });
     sampleGrid.getLazyDataView().addItemCountChangeListener(
         countChangeEvent -> setSampleCount((int) sampleGrid.getLazyDataView().getItems().count()));
     sampleGrid.recalculateColumnWidths();
   }
 
+  private void handleRequestFailed(RequestFailedException e) {
+
```
KochTobi, line 186: What was done before when the failure case occurred?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
```diff
@@ -474,12 +479,10 @@ private void setBatchAndSampleInformation() {
...
   private boolean noSamplesRegisteredInExperiment(Experiment experiment) {
-    return sampleInformationService.retrieveSamplesForExperiment(experiment.experimentId())
-        .map(Collection::isEmpty)
-        .onError(error -> {
-          throw new ApplicationException("Unexpected response code : " + error);
-        })
-        .getValue();
+    var result = asyncProjectService
+        .getSamples(context.projectId().orElseThrow().value(), experiment.experimentId().value())
+        .collectList().block();
```
KochTobi, line 484: Instead of waiting for all elements to be there you could only wait for the first element as it is your concern whether the list is empty.
https://projectreactor.io/docs/core/release/reference/apdx-operatorChoice.html#which.blocking
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#blockFirst-java.time.Duration-

KochTobi (dismissed):

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleDetailsComponent.java
sven1103: nothing? :D

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
sven1103: i think cheats the way Flux is working: we don't know how many samples will be send through the Flux. So maybe it is better to use a convenience Method in the Service API instead?

KochTobi: Why would this cheat the way flux is working?
The flux publishes elements and instead of waiting for the flux to complete, you can simply wait for the first element. No need for the flux to continue publishing afterwards as you do not want to use the result.

KochTobi: Of course you could think about adding a method Mono<Boolean> to determine whether there are any samples in an experiment. Depending on the way the flux is created this would minimize the work the system does. But I think both are legitimate ways to handle the situation.

KochTobi (approved):
LGTM

PR #1065, opened 2025-03-11
Update experiment API

KochTobi:


project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -360,7 +410,7 @@ record ExperimentalVariables(List<ExperimentalVariable> experimentalVariables) i
    * @param unit         the unit for the value of the level. Can be null if no unit is set
    * @since 1.9.0
    */
-  record VariableLevel(String variableName, String levelValue, @Nullable String unit) {
+  record VariableLevel(Long variableId, String variableName, String levelValue, @Nullable String unit) {
```
Steffengreiner, line 413: Please also update the JD

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -244,8 +244,9 @@ sealed interface ProjectUpdateResponseBody permits FundingInformation, ProjectCo
 
   }
 
-  sealed interface ExperimentUpdateRequestBody permits ConfoundingVariables, ExperimentDescription,
-      ExperimentalGroups, ExperimentalVariables {
+  sealed interface ExperimentUpdateRequestBody permits ConfoundingVariableAdditions,
+      ConfoundingVariableDeletions, ConfoundingVariableUpdates, ExperimentDescription,
+      ExperimentalGroups, ExperimentalVariableAdditions, ExperimentalVariableDeletions {
 
   }
```
Steffengreiner, line 251: Since i'm not up to date i assume that i missed something so here's my question:

What's the advantage of having an interface that could be any of these possibilities? 
As i saw below, this leads to the burden of the service implementation to account for all cases individually. 
Could this be seperated into individual use cases to avoid an extending switch statement?

Steffengreiner (approved):
Heya @KochTobi, I tested it out and it works, nicely done! However i have a question and minor request for updated JD which would be nice if they were addressed

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
KochTobi: It could be separated into different methods which would make it easier to specify what should be done with the input. Deciding based on what was sent works too.

KochTobi:
> Heya @KochTobi, I tested it out and it works, nicely done! However i have a question and minor request for updated JD which would be nice if they were addressed

Hi Steffen, the changes I introduced you cannot really test out. This PR only adds the interface methods for changing the experimental variables, experimental groups and the confounding variables of experiments. It does not offer any implementation.

Steffengreiner (approved):
Thanks for the explanation, makes sense to me.
To clarify i tested if any of the current processes break which doesn't happen 👍

PR #1069, opened 2025-03-11
Implement refactor for RO-Create download

sven1103:


KochTobi:
There are multiple test failures (related to the pom changes?)

project-management/src/main/java/life/qbic/projectmanagement/application/VirtualThreadScheduler.java
```diff
@@ -1,4 +1,4 @@
-package life.qbic.datamanager;
+package life.qbic.projectmanagement.application;
```
KochTobi, line 1: I think the VirtualThreadScheduler can be in application commons even. 
It is not bound to any domain nor to the ui and I think putting it in the globally accessible package makes the most sense. What do you think?

user-interface/src/main/java/life/qbic/datamanager/files/export/FileNameFormatter.java
```diff
@@ -1,5 +1,6 @@
 package life.qbic.datamanager.files.export;
 
+import java.time.Instant;
```
KochTobi, line 3: Why the import of `java.time.Instant`? I couldn't find the usage of this import.

KochTobi (changes requested):
2 small suggestions

project-management/src/main/java/life/qbic/projectmanagement/application/VirtualThreadScheduler.java
sven1103: I opt for delaying it, until we see that other modules also need this concept :)

KochTobi (approved):

PR #1081, opened 2025-03-19
Provide metadata validation API

sven1103:
Introduces the concept of validation for the async service API.


project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -222,6 +223,20 @@ Flux<SamplePreview> getSamplePreviews(String projectId, String experimentId, int
...
+   *
+   * @param requests a list of one or more {@link ValidationRequest}.
+   * @return a {@link Flux} of {@link ValidationResponse}, providing the validation results for the
+   * submitted requests. Exceptions are provided as {@link Flux#error(Throwable)}.
+   * @throws UnknownRequestException if an unknown request has been used in the service call
+   * @throws RequestFailedException  if the request was not successfully executed
+   * @throws AccessDeniedException   if the user has insufficient rights
+   * @since 1.10.0
+   */
+  Flux<ValidationResponse> validate(List<ValidationRequest> requests);
```
KochTobi, line 237: Wouldn't it make sense to allow for a `Flux<ValidationRequest>` as input here? This would allow the caller to choose when to emit requests. (and can also be a Flux.just call to wrap a list).

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -662,26 +696,44 @@ record ProjectCreationResponse(String projectId) {
...
+      }
+      if (projectId.isBlank()) {
+        throw new IllegalArgumentException("Project ID cannot be blank");
+      }
+      if (requestId != null && requestId.isBlank()) {
+        requestId = null;
+      }
+      if (requestId == null) {
+        requestId = UUID.randomUUID().toString();
+      }
```
KochTobi, line 719: You can combine those 2 statements into
```suggestion
      if (requestId == null || requestId.isBlank()) {
        requestId = UUID.randomUUID().toString();
      }
```
This way you avoid temporal coupling for the `isBlank` case.

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -697,6 +749,9 @@ record ProjectUpdateResponse(String projectId, ProjectUpdateResponseBody respons
       if (requestId != null && requestId.isBlank()) {
         requestId = null;
       }
+      if (requestId == null) {
+        requestId = UUID.randomUUID().toString();
+      }
```
KochTobi, line 754: You can combine those two if statements to avoid temporal coupling
```suggestion
      if (requestId == null || requestId.isBlank()) {
        requestId = UUID.randomUUID().toString();
      }
```

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -709,21 +764,62 @@ public Optional<String> retrieveRequestId() {
...
+      if (projectId.isBlank()) {
+        throw new IllegalArgumentException("Project ID cannot be blank");
+      }
+      if (requestId != null && requestId.isBlank()) {
+        requestId = null;
+      }
+      if (requestId == null) {
+        requestId = UUID.randomUUID().toString();
+      }
```
KochTobi, line 800: You can combine those if statements to avoid temporal coupling
```suggestion
      if (requestId == null || requestId.isBlank()) {
        requestId = UUID.randomUUID().toString();;
      }
```

project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
sven1103: So far we have not integrated Flux as parameter, and i wanted to keep it simple / in alignment with the API we have.

KochTobi: I still think it makes sense here. In contrast to the experimental variables (where we want a all or nothing return), here we are fine with providing the requests one by one.

PR #1082, opened 2025-03-19
API for requesting templates for sample registration and update

sven1103:
Introduces a reactive API for requesting templates to register or update samples for a data manager experiment.


KochTobi (approved):
LGTM

PR #1100, opened 2025-03-25
Implement template API 

sven1103:
Provides templates behind the reactive service for 

- sample registration
- sample updates
- sample information

Architectural changes:

- the Template Service is now an application service of the project management domain
- the Template Provider interface allows for more than one MimeType support for templates
- a new OpenXML template provider implementation to support spreadsheet templates
- spreadsheet generation is now in the infrastructure layer, where it belongs (I/O task!)
- template resources are now requested as generic Digital Objects and decoupled from concrete implementations

hint: the duplicates are a non-breaking fail safe for the measurement templates, that need to be refactored in an own unit of task.


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/template/provider/openxml/factory/SampleInformationFactory.java
```diff
@@ -1,30 +1,29 @@
...
+import life.qbic.projectmanagement.infrastructure.template.provider.openxml.column.ConfoundingVariableColumn;
+import life.qbic.projectmanagement.infrastructure.template.provider.openxml.column.InformationColumn;
+import life.qbic.projectmanagement.infrastructure.template.provider.openxml.util.XLSXTemplateHelper;
 import org.apache.poi.ss.usermodel.CellStyle;
 import org.apache.poi.ss.usermodel.CreationHelper;
 import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
 
-public class SampleInformationFactory implements WorkbookFactory {
+class SampleInformationFactory implements WorkbookFactory {
```
KochTobi, line 26: I like it when public gets removed :+1:

project-management/src/main/java/life/qbic/projectmanagement/application/api/template/TemplateService.java
```diff
@@ -0,0 +1,292 @@
...
+      ConfoundingVariableService confVariableService, TemplateProvider templateProvider) {
+    this.experimentService = Objects.requireNonNull(experimentService);
+    this.sampleService = Objects.requireNonNull(sampleService);
+    this.confVariableService = confVariableService;
+    this.templateProvider = Objects.requireNonNull(templateProvider);
+  }
+
+  private static Predicate<Sample> inBatchPredicate(String targetBatchId) {
+    return sample -> sample.assignedBatch().value().equals(targetBatchId);
+  }
```
KochTobi, line 63:
```suggestion
  private static Predicate<Sample> isInBatch(String targetBatchId) {
    return sample -> sample.assignedBatch().value().equals(targetBatchId);
  }
```

project-management/src/main/java/life/qbic/projectmanagement/application/api/template/TemplateService.java
```diff
@@ -0,0 +1,292 @@
...
+            sampleExtension.confoundingVariableLevels())
+    ));
+  }
+
+  private SampleExtension querySampleExtension(Experiment experiment, String projectId,
+      String experimentId) {
+    var experimentalGroups = experiment.getExperimentalGroups();
+    var samples = sampleService.retrieveSamplesForExperiment(
+        ProjectId.parse(projectId), experimentId).stream().toList();
+    var confoundingVariablesIds = confVariableService.listConfoundingVariablesForExperiment(
```
KochTobi, line 192: I don't think you need the confounding variables themselves? They are not used in the returned SampleExtension as far as I can see. I think they are part of your basic information and the sample extension only holds the levels of the samples.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
```diff
@@ -196,21 +203,71 @@ private void downloadSampleMetadata() {
...
-        return templateService.sampleBatchInformationXLSXTemplate(projectId.value(),
-            experimentId.value());
-      }
-    });
+    asyncProjectService.sampleInformationTemplate(projectId.value(), experimentId.value(),
+            OPEN_XML)
+        .doOnSubscribe(showInProgress("preparing sample information"))
+        .doOnSuccess(resource -> handleSuccess(projectCode, experimentName).accept(resource))
+        .doOnError(this::handleError)
+        .doFinally(this::closePendingTaskNotification)
```
KochTobi, line 212: I am a bit unseasy because the toast handling can be UI specific. Shouldn't we capture the UI in question and forward it to the single methods? What happens if the user navigates while a task is in progress?

KochTobi: Is this covered by `@UiScope`? Do the toasts close with navigation or do they stay open?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
```diff
@@ -94,9 +99,10 @@ public class SampleInformationMain extends Main implements BeforeEnterObserver {
   private final transient CancelConfirmationDialogFactory cancelConfirmationDialogFactory;
   private final transient MessageSourceNotificationFactory notificationFactory;
   private final transient SampleValidationService sampleValidationService;
-  private final transient TemplateService templateService;
   private final transient SampleRegistrationServiceV2 sampleRegistrationServiceV2;
   private final transient AsyncProjectService asyncProjectService;
+  private final MessageSourceNotificationFactory messageFactory;
+  private Toast pendingTask;
```
KochTobi, line 105: What happens if multiple tasks are started? The first task would then remove the toast disregarding other ongoing tasks in the background. Your assumption that there is one operation at a time and that the user waits for that to finish should go into the code (as comment for example) if it is desired.

KochTobi (changes requested):
Overall the code looks good. I like the abstractions you made. Some minor comments and one concern regarding the toast/state handling of ongoing operations.

project-management/src/main/java/life/qbic/projectmanagement/application/api/template/TemplateService.java
sven1103: They are required to make the call to the service to fetch the levels :)

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
sven1103: I agree, I will work on a solution

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/samples/SampleInformationMain.java
sven1103: I understand your concern, however i think it is beyond the scope of this PR and consider this as an own investigation / issue, since it concerns the notification subsystem as a hole.

project-management/src/main/java/life/qbic/projectmanagement/application/api/template/TemplateService.java
KochTobi: Ah you are right

KochTobi:
The file naming is wrong. 
Previously: `QVAMP_sample-metadata-edit-template_v1.xlsx`
Now: `2025-04-02_QVAMP_sample-metadata-update-template.xlsx`
I think the expected behaviour is to have the template version (and the date)

sven1103:
> For invalid information, It is possible to confirm the dialog (which it should not be) [2025-04-02_QVAMP_sample-metadata-update-template-1-1.xlsx](https://github.com/user-attachments/files/19565040/2025-04-02_QVAMP_sample-metadata-update-template-1-1.xlsx)
> 
> In my case, the terminology validation failed leading to an error displayed but possible batch update without any errors. Also can be observed with the file provided above (missing species)

I checked out the branch and I cannot reproduce your findings. The validation fails and a click on the "Save" button does not do anything, which is the supposed behaviour.

KochTobi (approved):
LGTM

PR #1123, opened 2025-04-08
Implement adding and deletion.of experimental variables

KochTobi:


project-management/src/main/java/life/qbic/projectmanagement/application/experiment/ExperimentInformationService.java
```diff
@@ -318,6 +318,70 @@ public void addVariableToExperiment(String projectId, ExperimentId experimentId,
...
+    List<DomainEvent> domainEventsCache = new ArrayList<>();
+    var localDomainEventDispatcher = LocalDomainEventDispatcher.instance();
+    localDomainEventDispatcher.reset();
+    localDomainEventDispatcher.subscribe(
+        new ExperimentUpdatedDomainEventSubscriber(domainEventsCache));
+
+    Experiment experiment = loadExperimentById(experimentId);
+    List<ExperimentalVariableInformation> addedVariables = new ArrayList<>();
+    try {
+      for (ExperimentalVariableAddition variableAddition : variableAdditions) {
```
sven1103, line 364: Since some of our concepts have a long name that introduce visual clutter, you can make more use of the `var` keyword, which makes it easier to read, especially if the definition of variable has only a local scope.

sven1103 (changes requested):
Overall looks good, i made provided some minor comments. 

Will checkout the UI part and test it locally as well.

sven1103 (approved):
LGTM

PR #1124, opened 2025-04-08
Introduce ontology term concept for service API

sven1103:


project-management/src/main/java/life/qbic/projectmanagement/application/api/AsyncProjectService.java
```diff
@@ -768,6 +770,33 @@ record ProjectCreationRequest(ProjectDesign design, ProjectContacts contacts,
...
+   * for humans, its assigned OBO identifier and its globally unique identifier.
+   *
+   * @param label an {@link OntologyTerm} label for visualisation
+   * @param oboId the assigned OBO identifier
+   * @param id    the globally unique identifier of the term
+   * @since 1.10.0
+   */
+  record OntologyTerm(String label, Curie oboId, URI id) {
+
+  }
```
KochTobi, line 784: In this case it could be useful to introduce null checks at least for the `Curie`

KochTobi (approved):
LGTM

PR #1125, opened 2025-04-08
Implement experiment query API

sven1103:
Implements the reactive API for querying experimental information.

Wires the async service call in the UI and renders information about species, specimen and analyte dynamically.


Steffengreiner (approved):
Well done 👍

PR #1134, opened 2025-04-24
Add service api documentation

sven1103:
Provides some class diagrams to provide an overview of the service API structure.

### DIFF
--- a/service_api.md
+++ b/service_api.md
@@ -271,7 +271,7 @@ To enable clients to update a certain sub-resource, a combination of the parent 
 ID of the sub-resource is required.
 
 For example to update an `Experimental Group` of an `Experiment`, the API shall provide a method like `update(ExperimentalGroupUpdateRequest request)`. The object request needs to provide
-the required IDs to ensure that the service can perform the task.
+the required IDs to ensure that the service can perform the task. Each request is transactional for the resource.
 
 ## Delete Example
 
--- a/service_api.md
+++ b/service_api.md
@@ -1,0 +1,47 @@
+# Service API
+
+```mermaid
+---
+title: Update Example
+---
+classDiagram
+    class AsyncProjectService {
+        Mono~ProjectUpdateResponse~ update(ProjectUpdateRequest request)
+    }
+    <<interface>> AsyncProjectService
+    AsyncProjectService ..> ProjectUpdateRequest
+    AsyncProjectService ..> ProjectUpdateResponse
+    
+    class ProjectUpdateRequest {
+        String projectId
+        ProjectUpdateRequestBody requestBody
+        String requestId
+    }
+    
+    class ProjectUpdateResponse {
+        String projectId
+        ProjectUpdateResponseBody responseBody
+        String requestId
+    }
+    
+    class ProjectUpdateRequestBody {
+       
+    }
+    <<interface>> ProjectUpdateRequestBody
+    
+    class ProjectUpdateResponseBody {
+        
+    }
+    <<interface>> ProjectUpdateResponseBody
+    
+    
+    class FundingInformation 
+    FundingInformation ..|> ProjectUpdateRequestBody
+    FundingInformation ..|> ProjectUpdateResponseBody
+  
+    ProjectUpdateRequest ..> ProjectUpdateRequestBody
+    ProjectUpdateResponse ..> ProjectUpdateResponseBody
+    
+    
+    
+```
--- a/service_api.md
+++ b/service_api.md
@@ -1,68 +1,186 @@
 # Service API
 
-The idea of the service API is to create a uniform interface to interact with Data Manager ressources from a user agent, for example the browser. Since the project uses Vaadin, the API is written in Java and provides an easy integration
+The idea of the service API is to create a uniform interface to interact with Data Manager
+ressources from a user agent, for example the browser. Since the project uses Vaadin, the API is
+written in Java and provides an easy integration
 in Vaadin Flow.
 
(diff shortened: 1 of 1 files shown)


Steffengreiner (changes requested):
I don't understand the mermaid chart flow logic for all examples.
As far as i understood the flow should be as follows:
1.) Client calls the `create()` method within the `AsyncProjectService` with the `Object<Process>Request`. 
2.) Through reactive magic, the information stored within the object to be processed is stored somewhere in the system and a response object is created, containing the <processed> object with a unique id. 
3.) The client calling the create() method should then receive the response object from the service contained in a `mono `class

This is not what the diagram flows shows however. According to the mermaid chart the client will never receive the response object since the arrows only point downwards.
Am i missing something?

service_api.md
```diff
@@ -0,0 +1,284 @@
...
+    class ProjectInformation
+
+    ProjectCreationRequest ..> ProjectInformation
+    ProjectCreationResponse ..> ProjectInformation
+```
+
+### Creating a sub-resource
+Creating a sub-resource that are in a many-to-one relationship with their parent resource should be made easy.
+
+For example, creating an `Example Group` for an existing `Experiment` shall be a no-brainer. Therefore, it is
```
KochTobi, line 67: I think this is `Experimental Group` instead of `Example Group`. If you intended to write `Example Group` here, it should be `Example Group` in the create method as well.

service_api.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Get all experiments for a project
+
+```mermaid
+---
+title: Get Experiments Example
+---
+classDiagram
+    class AsyncProjectService {
+        Flux~ExperimentInformation~ getExperiment(String projectId, String experimentId)
```
KochTobi, line 158: This will not compile as the method name and parameters are the same and thus java cannot decide which method to use. This method should be named `getExperiments` with an `s`.

service_api.md
```diff
@@ -0,0 +1,284 @@
...
+## Update examples
+
+Updates of resources can be done partially, for example by providing the information for the
+smallest information container the API provides for a certain ressource. For example
+`Project Information` contains
+`Funding Information`. The API shall provide partial updates of a project for `Funding Information`
+without
+providing all other information properties of a project.
+
+The size and semantics for the smallest information unit of a resource are subject to
```
KochTobi, line 172: Important :+1:

service_api.md
```diff
@@ -0,0 +1,284 @@
...
+
+Sometimes a resource contains more than one sub-resource of the same type. This happens, when there
+is a one-to-many relationship between the resource and its sub-resource. For example an experiment can
+contain one or more experimental groups.
+
+To enable clients to update a certain sub-resource, a combination of the parent resource's ID and the local 
+ID of the sub-resource is required.
+
+For example to update an `Experimental Group` of an `Experiment`, the API shall provide a method like `update(ExperimentalGroupUpdateRequest request)`. The object request needs to provide
+the required IDs to ensure that the service can perform the task.
```
KochTobi, line 274: When creating/updating multiple sub-resources, how do we deal with rollback strategies? What do we do in case of failures? When we return only the changed information we do not have the possibility to rollback single executions? Do we require each request in the update method to be `all-or-nothing`? So will a desired update of multiple groups lead to a request body with information about multiple groups or do we want to have multiple calls to update with multiple request bodies with a single group each?

service_api.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Delete Example
+
+Deletion of a resource or sub-resource shall be possible with the resource ID or the combination of the resource ID
+and its sub-resource local ID.
+
+For example an `Experimental Group 1` in an `Experiment X` for `Project A` should be deletable by the client by e.g. calling `delete(ExperimentalGroupDeletionRequest request)`. The combination of project ID, experiment ID and experimental group ID ensures the
+identity of the resource manipulation,
```
KochTobi, line 282: Same question as for the update of resources. Do I require all deletions to fail if one fails? How to recover from this?

service_api.md
sven1103: If there is a use case, where a user wants to submit an update of more than one sub-resource, then we need to extend the API for it (e.g. a list of `ExperimentalGroup`. 


Each request is transactional for the resource.

service_api.md
sven1103: Please see my previous answer :)

sven1103:
> I don't understand the mermaid chart flow logic for all examples. As far as i understood the flow should be as follows: 1.) Client calls the `create()` method within the `AsyncProjectService` with the `Object<Process>Request`. 2.) Through reactive magic, the information stored within the object to be processed is stored somewhere in the system and a response object is created, containing the object with a unique id. 3.) The client calling the create() method should then receive the response object from the service contained in a `mono `class
> 
> This is not what the diagram flows shows however. According to the mermaid chart the client will never receive the response object since the arrows only point downwards.
> 
> Additionally, the contained information within the request object should be shown linked to the request object and not appear in the middle of the diagram (e.g. Funding Information/Project Contacts)

Dear @Steffengreiner , 
mermaid is an engine that renders diagrams of different types. This is a class diagram, not a flow diagram.

service_api.md
KochTobi, line 274:
```suggestion
the required IDs to ensure that the service can perform the task. Each request is transactional for the resource.
```

service_api.md
KochTobi:
```suggestion
identity of the resource manipulation. Each request is transactional for the resource.
```

PR #1143, opened 2025-05-09
Introduce announcement banner for maintenance notification

KochTobi:
image

PR #1145, opened 2025-05-14
Provide adaptive banner space with css

Steffengreiner:
Provide allocated space with adaptive css layout and account for space below project collection


Steffengreiner:
Draftwork is included in #1143

PR #1153, opened 2025-05-26
Fix Link to documentation for data download 

Steffengreiner:
The data download section link is currently broken since it does not account for different version documentation in its URL


Steffengreiner:
Closed since it's not checked out from main see #1158

PR #1155, opened 2025-05-26
Fix Typo in Confounding Variable Registration Title

Steffengreiner:


Steffengreiner:
Closed since it should be a hotfix from main see #1156

PR #1159, opened 2025-05-27
Enable optimistic locking for experiments

sven1103:
To deal with race conditions, this PR introduces a version field
for the experiment object to enable optimistic locking strategy
on the database layer.


Steffengreiner (approved):
Well done :)

PR #1183, opened 2025-06-17
Add AppDialog, parse and validate

sven1103:


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+      Context context,
+      MetadataConverterV2<? extends ValidationRequestBody> converter,
+      MessageSourceNotificationFactory notificationFactory) {
+    this.uploadBuffer = new EditableMultiFileMemoryBuffer();
+    this.measurementFileItems = new ArrayList<>();
+    this.service = requireNonNull(service);
+    this.context = requireNonNull(context);
+    this.validationProgress = new Div();
+    this.converter = requireNonNull(converter);
+    this.notificationFactory = requireNonNull(notificationFactory);
```
KochTobi, line 102: NIT: I find this section hard to read. Mainly because it is unclear where the method parameters are assigned. I suggest reordering the statements to first have the assignments like `this.service = requireNonNull(service)` and then having the internal fields initialized e.g. `this.validationProgress = new Div()`
This might be a personal preference so feel free to ignore it if you do not agree.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+    add(validationProgress);
+  }
+
+  private void refreshFileItemDisplay() {
+    if (measurementFileItems.isEmpty()) {
+      uploadItemsDisplay.hide();
+      return;
+    }
+    uploadItemsDisplay.show();
+  }
```
KochTobi, line 122:
```suggestion
  private void refreshFileItemDisplayVisibility() {
    if (measurementFileItems.isEmpty()) {
      uploadItemsDisplay.hide();
      return;
    }
    uploadItemsDisplay.show();
  }
```
You are not modifying the item display here in the sense that you add or remove nodes. Maybe adding the hint to the method name expresses your intent for this function more clearly?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+
+  private void removeFile(String fileName) {
+    uploadBuffer.remove(fileName);
+    measurementFileItems.removeIf(
+        measurementFileItem -> measurementFileItem.fileName().equals(fileName));
```
KochTobi, line 127: 👍

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+    uploadItemsDisplay.removeFileFromDisplay(fileName);
+    refreshFileItemDisplay();
+  }
+
+  private void onFileRemoved(FileRemovedEvent fileRemovedEvent) {
+    removeFile(fileRemovedEvent.getFileName());
+  }
+
+  private void onUploadFailed(FailedEvent failedEvent) {
+    displayError();
```
KochTobi, line 137: NIT: do we want to log something to the console here? Otherwise it will be impossible to help users figure out what went wrong? Might be out of scope and not needed?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  private void displayError(String message) {
+    getUI().ifPresent(ui -> ui.access(() -> {
+      var toast = notificationFactory.toast("measurement.upload.failed.reason",
```
KochTobi, line 142: I think we could list the reasons in the message property file instead. `measurement.upload.failed.unsupported-file-type` or similar. Otherwise we have the error message in the code again and do not gain from the message source here.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+    getUI().ifPresent(ui -> ui.access(() -> {
+      var measurementFileItem = new MeasurementFileItem(fileName,
+          new MeasurementValidationReport(totalValidations, combinedResult));
+      addAndDisplayFile(measurementFileItem);
+    }));
+  }
+
+  private void addAndDisplayFile(MeasurementFileItem measurementFileItem) {
+    measurementFileItems.add(measurementFileItem);
+    showFile(measurementFileItem);
```
KochTobi, line 231: why did you decide to separate this into two methods?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+        .filter(ValidationResult::containsFailures).findAny();
+    if (inputWithFailureSearch.isEmpty()) {
+      return InputValidation.passed();
+    }
+    return InputValidation.failed();
+  }
+
+  @Override
+  public boolean hasChanges() {
+    return !measurementFileItems.isEmpty();
```
KochTobi, line 267: 👍 nice and simple

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+      add(uploadSection, uploadedItemsSection);
+      addClassName("upload-items-display");
+    }
+
+    private void addFileToDisplay(MeasurementFileDisplay measurementFileDisplay) {
+      uploadedItemsDisplays.add(measurementFileDisplay);
+    }
+
+    private void removeFileFromDisplay(String fileName) {
+      MeasurementFileDisplay[] fileDisplays = fileDisplaysWithFileName(fileName);
```
KochTobi, line 312: can there be multiple with the same filename?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+      uploadedItemsDisplays.setVisible(true);
+    }
+  }
+
+  record MeasurementValidationReport(int validatedRows,
+                                     ValidationResult validationResult) {
+
+  }
+
+  public record MeasurementFileItem(String fileName,
```
KochTobi, line 339: See my comment above, if there are multiple with the same filename, they must differ in the validation report. Otherwise there will be display problems as presumably vaadin uses the equals and hash code methods?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -0,0 +1,424 @@
...
+      displayBox.addClassName("validation-display-box");
+      add(displayBox);
+      addClassName("measurement-item");
+    }
+
+    public MeasurementFileItem measurementFileItem() {
+      return measurementFileItem;
+    }
+
+    private void createDisplayBox(MeasurementValidationReport measurementValidationReport) {
```
KochTobi, line 374: Is this creating a display box? 
```suggestion
    private void setDisplayBoxContent(MeasurementValidationReport measurementValidationReport) {
```
```suggestion
    private void fillDisplayBox(MeasurementValidationReport measurementValidationReport) {
```

KochTobi (changes requested):
Very nice code. Some minor suggestions and remaks

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
sven1103: hm, the i think refreshing a display is already descriptive. You do not refresh a visibility :shrug:

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
sven1103: legacy code from the upload dialog. But i see that it does not make sense to offer the showFile method explicitly

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
sven1103: technically yes

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
sven1103: I think this is out of scope for this work. However,  if you think this causes an actual issue for the users, please feel free and create an own issue.

KochTobi (approved):

PR #1239, opened 2025-08-25
Introduce eventual consistent project modification date update

sven1103:
Handles optimistic locking exceptions in case of massive concurrent project resource access.

### DIFF
--- a/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
+++ b/project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
@@ -5,8 +5,6 @@ import static java.util.function.Predicate.not;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.time.temporal.Temporal;
-import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
```diff
@@ -205,6 +210,35 @@ public void removeFunding(ProjectId projectId) {
...
+        Thread.sleep(calcBase2Duration(attempt));
+      } catch (InterruptedException e) {
+        log.error("Interrupted while updating modified date for project " + projectID);
+        // We try one last time
+        tryToUpdateModifiedDate(projectID, modifiedOn);
+        Thread.currentThread().interrupt();
+      }
+      attempt++;
+    }
+  }
```
KochTobi, line 234: As this is only updating the modified date, maybe add a break to the while loop in case the current date is already later then the modifiedOn date?

KochTobi (approved):
lgtm

project-management/src/main/java/life/qbic/projectmanagement/application/ProjectInformationService.java
```diff
@@ -204,7 +207,54 @@ public void removeFunding(ProjectId projectId) {
...
+        return;
+      } catch (OptimisticEntityLockException e) {
+        log.debug("Optimistic lock exception occurred while updating modified date for project " + projectID);
+      }
+      try {
+        Thread.sleep(calcBase2Duration(attempt));
+      } catch (InterruptedException e) {
+        log.error("Interrupted while updating modified date for project " + projectID);
+        // We try one last time
+        tryToUpdateModifiedDate(project, modifiedOn);
```
KochTobi, line 228: Why do we try one last time? What happens if an OptimisticEntityLockException is thrown here? The thread interrruption is then ignored?

sven1103: No, if the Thread is interrupted and the task has not yet completed we must try again. Otherwise it will get lost.

PR #1278, opened 2025-09-16
Fixes the orcid icon size in user management

KochTobi:


Steffengreiner (dismissed):
Heya @KochTobi, Thank you for the quick fix. I checked the figma prototype and would recommend the usage of the `icon-size-m` class to adhere to the sizes specified there. (See screenshot below)
Screenshot 2025-09-16 at 09 50 35

PR #1296, opened 2025-09-18
Fix missing measurement name

KochTobi:
Cleans up unused code a bit and makes sure the measurement name is propagated.

Previously the measurement name was read from the user input but never reached a persisted state. This PR propagates the measurement name to the domain so that it is persisted.


sven1103 (dismissed):

Steffengreiner (changes requested):
Heya @KochTobi, Thank you for your contribution and the code looks good. 
However this PR misses the propagation of changes in the measurement name for both the pxp and ngs update templates. 
I'd approve it once this is included 👍

KochTobi:
> Heya @KochTobi, Thank you for your contribution and the code looks good. However this PR misses the propagation of changes in the measurement name for both the pxp and ngs update templates. I'd approve it once this is included 👍

I thought it does? The underlying model is the same. I will have another look.

KochTobi:
> > Heya @KochTobi, Thank you for your contribution and the code looks good. However this PR misses the propagation of changes in the measurement name for both the pxp and ngs update templates. I'd approve it once this is included 👍
> 
> I thought it does? The underlying model is the same. I will have another look.

You were right :+1:

sven1103 (approved):

PR #1307, opened 2025-09-24
Introduce signature check of dependencies

sven1103:
Introduces the maven pgp verify plugin and activates it for every module that has a src/main directory. 

Builds will now fail, if the signature does not validate and there is no rule specified for it in the pgp-keys.override.list.

To skip manual curation of known maven artifacts, we use a pgp-keys-map from org.simplify4u.


pgp-keys-override.list
```diff
@@ -0,0 +1,230 @@
+# syntax:
+# groupId:artifactId:version =  *,any,badSig,noKey,noSig
+# wildcards * are allowed in artifactId/version
+
+# We have not yet figured out from where this is pulled :(
+io.smallrye:jandex:3.2.0 = badSig
+io.smallrye:jandex:3.2.0:pom  = badSig
```
KochTobi, line 7: What do you mean by that comment?
https://mvnrepository.com/artifact/io.smallrye/jandex/3.2.0 <- this is the maven repository related to the artifact. The artifact is pulled by hibernate-core https://mvnrepository.com/artifact/io.smallrye/jandex/usages

pgp-keys-override.list
```diff
@@ -0,0 +1,230 @@
+# syntax:
+# groupId:artifactId:version =  *,any,badSig,noKey,noSig
```
KochTobi, line 2: what does 
noSig, noKey, badSig mean?
Maybe a comment could help?
`noSig` we don't need a signature

pgp-keys-override.list
```diff
@@ -0,0 +1,230 @@
...
+io.smallrye:jandex:3.2.0:pom  = badSig
+
+# Allow your reactor modules (produced in this build)
+life.qbic:*:* = noSig
+life.qbic.logging:*:* = noSig
+life.qbic.datamanager:*:* = noSig
+
+# Unsigned runtime dep you must keep for now
+org.jboss.logging:jboss-logging:3.6.1.Final = noSig
+
```
KochTobi, line 16:
```suggestion
```

pgp-keys-override.list
```diff
@@ -0,0 +1,230 @@
...
+# AspectJ
+org.aspectj:*:* = 0xDB5D8331654BFCC24475252F30C4AFF0AF1DADE1
+
+# Hibernate ORM
+org.hibernate.orm:*:* = 0x1452F35849B50750F6A3BBB4B54011358B352F85
+
+# --- Apache ActiveMQ Artemis (many modules in your log) ---
+# One key signs all the 2.40.x/2.41.x artifacts you hit
+org.apache.activemq:*:*       = 0x28417C95E8906D108392822354A43F3254868410
```
KochTobi, line 97:
```suggestion
org.apache.activemq:*:2.4*       = 0x28417C95E8906D108392822354A43F3254868410
```
According to your comment?

KochTobi:
Sadly the tests fail :/

pom.xml
```diff
@@ -176,7 +176,11 @@
         <artifactId>jackson-annotations</artifactId>
         <version>${jackson.version}</version>
       </dependency>
-
+          <dependency>
+            <groupId>com.apicatalog</groupId>
+            <artifactId>titanium-json-ld</artifactId>
+            <version>1.6.0</version>
+          </dependency>
```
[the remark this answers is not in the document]
sven1103: Not added but overrides the outdated and pgp key revoked included version (1.3.x)

pom.xml
```diff
@@ -259,8 +263,81 @@
...
+              </execution>
+            </executions>
+            <configuration>
+              <!-- verify build plugins too (compiler, surefire, etc.) -->
+              <verifyPlugins>true</verifyPlugins>
+              <verifyPluginDependencies>true</verifyPluginDependencies>
+              <!-- include provided deps if you want to check servlet APIs etc. -->
+              <verifyProvidedDependencies>false</verifyProvidedDependencies>
+              <!-- optionally check SNAPSHOTs as well -->
+              <verifySnapshots>false</verifySnapshots>
```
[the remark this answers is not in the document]
sven1103: I like to have it in to be explicit :)

pom.xml
```diff
@@ -176,7 +176,11 @@
         <artifactId>jackson-annotations</artifactId>
         <version>${jackson.version}</version>
       </dependency>
-
+          <dependency>
+            <groupId>com.apicatalog</groupId>
+            <artifactId>titanium-json-ld</artifactId>
+            <version>1.6.0</version>
```
KochTobi, line 182:
```suggestion
<!-- Use specific version to avoid revoked outdated version-->
            <version>1.6.0</version>
```

KochTobi: https://github.com/qbicsoftware/data-manager-app/pull/1307/files#r2374972141 maybe we add a small comment here for that

pom.xml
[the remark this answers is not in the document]
KochTobi: Then we should include the other configuration properties as well. Why do you want to be specific about the snapshot and provided dependency verification?

project-management-infrastructure/pom.xml
```diff
@@ -147,5 +147,13 @@
       <artifactId>spring-boot-autoconfigure</artifactId>
     </dependency>
   </dependencies>
-
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.apicatalog</groupId>
+        <artifactId>titanium-json-ld</artifactId>
+        <version>1.6.0</version>
```
KochTobi, line 155:
```suggestion
<!-- Use specific version to avoid revoked outdated version-->
        <version>1.6.0</version>
```

PR #1319, opened 2025-10-02
Fix UI freeze during measurement validation and registration

sven1103:
Removes some smelling code that can produce race conditions for the UI update. Also passes now the current UI instance as reference for background jobs with reactor. Ui getters are not thread-safe and are not guaranteed to return the intended UI reference. According to Vaadin's documentation, it is better to fetch the UI reference in the main thread and pass it to background threads explicitly.

Also, this PR enforces the server to push changes to the UI, since with automatic polling we see that the UI is not always updated despite this should be the expected behaviour when the user session lock is released. This resulted in ui components not being updated from background tasks (validation, registration pending task toast, etc) and the update was waiting for a ui interaction from the user.


user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -258,54 +260,76 @@ private void runValidation(ArrayList<ValidationRequest> requests, int itemsToVal
...
+    setValidationProgressText(ui, "Starting validation ...");
     isValidationInProgress.set(true);
   }
 
-  private void validationFinished() {
-    hideValidationProgress();
+  private void validationFinished(UI ui) {
+    if (ui == null || !ui.isAttached()) {
+      return;
+    }
```
KochTobi, line 292: Here you risk returning before resetting the `validationInProgress` field.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -258,54 +260,76 @@ private void runValidation(ArrayList<ValidationRequest> requests, int itemsToVal
...
 
 
-  private void hideValidationProgress() {
-    getUI().ifPresent(ui -> ui.access(() -> validationProgress.setVisible(false)));
+  private void hideValidationProgress(@NonNull UI ui) {
+    if (ui.isAttached() && ui.getSession().hasLock()) {
+      validationProgress.setVisible(false);
+    } else {
+      ui.access(() -> validationProgress.setVisible(false));
+    }
```
KochTobi, line 303: nice :+1:

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -258,54 +260,76 @@ private void runValidation(ArrayList<ValidationRequest> requests, int itemsToVal
...
   }
 
-  private void showValidationProgress() {
-    getUI().ifPresent(ui -> ui.access(() -> validationProgress.setVisible(true)));
+  private void showValidationProgress(@NonNull UI ui) {
+    if (ui.isAttached() && ui.getSession().hasLock()) {
+      validationProgress.setVisible(true);
+    } else {
+      ui.access(() -> validationProgress.setVisible(true));
+    }
```
KochTobi, line 311: Nice but is `validationProgess` already populated with the correct information here?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -258,54 +260,76 @@ private void runValidation(ArrayList<ValidationRequest> requests, int itemsToVal
...
         .reduce(result, ValidationResult::combine);
-
-    getUI().ifPresent(ui -> ui.access(() -> {
-      var measurementFileItem = new MeasurementFileItem(fileName,
-          new MeasurementValidationReport(totalValidations, combinedResult));
-      addAndDisplayFile(measurementFileItem);
-    }));
+    var measurementFileItem = new MeasurementFileItem(fileName,
+        new MeasurementValidationReport(totalValidations, combinedResult));
+    addAndDisplayFile(measurementFileItem);
```
KochTobi, line 332: You removed the ui access here and introduced an ui.push in the calling code.
This is different from `validationFinished(UI ui)`. As these things always occur together wouldn't it make sense to handle them the same way?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMain.java
```diff
@@ -418,33 +403,62 @@ private void handleDeletionResults(Result<Void, MeasurementDeletionException> re
...
+
     asyncService.measurementUpdatePxP(projectId.value(), selectedMeasurementIds, OPEN_XML)
-        .doOnSubscribe(ignore -> executeInUi(inProgressToast::open))
-        .doOnSuccess(this::triggerDownload)
-        .doFinally(it -> executeInUi(inProgressToast::close))
-        .subscribe();
+        .subscribe(result -> ui.access(() -> {
+          inProgressToast.close();
+          ui.push();
+          triggerDownload(result, ui);
```
KochTobi, line 425: Why do you force a ui update before the download starts?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMain.java
```diff
@@ -418,33 +403,62 @@ private void handleDeletionResults(Result<Void, MeasurementDeletionException> re
...
+    }
     ProjectId projectId = context.projectId().orElseThrow();
     var inProgressToast = messageFactory.pendingTaskToast("measurement.preparing-download",
         MessageSourceNotificationFactory.EMPTY_PARAMETERS, getLocale());
+
+    if (ui.getSession() != null && ui.getSession().hasLock()) {
+      inProgressToast.open();
+    } else {
+      ui.access(inProgressToast::open);
+    }
```
KochTobi, line 419: you show the inprogress toast efore the async service started work. The reactor publisher is not active at this point in time. 
Is this intended?

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/MeasurementMain.java
```diff
@@ -418,33 +403,62 @@ private void handleDeletionResults(Result<Void, MeasurementDeletionException> re
...
+
+    asyncService.measurementUpdateNGS(projectId.value(), selectedMeasurementIds, OPEN_XML)
+        .subscribe(result -> ui.access(() -> {
+          inProgressToast.close();
+          ui.push();
```
KochTobi, line 450: why do you enforce a ui push before the download is triggered?
For reactor flux or mono, vaadin recommends using UI#accessLater https://vaadin.com/docs/latest/building-apps/deep-dives/presentation-layer/server-push/reactive

KochTobi:
I am not sure I understand when you chose which approach to update the UI.
As far as I can see there seems to be the direct modification in case the session lock belongs to the current thread and the ui is presently attached, the ui.access optionally combined with ui.push otherwise and sometimes a ui.access without any ui.push
I am also not clear on when you decide to use the ui.access in a calling code and when the ui is incorporated as a method parameter for the method to access the ui.

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -258,22 +258,27 @@ private void runValidation(ArrayList<ValidationRequest> requests, int itemsToVal
...
     service.validate(Flux.fromIterable(requests))
-        .doFirst(this::validationStarted)
         .doOnNext(ignored -> counter.incrementAndGet())
         .bufferTimeout(20, Duration.ofMillis(200))
-        .doOnNext(item -> setValidationProgressText(
-            "Processed " + counter.get() + " requests from " + itemsToValidate))
+        .doOnNext(item -> UiUtil.onUI(ui, () -> {
+          setValidationProgressText(
+              "Processed " + counter.get() + " requests from " + itemsToValidate);
+          ui.push(); // Ensure client is informed
```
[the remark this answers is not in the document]
sven1103: yes!

user-interface/src/main/java/life/qbic/datamanager/views/projects/project/measurements/registration/MeasurementUpload.java
```diff
@@ -74,6 +77,10 @@ public class MeasurementUpload extends Div implements UserInput {
 
   private final Map<String, List<? extends ValidationRequestBody>> validationRequestsPerFile = new HashMap<>();
   private final MessageSourceNotificationFactory notificationFactory;
+  private final UI ui = UI.getCurrent();  // keeps a reference to the current UI
```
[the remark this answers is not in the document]
sven1103: No, the attachment is not a pre-condition. The Utility class checks this also explicitly.

KochTobi (dismissed):

PR #1323, opened 2025-10-06
Introduce FilterGrid as a reusable component

sven1103:
Introduces a Filter Grid component that is build on top of the MultiSelectLazyLoadingGrid and provides features like action button groups, multi-type tab sheet support and automatic generation of show/hide column options.


Shraddha0903 (commented):
Changes requested have been made. Looks good to me 👌

PR #1230, opened 2025-08-08
Add Client ID for ROR repository

KochTobi:
Due to a change coming december, the ROR repository requires the addition of a client id to the http request. This PR introduces these changes. It does not adapt the data manager to being able to use the `v2` api of the ROR repository. For more information see https://ror.readme.io/docs/client-id.

PR #1359, opened 2025-12-10
Add client ID to ROR request and switch to APIv2

KochTobi:
Replaces https://github.com/qbicsoftware/data-manager-app/pull/1230

### DIFF
--- a/datamanager-app/src/main/java/life/qbic/datamanager/AppConfig.java
+++ b/datamanager-app/src/main/java/life/qbic/datamanager/AppConfig.java
@@ -122,8 +122,7 @@ public class AppConfig {
       @Value("${qbic.external-service.organisation-search.ror.organisation-api-endpoint}") String rorOrganisationEndpoint) {
     Objects.requireNonNull(rorOrganisationEndpoint);
     Objects.requireNonNull(clientId);
-    RorApiV2 rorApiV2 = new RorApiV2(rorOrganisationEndpoint, clientId);
-    return rorApiV2;
+    return new RorApiV2(rorOrganisationEndpoint, clientId);
   }
 
   @Bean


project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/organisations/CachedOrganisationRepository.java
```diff
@@ -81,27 +103,45 @@ private Optional<Organisation> lookupROR(String iri) {
...
+
   private Organisation findOrganisationInROR(String rorId) {
     try {
       HttpClient client = HttpClient.newBuilder().version(Version.HTTP_2)
           .followRedirects(Redirect.NORMAL).connectTimeout(
               Duration.ofSeconds(10)).build();
-      HttpRequest rorQuery = HttpRequest.newBuilder().uri(URI.create(ROR_API_URL.formatted(rorId)))
-          .header("Content-Type", "application/json").GET().build();
+
+      var queryBuilder = HttpRequest.newBuilder().uri(URI.create(ROR_API_URL.formatted(rorId)))
```
sven1103: This will lead to runtime exceptions, since the ROR_API_URL can be an invalid URL. I suggest to validate the URI and make it configurable-

sven1103 (changes requested):
Some minor but imo still notable requests

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/organisations/CachedOrganisationRepository.java
```diff
@@ -40,23 +41,44 @@ public class CachedOrganisationRepository implements OrganisationRepository {
...
+  /**
+   * Creates a Repository without an API Key. Not advised for use in production.
+   */
+  protected CachedOrganisationRepository(int cacheSize) {
+    log.warn(
+        "Creating organisation repository without API Key. This is not advised to be used in production.");
+    apiClientId = null;
+    this.configuredCacheSize = cacheSize;
+  }
+
```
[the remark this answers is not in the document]
KochTobi: I could not figure out how to adapt the tests to have a configurable API key there. As they connect to the ROR repository, we need one in the tests as well or need a way to create the CachedOrganisationRepository without an api key. That is also why the constructors are protected.

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/organisations/CachedOrganisationRepository.java
KochTobi: URI.create will fail for invalid URLs. I do not get your point here?

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/organisations/RorApi.java
```diff
@@ -0,0 +1,152 @@
...
+          return Optional.empty();
+        }
+        RorEntry rorEntry = new ObjectMapper().configure(
+                DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .readValue(result.body(), RorEntryV2.class);
+        return Optional.of(rorEntry);
+      } catch (IOException | InterruptedException e) {
+        log.error("Finding ROR entry failed for organisation: %s".formatted(rorId), e);
+        /* Clean up whatever needs to be handled before interrupting  */
+        Thread.currentThread().interrupt();
```
sven1103: Only propagate the thread interrupten when an interrupted exception is thrown. Otherwise you signal the current thread to stop on IOExceptions (e.g. network) which can cause nasty side effects in our application. 

So either you have two separate catch blocks here or you check the `Thread.current().isInterrupted()` flag.

project-management-infrastructure/src/main/java/life/qbic/projectmanagement/infrastructure/organisations/RorApi.java
```diff
@@ -0,0 +1,252 @@
...
+        result = client.send(request, BodyHandlers.ofString());
+        //If a valid RoRId was provided but the ID does not exist we fail
+      } catch (IOException e) {
+        throw new RorRequestException(e);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new RorRequestException(e);
+      }
+
+      if (result.statusCode() == HttpStatus.NOT_FOUND_404) {
```
sven1103: int in comparison vs an enum?

KochTobi: That is not an enum but a static int value 😄

Issue #1364, opened 2026-01-07
Proteomics Validation does not check the injection volume directly after download


KochTobi:
### What happened?

When providing an invalid injection volume, valid being anything but whole numbers at the moment, the dialog shows successful validation. After clicking on register measurement, the process fails.
This is can be overlooked.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1365, opened 2026-01-07
Add injection volume validation

KochTobi:
Solves #1364
Previously:
image

With this PR:
Screenshot 2026-01-07 at 14 36 43


sven1103 (approved):
Thank you @KochTobi to solve it so quickly.

PR #1373, opened 2026-02-03
Inform user about duplicate condition during input

sven1103:
Given the user enters an condition that is already defined in another experimental group of the same experiment, 
the app informs the user now about it, highlights which on is duplicate, and prevents them from submitting the task.

### DIFF
--- a/datamanager-app/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupInput.java
+++ b/datamanager-app/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupInput.java
@@ -12,7 +12,6 @@ import com.vaadin.flow.component.icon.VaadinIcon;
 import com.vaadin.flow.component.textfield.NumberField;
 import com.vaadin.flow.component.textfield.TextField;
 import com.vaadin.flow.data.binder.Binder;
-import com.vaadin.flow.shared.Registration;
 import jakarta.validation.constraints.Min;
 import java.io.Serial;
 import java.util.ArrayList;
@@ -25,7 +24,6 @@ import java.util.stream.Collectors;
 import life.qbic.datamanager.views.general.Container;
 import life.qbic.datamanager.views.projects.project.experiments.experiment.ExperimentalGroupInput.ExperimentalGroupBean;
 import life.qbic.projectmanagement.application.VariableValueFormatter;
-import life.qbic.projectmanagement.domain.model.experiment.Condition;
 import life.qbic.projectmanagement.domain.model.experiment.Experiment;
 import life.qbic.projectmanagement.domain.model.experiment.ExperimentalGroup;
 import life.qbic.projectmanagement.domain.model.experiment.VariableName;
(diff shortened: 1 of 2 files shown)


datamanager-app/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupInput.java
```diff
@@ -31,10 +32,62 @@
...
+ *     automatically deselected (see {@link #overwriteSelectionOfSameVariable(MultiSelectComboBox)}).
+ *   </li>
+ *   <li>
+ *     <b>Completeness validation:</b> The component derives the number of distinct variables from
+ *     {@code availableLevels} ({@link #variableCount}) and validates that the selection size equals
+ *     that count. This expresses the requirement "please select one level for each variable".
+ *   </li>
+ *   <li>
+ *     <b>Model synchronization:</b> Any change to the name, condition selection, or replicate count
+ *     triggers {@link #updateValue()} and emits an {@link InputChangedEvent}. This makes it easy for
```
KochTobi: Not relevant here what method is called. Please remove that detail. 
When the user input changes, the component changes its value and a `InputChangeEvent` is thrown. This is all you need to say here. No need for the overwhelming details. 
The detailed ways this is done can go to the public methods exposed but do not belong in the class Javadoc imho.

datamanager-app/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/ExperimentalGroupInput.java
```diff
@@ -48,20 +101,69 @@ public class ExperimentalGroupInput extends CustomField<ExperimentalGroupBean> {
...
+
   private final MultiSelectComboBox<VariableLevel> variableLevelSelect;
   private final NumberField replicateCountField;
+
+  /**
+   * Number of distinct experimental variables inferred from the provided available levels.
+   * <p>
+   * Used to validate that the user selected exactly one level per variable.
+   * </p>
+   */
```
KochTobi: probably this can be private and the javadoc removed.

KochTobi (changes requested):
Works great!
Some concerns regarding the javadoc and code.

datamanager-app/src/main/java/life/qbic/datamanager/views/projects/project/experiments/experiment/components/ExperimentalGroupsDialog.java
```diff
@@ -214,34 +319,153 @@ public boolean isValid() {
...
+   *
+   * @param level a variable level (UI representation)
+   * @return stable key
+   */
+  private String variableLevelKey(VariableLevel level) {
+    return new StringJoiner("::")
+        .add(Objects.toString(level.variableName(), ""))
+        .add(Objects.toString(level.value(), ""))
+        .add(Objects.toString(level.unit(), ""))
+        .toString();
```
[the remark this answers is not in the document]
sven1103: It is null safe :)

PR #1400, opened 2026-03-19
Update the bug report template

KochTobi:
Currently there is no clear capture of expected behaviour and observed behaviour. It is hard to understand and distinguish what went wrong and what was expected.

### DIFF
--- a/.github/ISSUE_TEMPLATE/02-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/02-bug-report.yml
@@ -75,7 +75,7 @@ body:
         - Safari
         - Microsoft Edge
         - Not browser-specific / backend issue
-        -
+        
   - type: textarea
     id: additional-context
     attributes:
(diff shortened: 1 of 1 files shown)


sven1103 (changes requested):
Nice improvement splitting "What actually happened?" vs "What did you expect to happen?"; much clearer for triage.

One fix needed before merge: in .github/ISSUE_TEMPLATE/02-bug-report.yml, the browsers dropdown options include a blank entry ('-'). This will render as an empty selectable option in the GitHub UI; please remove it.

KochTobi:
> Nice improvement splitting "What actually happened?" vs "What did you expect to happen?"; much clearer for triage.
> 
> One fix needed before merge: in .github/ISSUE_TEMPLATE/02-bug-report.yml, the browsers dropdown options include a blank entry ('-'). This will render as an empty selectable option in the GitHub UI; please remove it.

good spotting :+1: I changed it so feel free to have another look

sven1103 (approved):
Confirmed: empty browser option removed from .github/ISSUE_TEMPLATE/02-bug-report.yml. Looks good to merge.

PR #1423, opened 2026-04-21
Fix sprint date parsing in label workflow

### LINKED ISSUE #1420: Create GH action to regularly check and label issues that are part of the current sprint
In order to easier checkout and track issues associated with a sprint, it would help if the issues are labeled with sth like spring:current so developers can filter for these issues when working during a sprint. 

Issues that carry the label but are not part of the current sprint shall not carry the sprint label anymore, and issues missing the label but are part of the current sprint, shall have it.

The action shall check regularly, at least twice a day to capture updates

sven1103:
## Problem
The workflow was not detecting the current sprint because GitHub Projects API returns dates in ISO 8601 format (`2026-04-21T00:00:00Z`), but the script was comparing with simple date strings (`2026-04-21`).

## Solution
- Extract date portion using `split("T")[0]` before comparison
- Handle both `.duration.endDate` and `.endDate` field names (fallback)
- Add debug output to show raw iteration data in logs

## Changes
| File | Change |
|------|--------|
| `.github/workflows/label_sprint_issues.yml` | Fixed jq date parsing logic |

## Testing
After merging, trigger the workflow manually via Actions tab to verify sprint detection works.

Fixes #1420

### DIFF
--- a/.github/workflows/label_sprint_issues.yml
+++ b/.github/workflows/label_sprint_issues.yml
@@ -35,7 +35,7 @@ jobs:
           echo "Ensuring label '$LABEL_NAME' exists..."
           gh label create "$LABEL_NAME" --color "0E8A16" --description "Issue is part of the current sprint" 2>/dev/null || echo "Label already exists"
           
-          # Get all project items with their iteration info
+          # Get all project items
           echo "Fetching project items from project $PROJECT_NUMBER..."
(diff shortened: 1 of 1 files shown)


KochTobi (approved):
LGTM.
This PR not only changes the date parsing but removes the counting of labeled issues in the sprint. Is this intentional?

Issue #1410, opened 2026-03-23
[Task] Sub-Issue 2: Apply User-Triggered UI Sorting to exported files


Steffengreiner:
### What's the issue?

**Context**
When a user applies a custom sort to a grid in the UI (e.g. by clicking a column header), that sort order is currently not carried over into the exported files. 
This means the sheet the user receives may be ordered differently from what they were looking at in the UI, making it harder to work with the exported data. 
The expected behaviour is that the exported sheet always mirrors the current sort state of the grid at the time of export

**Current Behaviour**
When a user applies a custom column sort in the following UI elements:

1.) Sample grid, 
2.) Measurement grid
3.) Raw data grid 

and then downloads the metadata sheet, the downloaded sheet does not reflect the user's chosen sort order within the UI 

**Expected Behaviour**
When a user exports a metadata sheet after having applied a custom sort in the UI grid, 
the downloaded sheet reflects the sort order that was active in the grid at the time of the export.

**Acceptance Criteria**

- Given a user has sorted the sample grid by a custom column, when they download the edit metadata sheet or the full export sheet, then the rows in the downloaded sheet are ordered according to the active UI sort.
- Given a user has sorted the measurement grid by a custom column, when they download the edit metadata sheet or the full export sheet, then the rows in the downloaded sheet are ordered according to the active UI sort.
- Given a user has sorted the raw data grid by a custom column, when they download the Download URL List, then the URLs in the downloaded text file are ordered according to the active UI sort.
- Given a user has not applied any custom sort, when they download a sheet, then the default ascending sort by ID (as defined in sub-issue 1) is applied — this behaviour is unchanged.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #1409, opened 2026-03-23
[Task] Sub-Issue 1: Establish a default sorting order between sample, measurement and raw data entities


Steffengreiner:
### What's the issue?

**Context**
Currently, the sample, measurement, and raw data grids in the UI and the exported sheets have no consistent default sort order. 
This causes rows to be misaligned when users copy or compare data across sheets, making cross-referencing error-prone.

**Current Behaviour**
The following UI elements and exported files have no meaningful default sort order: 

- The sample grid 
- The measurement grid for both proteomics and genomics
- The raw data grid
- Downloaded sample and measurement edit templates
- Downloaded sample and measurement export sheets
- Downloaded URL List for raw data download

**Expected Behaviour**
All grids and their corresponding downloaded sheets share a consistent user understandable default sorting order. 

Examples include either: 
1.) Ascending by their respective ID (sample ID for samples, measurement ID for measurements and raw data).
2.) FIFO Principle (Store the row numbers for each of the user provided filled-in templates and sort by this row number) 

**Acceptance Criteria**

- The default sorting is consistent between the raw data grid, the measurement grid and the sample grid .
- The default sorting is consistent between the exported sample and measurement exports and edit template and in accordance with the downloaded URL List from the raw data grid.
- The default sorting is consistent between aforementioned grid elements and the exported files.
- Given the sample grid is loaded, when no custom sort has been applied by the user, then rows are sorted lexicographically ascending by sample ID or FIFO.
- Given the measurement grid is loaded, when no custom sort has been applied by the user, then rows are sorted lexicographically ascending  by measurement ID or FIFO
- Given the raw data grid is loaded, when no custom sort has been applied, then rows are sorted lexicographically ascending by the measurement ID or FIFO
- Given a user downloads the edit metadata sheet for samples or measurements then rows in the sheet are sorted ascending lexicographically by the respective ID or FIFO
- Given a user downloads the full export sheet for samples, measurements and raw data url the entries are sorted ascending lexicographically by the respective ID or FIFO

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Steffengreiner:
Default Sorting is now based on the ascending lexicographically permanent identifier:
https://github.com/qbicsoftware/data-manager-app/pull/1482

PR #1482, opened 2026-07-13
Introduce a default ascending lexicographical sorting order within exported and shown entities based on their persistent identifier

Steffengreiner:
## Description

Introduce Sorting of Samples, Measurements and Raw Data via an Ascending sort of the SampleId and MeasurementIds within the UI grids and the downloaded files (Excel Templates and URL Link List).
This PR introduces a natural order comparator for measurementId and Sample Ids to account for the issue of the default natural order comparator (01, 10, 02, 03...). 

### Linked Task Issue
#1410 
#1409 

### Justification (if applicable)
This issue is constantly referenced in internal stakeholder discussions. It's currently a lot of manual overhead to get a clean overview of the measurements samples and raw data. 
Manual sorting of the downloaded Entitites and within the UI takes time and is deleted once the user switches the view. 
Since there is currently no sorting mechanism applied, the easiest solution is to start a default sorting and adapt it if the need changes. 

### Modified Areas
- [x ] Application layer (`application/` services, use cases)
- [x ] UI / Views (`views/` components, Vaadin routes)

### Behavioral Changes
- [x] Yes, behavior changes: The current non-sensical and non-harmonized sorting is replaced with an internally coherent sorting structure based on the persistent identifiers of the specific entitites.  

## Pre-Submission Checklist

- [x] **Issue Linked:** PR references a Task issue (see "Linked Task Issue" section above)
- [x] **Requirements Listed: Requirements are documented within the linked Tasks
- [x]  **Other:


application-commons/src/test/groovy/life/qbic/application/commons/MeasurementCodeComparatorSpec.groovy
```diff
@@ -0,0 +1,147 @@
+package life.qbic.application.commons
+
+import spock.lang.Specification
+
+class MeasurementCodeComparatorSpec extends Specification {
```
KochTobi: I like that you added tests! Please add some tests for the API of the compare method.
> Compares its two arguments for order. Returns a negative integer, zero, or a positive integer as the first argument is less than, equal to, or greater than the second.
The implementor must ensure that signum(compare(x, y)) == -signum(compare(y, x)) for all x and y. (This implies that compare(x, y) must throw an exception if and only if compare(y, x) throws an exception.)
The implementor must also ensure that the relation is transitive: ((compare(x, y)>0) && (compare(y, z)>0)) implies compare(x, z)>0.
Finally, the implementor must ensure that compare(x, y)==0 implies that signum(compare(x, z))==signum(compare(y, z)) for all z.

KochTobi (changes requested):

KochTobi:
Context: @Steffengreiner and I discussed the increased weight of the identifiers. We agreed that some sorting is necessary for now.

KochTobi (approved):
Amazing

PR #1484, opened 2026-07-22
feat(toast): redesign toast notifications with dark mode design system

sven1103:
## Summary

Redesign the toast notification UI to align with the data-manager style guide and improve visual design, accessibility, and consistency.

## Bug fixes

- **Missing status icons**: extract `levelIcon` as a separate field so subsequent `withContent()` calls cannot accidentally destroy the previously set icon
- **Routing/info links invisible**: add CSS overrides for `RouterLink`/`Button` elements rendered inside the toast
- **Info icon missing glyph**: `lumo:info-circle` does not exist — switch to `vaadin:info` (symbol-only)
- **Icons rendered as colored blobs**: `vaadin:*-circle` icons have their own filled-circle path — switch to symbol-only variants `vaadin:check`, `vaadin:close`, `vaadin:info`
- **Pending task title rendered as raw HTML markup**: `withProgressBar()` used `Div(title)` — replace with wrapped `Html` element
- **Trailing whitespace on HTML toast messages**: replace `<div style='display:contents'>` with inline `<span>` wrapper

## Visual design

- Add 14 new CSS variables in `variables.css`: design tokens (`--color-primary`, `--color-error`, `--color-success`), dark-surface shade tokens, and toast-specific tokens (`--toast-background-color`, `--toast-link-color`, etc.)
- Link color increased from `#1676f3` to `#66A8FF` to meet WCAG AA contrast
- Font sizes use Lumo variables (`--lumo-font-size-*`) instead of hardcoded px
- Icon size reduced from 20×20 to 14×14 inside the 28×28 circle

## Layout

- Single-line approach with ellipsis-based text truncation via child-combinator selectors (only the text-bearing node has `overflow: hidden`; action buttons and close button are never clipped)
- Routing container uses flexbox with `flex: 1 1 0` so text shrinks and buttons stay at natural width
- Toast overlay max-width removed so width adapts to content

## Toast message catalog (`toast-notifications.properties`)

- Added `<key>.message.subtext` documentation comment and sample entries for progress toasts
- Added `measurement.registration.in-progress.message.subtext` ('This might take a few minutes')
- Cleaned up `project.updated.error.retry.message.text`: removed inline `<a>` link in favor of the routed-link toast flow (the routing toast now renders the link as a styled button)

## ComponentDemo (test route)

- Complete overhaul of the `ComponentDemo` demo page with a two-column layout showing all toast types side by side (basic success/info/error, pending-task progress, routing, action-button, error with routing) — makes it easier to visually verify every toast variant in one place
- New message keys added: `experiment.created.success`, `task.failed.action`, `measurement.completed.success`, `measurement.completed.routing`, `measurement.failed.retry`

## Refactoring

- `pendingTaskToast`: build progress layout directly instead of delegating to text-toast path
- Add `actionToast` factory method for error toasts with a `Try Again` button

## Files changed

| File | Description |
|---|---|
| `toast.css` | Full rewrite — design tokens, flex layout, truncation, icon styling |
| `variables.css` | 14 new design-token / toast-specific CSS variables |
| `Toast.java` | Icon extraction, `vaadin:\*` icons, `Html` wrappers for progress title/subtext |
| `MessageSourceNotificationFactory.java` | `<span>` wrapper for HTML content, new `actionToast` factory |
| `toast-notifications.properties` | Subtext docs, new subtext entries, cleaned up inline links in retry message |
| `ComponentDemo.java` | Two-column demo page exercising all toast variants |


datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -33,14 +39,18 @@
...
+
   static final boolean DEFAULT_CLOSE_ON_NAVIGATION = true;
   static final Duration DEFAULT_OPEN_DURATION = Duration.ofSeconds(5);
   private static final Position DEFAULT_POSITION = Position.BOTTOM_START;
   private final List<Registration> closeOnNavigationListeners = new ArrayList<>();
   private final Button closeButton;
 
   private Component content;
-
+  private Optional<Component> levelIcon = Optional.empty();
```
KochTobi: Please avoid `Optional` as fields. Rather take a nullable `Component` in this case. With `Optional` as field or method parameter you always increase complexity as the `Optional` itself may be `null`. I know that you are assigning an empty Optional here, but the field itself can be set to `null` by accident.
Instead of using it in the field, you can provide a private method, that wraps the field in an `Optional.ofNullable` so you can access it as `Optional` in your code. This way the returned `Optional` can never be `null`.

`Optional` also can cause problems with serialization (as it is not Serializable). For this reason a nullable field is preferred as well.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -33,14 +39,18 @@
...
+  private Optional<Component> actionButton = Optional.empty();
```
KochTobi: Please do not use Optional as field.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -33,14 +39,18 @@
...
+  private Optional<String> title = Optional.empty();
```
KochTobi: Please do not use Optional as field.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -33,14 +39,18 @@
...
+  private Optional<String> subtext = Optional.empty();
```
KochTobi: Please do not use Optional as field.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,126 @@ public final class Toast extends Notification {
...
-    add(closeButton);
+    closeButton = createCloseButton();
     closeOnNavigation(DEFAULT_CLOSE_ON_NAVIGATION);
   }
 
-  private static Button buttonClosing(Toast toast) {
-    var button = new Button(LumoIcon.CROSS.create());
+  private Button createCloseButton() {
+    var icon = new Icon();
+    icon.getElement().setAttribute("icon", "vaadin:close-small");
```
KochTobi, line 74: Why not use the vaadin icon class? The creation of an Icon with attribute seems complicated.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,126 @@ public final class Toast extends Notification {
...
+      case INFO -> {
+        var icon = new Icon();
+        icon.getElement().setAttribute("icon", "vaadin:info");
+        icon.getElement().getStyle().setColor("white");
+        icon.getElement().getStyle().setWidth("14px");
+        icon.getElement().getStyle().setHeight("14px");
+        yield icon;
+      }
+    };
+  }
```
KochTobi, line 122: Why not use a CSS class for this?  I think there are some error, info, ... classes already.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,126 @@ public final class Toast extends Notification {
...
+    return switch (level) {
+      case SUCCESS -> "toast-icon-success";
+      case ERROR, WARNING -> "toast-icon-error";
+      case INFO -> "toast-icon-info";
+    };
+  }
+
+  /**
+   * Sets a title for the toast (used in progress/indeterminate toasts).
+   * Title is displayed in 18px bold text above the main content.
```
KochTobi: I don't think we should use hard pixel sizes for text??? They are usually expressed in `em` or `rem` to appear the same on every display without depending on the resolution! We have some classes for text size already. Why not use them?

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -121,29 +224,15 @@ public void setDuration(Duration duration) {
...
-  Toast withRouting(String linkText, Class<? extends Component> navigationTarget,
-      RouteParameters routeParameters) {
-    return this.withContent(Toast.createRoutingContent(this.content, createRoutingComponent(
-        linkText, navigationTarget, routeParameters)));
-  }
-
-  /**
-   * Sets the content of the toast
+   * Sets the content of the toast.
+   * Content is displayed in 16px text with the toast's primary text color.
```
KochTobi: Why provide hard pixel values? I don't think this javadoc helps very much.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -121,29 +224,15 @@ public void setDuration(Duration duration) {
...
    *
    * @param content the content of the toast to set.
-   * @return
+   * @return the modified toast
    */
   Toast withContent(Component content) {
     requireNonNull(content, "content must not be null");
-    //we only want to remove the current content from its parent if the parent is not the new content.
-    //Otherwise the current content would be removed from the new content as well.
+    // Remove current content from parent if different
```
KochTobi, line 235: This is not what happens. As the JavaDoc stated before: the new content might be a parent to the existing content. If this is the case, we do not want to add the new content only to remove part of it again.
It does not compare the current content to the new content but checks if the current content is a child of the new content.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -153,9 +242,80 @@ Toast withContent(Component content) {
...
+    requireNonNull(progressBar, "progressBar must not be null");
+    var barContainer = new Div(progressBar);
+    barContainer.addClassName("progress-bar-container");
+
+    var verticalLayout = new Div();
+    verticalLayout.addClassName("progress-vertical");
+
+    if (title.isPresent()) {
+      var titleWrapper = new com.vaadin.flow.component.Html(
+          "<div class=\"toast-title\">" + title.get() + "</div>");
```
KochTobi: This should be `new Div("my text here")` or similar. No need for the `Html` Object.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -153,9 +242,80 @@ Toast withContent(Component content) {
...
+
+    verticalLayout.add(barContainer);
+
+    if (subtext.isPresent()) {
+      var subtextWrapper = new com.vaadin.flow.component.Html(
+          "<span class=\"toast-subtext\">" + subtext.get() + "</span>");
+      verticalLayout.add(subtextWrapper);
+    }
+
+    refreshWithProgress(verticalLayout);
```
KochTobi: refresh is maybe wrong here. Replace with Progress as this method removes everything existing. `withProgressBar` should state that somewhere that it cannot used in a fluent API as originally designed as it would remove everything put into the toast from before.

datamanager-app/frontend/themes/datamanager/components/variables.css
```diff
@@ -99,4 +99,28 @@ First some general property definitions
...
+
+  /* --- Design tokens (from style guide, --color-* family) --- */
+  --color-primary: #1676f3;
+  --color-error: #ff4238;
+  --color-warning: #fec901;
+  --color-success: #15c15d;
+  --color-text-primary: #1676f3;
+  --color-text-error: #f52419;
+  --color-text-warning: #f2c004;
+  --color-text-success: #08b344;
```
KochTobi: There are variables for this already. `--primary-color`,... Please use them instead of cluttering the variables css

datamanager-app/frontend/themes/datamanager/components/variables.css
```diff
@@ -99,4 +99,28 @@ First some general property definitions
...
+
+  /* --- Dark-surface shade tokens (from style guide) --- */
+  --color-shade-base: #192434;
+  --color-shade-90: rgba(24, 39, 58, 0.94);
+  --color-shade-50: rgba(28, 48, 74, 0.5);
+  --color-shade-10: rgba(26, 56, 96, 0.1);
+  --color-shade-5: rgba(25, 59, 103, 0.05);
```
KochTobi: There are variables for this already. Please remove the new variables and use the existing ones.

datamanager-app/frontend/themes/datamanager/components/variables.css
```diff
@@ -99,4 +99,28 @@ First some general property definitions
...
+
+  /* --- Toast dark-surface tokens --- */
+  --toast-background-color: #223348;
+  --toast-text-color: rgba(235, 244, 255, 0.9);
+  --toast-subtle-text-color: rgba(209, 227, 250, 0.6);
+  --toast-link-color: #66A8FF;
+  --toast-progress-track-color: rgba(173, 200, 235, 0.14);
```
KochTobi, line 108: I like that you have different variables for toasts. Please consider aligning the values to existing values. e.g. `--toast-link-color` to `hsl(214, 100%, 48%)` or `--toast-background-color` to `hsl(214, 40%, 16%)`
If they need to be slightly different they can stay the same

datamanager-app/frontend/themes/datamanager/components/toast.css
```diff
@@ -1,44 +1,232 @@
...
+ * Backgrounds set directly on ::part(overlay) — CSS custom properties
+ * do not reliably cascade through Vaadin's Shadow DOM parts.
+ **********/
+
+/* === Default surface (base level) === */
+.toast-notification::part(overlay) {
+  background: linear-gradient(0deg, var(--color-shade-5), var(--color-shade-5)), var(--toast-background-color);
+  border: 1px solid var(--shade-color-20pct);
+  border-radius: 8px;
+  box-shadow: 0px 3px 12px -1px var(--shade-color-30pct),
```
KochTobi, line 14: Please use existing variables

datamanager-app/frontend/themes/datamanager/components/toast.css
```diff
@@ -1,44 +1,232 @@
...
+}
+
+/* --- Icon container (28x28 colored circle) --- */
+
+.toast-icon {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 28px;
+  height: 28px;
```
KochTobi: Having a fixed icon size in px causes the icon to appear differntly sized in relation to the other content on different monitors. I suggest using percentage based on the toast if alignment with the toast layout is desired or `em` or `rem` for scaling with the text if alignment to the text is desired.

datamanager-app/frontend/themes/datamanager/components/toast.css
```diff
@@ -1,44 +1,232 @@
...
+  border-radius: 50%;
+  flex-shrink: 0;
+}
+
+/* Force the icon glyph inside the circle to white */
+.toast-icon vaadin-icon {
+  color: white !important;
+  width: 14px !important;
+  height: 14px !important;
```
KochTobi: There should be no need for `!important`. Please use `!important` sparingly, as it cannot be overwritten by CSS selector specificity.
https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity

This also overwrites the size you set with `.toast-icon` which might be an undesired side-effect.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/MessageSourceNotificationFactory.java
```diff
@@ -66,14 +66,15 @@ public Toast toast(String key, Object[] parameters, Locale locale) {
     String messageText = parseMessage(key, parameters, locale);
 
     Component content = switch (type) {
-      case HTML -> new Html("<div style=\"display:contents\">%s</div>".formatted(messageText));
+      case HTML -> new Html("<span>%s</span>".formatted(messageText));
```
KochTobi: Why do you want a span in the HTML tree? This breaks formatting if you provide `<p>` or similar in the message text. Please revert

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/MessageSourceNotificationFactory.java
```diff
@@ -115,27 +116,74 @@ public Toast routingToast(String key, Object[] messageArgs, Object[] routeArgs,
...
    * <p>
-   * The following message keys have to be present:
+   * The toast uses the progress layout (no icon) with:
    * <ul>
-   *   <li>{@code <key>.message.type}
-   *   <li>{@code <key>.message.text}
-   *   <li>{@code <key>.routing.link.text}
+   *   <li>Title (from message.text) in 18px bold</li>
+   *   <li>Indeterminate progress bar</li>
+   *   <li>Optional subtext (from message.subtext) in 16px regular</li>
```
KochTobi, line 123: I think you should keep the information on how the key is integrated in the message properties. This was included as it was unclear to developers.

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/MessageSourceNotificationFactory.java
```diff
@@ -163,7 +211,7 @@ public NotificationDialog dialog(String key, Object[] parameters, Locale locale)
     MessageType type = parseMessageType(key, locale);
     String messageText = parseMessage(key, parameters, locale);
     Component content = switch (type) {
-      case HTML -> new Html("<div style=\"display:contents\">%s</div>".formatted(messageText));
+      case HTML -> new Html("<span>%s</span>".formatted(messageText));
```
KochTobi: Same error as before. If you wrap in a span instead of an invisible (display content) div, you loose any formatting in the messageText for linebreaks. This makes multiline messages impossible

KochTobi (changes requested):
Hi Sven the new toasts look fancy. Please enable multiline messages again, use css classes and prevent duplicating css variables.

KochTobi:
Toast have different text sizes.

https://github.com/user-attachments/assets/d74e9b2f-1c95-4a96-94d7-789da7e2747b

datamanager-app/frontend/themes/datamanager/components/variables.css
sven1103: Please check the Style Guide and token names. This implementation honors our UX-designer's official style guide.

datamanager-app/frontend/themes/datamanager/components/variables.css
sven1103: https://github.com/qbicsoftware/data-manager-app/pull/1484/changes/BASE..f983a8f795919acda21bf6aad5ad7f3c2a2baed4#r3644922091

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,128 @@ public final class Toast extends Notification {
...
-    add(closeButton);
+    closeButton = createCloseButton();
     closeOnNavigation(DEFAULT_CLOSE_ON_NAVIGATION);
   }
 
-  private static Button buttonClosing(Toast toast) {
-    var button = new Button(LumoIcon.CROSS.create());
+  private Button createCloseButton() {
+    var icon = new Icon();
+    icon.getElement().setAttribute("icon", "vaadin:close-small");
```
KochTobi, line 74: why not have a vaadin icon here?

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,128 @@ public final class Toast extends Notification {
...
+  }
+
+  private Component getInfoForLevel(NotificationLevel level) {
+    return switch (level) {
+      case SUCCESS -> {
+        var icon = new Icon();
+        icon.getElement().setAttribute("icon", "vaadin:check");
+        icon.getElement().getStyle().setColor("white");
+        icon.getElement().getStyle().set("width", "var(--icon-size-xs)");
+        icon.getElement().getStyle().set("height", "var(--icon-size-xs)");
```
KochTobi: Please avoid the manual style overrides when we have css classes

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,128 @@ public final class Toast extends Notification {
...
+        yield icon;
+      }
+      case ERROR, WARNING -> {
+        var icon = new Icon();
+        icon.getElement().setAttribute("icon", "vaadin:close");
+        icon.getElement().getStyle().setColor("white");
+        icon.getElement().getStyle().set("width", "var(--icon-size-xs)");
+        icon.getElement().getStyle().set("height", "var(--icon-size-xs)");
```
KochTobi: please use the vaadin icon and css classes

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
```diff
@@ -54,33 +64,128 @@ public final class Toast extends Notification {
...
+        yield icon;
+      }
+      case INFO -> {
+        var icon = new Icon();
+        icon.getElement().setAttribute("icon", "vaadin:info");
+        icon.getElement().getStyle().setColor("white");
+        icon.getElement().getStyle().set("width", "var(--icon-size-xs)");
+        icon.getElement().getStyle().set("height", "var(--icon-size-xs)");
```
KochTobi: please use the vaadin icon and css classes

KochTobi (dismissed):
LGTM. However still some `getStyle().set(\`width\`, some-value)` exist. Please have a look if you can replace them with existing or speaking CSS classes. Regarding the icons, why not use the vaadin icon instead of manipulating the HTML

datamanager-app/frontend/themes/datamanager/components/toast.css
sven1103: applied

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/MessageSourceNotificationFactory.java
sven1103: resolved

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/MessageSourceNotificationFactory.java
sven1103: addressed

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
sven1103: addressed

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
sven1103: addressed

datamanager-app/src/main/java/life/qbic/datamanager/views/notifications/Toast.java
sven1103: i dont know what is to resolve here :shrug:

KochTobi: The comment is wrong. It does not remove the current content from the parent if the content is different. It removes the current content if the content is not a child of the new content.

KochTobi (dismissed):
Approved 😄 9,5/10 for a 10/10 please fix the ambivalent comment

KochTobi (dismissed):
lgtm

PR #1513, opened 2026-08-31
Fix jobrunr prefix and specify hibernate dialect for mariadb

Steffengreiner:
## Description

This PR addresses the attribute name change in the `application.properties` associated with jobrunr during the migration from v7 to v8.

See here for more information.
https://www.jobrunr.io/en/guides/migration/v8/#breaking-changes

Additionally, it specifies the mariadb hibernate dialect to be used to fix the No longer supported default specification for pessmistic lookup of jobrunr jobs as shown below: 
`WARN o.h.d.Dialect: HHH000511: The 5.5.5 version for [org.hibernate.dialect.MySQLDialect] is no longer supported..`

If the dialect is not specified some jobs can fail with the following exception:

```
org.springframework.dao.InvalidDataAccessResourceUsageException: JDBC exception executing SQL [You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'of p1_0' at line 1] 
Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'of p1_0' at line 1
```

### Modified Areas

- [x] Application layer (`application/` services, use cases)
- [x] Infrastructure layer (JPA repositories, external integrations, config)

### Behavioral Changes

- [x] No behavioral changes (internal refactoring, optimization, test addition, etc.)
- [ ] Yes, behavior changes: 

## Sensitive Changes Requiring Review

- [ ] **Database schema changes** (`sql/complete-schema.sql`, table/column DDL)
- [ ] **Spring Security configuration** (`security/`, ACL setup, authentication/authorization)
- [ ] **FAIR / RO-Crate export format** (`docs/fair/`, RO-Crate builder logic)
- [ ] **Artemis messaging topics** (JMS topic names, consumer configuration)
- [ ] **Requirement file edits** (`docs/requirements.md`)
- [x] None of the above

### DIFF
--- a/datamanager-app/src/main/java/life/qbic/datamanager/configuration/JobRunrDatasourceConfig.java
+++ b/datamanager-app/src/main/java/life/qbic/datamanager/configuration/JobRunrDatasourceConfig.java
@@ -16,7 +16,7 @@ import org.springframework.context.annotation.Configuration;
 public class JobRunrDatasourceConfig {
 
   @Bean(name = "jobRunrDatasourceProperties")
-  @ConfigurationProperties("org.jobrunr.database.datasource")
+  @ConfigurationProperties("jobrunr.database.datasource")
   public DataSourceProperties dataSourceProperties() {
     return new DataSourceProperties();
   }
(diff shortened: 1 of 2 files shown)


sven1103 (approved):
Great!

PR #1522, opened 2026-09-03
Persist Drawer State within Application

Steffengreiner:
## Description

This PR introduces a variable that persists the drawer state within the application to ensure an opened drawer stays open while a closed drawer stays closed while navigating the application. 
### Modified Areas
- [x] UI / Views (`views/` components, Vaadin routes)

### Behavioral Changes
- [x] Yes, behavior changes: Drawer state persists while navigating the application


datamanager-app/src/main/java/life/qbic/datamanager/views/DataManagerLayout.java
```diff
@@ -51,4 +55,21 @@ public void showRouterLayoutContent(HasElement content) {
...
+   * {@link life.qbic.datamanager.views.projects.project.experiments.ExperimentMainLayout}
+   */
+  public void persistDrawerStateBetweenLayouts() {
+    Boolean wasOpen = (Boolean) VaadinSession.getCurrent()
+        .getAttribute(DRAWER_STATE_KEY);
+    if (wasOpen != null) {
+      setDrawerOpened(wasOpen);
+    }
+    setDrawerOpened(wasOpen == null || wasOpen); // default true/open
+    getElement().addPropertyChangeListener(DRAWER_STATE_KEY, event -> VaadinSession.getCurrent().setAttribute(DRAWER_STATE_KEY, isDrawerOpened()));
```
KochTobi, line 70: How is the property change listener triggered?

datamanager-app/src/main/java/life/qbic/datamanager/views/DataManagerLayout.java
```diff
@@ -51,4 +55,21 @@ public void showRouterLayoutContent(HasElement content) {
...
+   * {@link life.qbic.datamanager.views.projects.project.ProjectMainLayout},
+   * {@link life.qbic.datamanager.views.projects.project.experiments.ExperimentMainLayout}
+   */
+  public void persistDrawerStateBetweenLayouts() {
+    Boolean wasOpen = (Boolean) VaadinSession.getCurrent()
+        .getAttribute(DRAWER_STATE_KEY);
+    if (wasOpen != null) {
+      setDrawerOpened(wasOpen);
+    }
+    setDrawerOpened(wasOpen == null || wasOpen); // default true/open
```
KochTobi: You can get rid of the if statement. Your second option `    setDrawerOpened(wasOpen == null || wasOpen); // default true/open` already does that for you.

Steffengreiner: Dang you're absolutely right nice catch 👀

datamanager-app/src/main/java/life/qbic/datamanager/views/DataManagerLayout.java
Steffengreiner: Heya @KochTobi, good question!
The property change listener is triggered if the state of the element with the property "drawerOpened" is changed. 
This is handled within the underlying applayout (See method isDrawerOpened()) which links this property to the drawer itself. 
Basically if the drawer state changes from open to close and vice versa the state of the property is changed and the event is thrown to which we can listen here.

KochTobi: As far as I can tell, you never set the property of the element to a specific value. Only the java class has a property with the name. Is this mapped by vaadin? Do you still need to call getElement().setProperty(xyy) or similar?

Steffengreiner: Heya @KochTobi, 

You're right this property is set by vaadin and this PR just uses the already established nomenclature.
See the screenshot below from the default applayout class:

Screenshot 2026-09-04 at 12 35 53

KochTobi (approved):
LGTM

KochTobi (changes requested):
Did you adapt for the `@Synchronize` annotation?

KochTobi (approved):

PR #1524, opened 2026-09-07
chore: make log retention configurable via application properties

sven1103-agent:
## Summary

Configures the log file rolling policy to use a configurable max retention period instead of a hardcoded value.

## Changes

- **application.properties**: Added `logging.file.max-history=60` (default: 60 days)
- **logback-spring.xml**: Reads the property via `<springProperty>` and passes it to the `SizeAndTimeBasedRollingPolicy`

## How to configure

Edit `logging.file.max-history` in `application.properties` to change the number of days archived logs are retained.

## Checklist

- [x] No hardcoded values — retention is fully configurable
- [x] Sensible default of 60 days


KochTobi (approved):
LGTM

Issue #1467, opened 2026-07-10
[Story] FEAT-DATSET-01: Connecting open, published datasets


sven1103-agent:
### Parent Feature

#1466

### Requirement IDs

`DATA-R-01`, `COMM-R-01`

### User Story

> As a researcher, I want to connect an open, published dataset on InvenioRDM with my project, so I can retain an overview of associated data for myself and collaborators on the project.

### Acceptance Criteria

- Given a user has write access to a project, When they navigate to the project's dataset section, Then they can access a **Search Dataset** section for available InvenioRDM instances.
- Given the user is in the **Search Dataset** section, When they are presented with InvenioRDM instances, Then they can select the instance they want to search in (e.g. Zenodo, FDAT).
- Given no search term is provided, When the user opens the search, Then a paginated result of the search is shown to the user.
- Given the user enters a search term, When they submit it, Then the system presents them with the matching search results from the selected InvenioRDM instance.
- Given a search result is displayed, When the user clicks **Connect**, Then the selected dataset is successfully added to the project in the system.
- Given the connection fails, When the operation cannot complete, Then the system informs the user about the failure.
- Given the connection is successful, When the operation completes, Then the system informs the user about success.
- Given a dataset is connected, When another user with project access views the project, Then that user is informed about the connected resource.
- Given a user is connecting a dataset, When they choose to narrow the connection, Then they can optionally set the experiment to associate the dataset with a specific experiment within the project.

### Notes & Context

- Part of **FEAT-DATASET-CONNECTION** (Connect associated InvenioRDM datasets with Data Manager projects)
- Stable Story ID: `FEAT-DATSET-01`
- Persona: Researcher (primary)
- Source document: 2026-06-19_FEATURE_Connect_associated_datasets_with_DM_projects.docx (stakeholder space — see parent feature #1466 §Stakeholder Artifacts) — Story 01

Issue #1466, opened 2026-07-10
[Feature] Connect associated InvenioRDM datasets with Data Manager projects


sven1103-agent:
### Feature ID

`FEAT-DATASET-CONNECTION`

### Requirement IDs

`DATA-R-01`, `DATA-R-02`, `DATA-R-03`

> **Note:** These requirement IDs are proposed pending confirmation and formal additions to `docs/requirements.md`. See **Notes & Context** for the full proposal.

---

### Description

Researchers can connect external datasets hosted on public InvenioRDM instances (e.g. Zenodo, FDAT) to their Data Manager projects, manage the connections, and synchronise connection state with the source platform. Both open published datasets and access-restricted published datasets (requiring user-provided tokens) are supported.

Data Manager serves as the **hub**: a central access point for project data and its associated published datasets. InvenioRDM remains the **publishing platform**: the canonical home of the dataset, its DOI, and its publication metadata.

The integration uses a **conventional API-based approach** against the InvenioRDM OpenAPI to search records, resolve metadata, and manage connections.

---

### Scope / Boundaries

**In scope:**
- Connecting open published datasets from InvenioRDM instances to a project
- Connecting access-restricted published datasets (requires user token)
- Viewing connected dataset metadata and access links
- Removing dataset connections to maintain data integrity
- Synchronising connected datasets with latest metadata from the source instance
- User-level credential management for InvenioRDM instances (add/remove validated access tokens)
- Searching connected datasets within a project
- Sorting connected datasets by publication date
- Notifying project members of connection changes
- Target platforms (initial): Zenodo, FDAT (University of Tübingen)

**Out of scope:**
- Creating new datasets on InvenioRDM from within Data Manager (future feature)
- Changes to Data Manager's internal data handling (OpenBIS, measurement management)
- Cross-project dataset connection features
- Full dataset content download/preview in Data Manager (access happens via link to InvenioRDM)

---

### Notes & Context

- **Source documents:** held in stakeholder space (Word Online / SharePoint); not committed to this repository. See the **Stakeholder Artifacts** table below for the list and description.
- **Connected Resource Information:** held in stakeholder space; defines the dataset metadata fields displayed when viewing a connected dataset (see **Stakeholder Artifacts** below).
- **InvenioRDM OpenAPI:** https://inveniosoftware.github.io/invenio-openapi/
[...]

PR #1528, opened 2026-09-07
fix: reset dataset selection on repository change and fix sidebar footer layout on small screens

sven1103-agent:
## Description

Fixes two issues in the **Connect Dataset** sidebar (right-side panel for connecting InvenioRDM datasets to a project):

### 1. Dataset selection not reset when switching the repository

When users selected datasets and then changed the repository in the instance dropdown, the selection was kept — the selected cards belonged to the *previous* provider but stayed selected against the new provider's search results, which is confusing and could lead to connecting the wrong datasets.

**Fix:** whenever the selected repository changes, the sidebar now:
- deselects all grid rows (which also resets the "N selected" label and disables the "Connect Selected" button via the existing selection listener),
- clears the cached search results of the previous provider and refreshes the grid data provider, so no stale hits from another provider can ever be rendered.

This guarantees selections can only ever be made for the currently selected provider.

### 2. Sidebar footer displaced next to the results grid on small screens

On small/short viewports the footer (experiment picker + connect button) was rendered *to the right of* the results grid instead of pinned at the bottom of the panel.

**Root cause:** the project-wide `.flex-vertical` utility class sets `flex-wrap: wrap`. In a column flex container with a definite height (the panel is `height: 100%`) that is too short for all children, CSS flex line-breaking ignores `flex-shrink` and wraps the overflowing footer into a second flex line laid out along the cross axis — i.e. next to the grid.

Additionally, below 460px viewport width the panel's `min-width: 460px` clamped `max-width`, making the panel overflow the left edge of the screen.

**Fix (scoped to the sidebar, no other views affected):**
- `flex-wrap: nowrap` on the sidebar's vertical chains (`.cds-body`, `.cds-content`, `.cds-results`, `.cds-footer`); overflow is handled by the scrollable content area as intended, keeping the footer pinned at the bottom.
- `@media (max-width: 460px)` override so the panel spans the full viewport width instead of overflowing off-screen.

## Related issues

- Feature: #1466 — Connect associated InvenioRDM datasets with Data Manager projects
- Story: #1467 (FEAT-DATSET-01) — Connecting open, published datasets

## Requirement IDs

- `DATA-R-01`
- `COMM-R-01`

## Testing

- Select datasets in the sidebar, switch the repository in the dropdown → selection is cleared, connect button disabled.
- Resize the browser to a short/narrow viewport and open the sidebar → footer stays pinned at the bottom, panel does not overflow the screen.

### DIFF
--- a/datamanager-app/frontend/themes/datamanager/components/connect-dataset-sidebar.css
+++ b/datamanager-app/frontend/themes/datamanager/components/connect-dataset-sidebar.css
@@ -194,6 +194,37 @@
   border-top: 1px solid var(--lumo-contrast-10pct);
 }
 
+/* ── Responsive hardening ──────────────────────────────────────────────── */
+
+/*
+ * The project-wide .flex-vertical utility class sets flex-wrap: wrap. In a
+ * column flex container with a definite height that is too short for all
+ * children (small / short viewports), flex line breaking ignores flex-shrink
+ * and wraps the overflowing items into a SECOND column laid out along the
+ * cross axis — i.e. the footer gets rendered to the RIGHT of the results
+ * grid. Forcing nowrap on every vertical flex chain of the sidebar keeps the
+ * header / scrollable-content / footer stack intact; overflow is handled by
+ * the scrollable content area (.cds-content) instead of wrapping.
+ */
+.connect-dataset-sidebar .cds-body,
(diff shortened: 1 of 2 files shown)


Steffengreiner (approved):
Thanks for adressing my points so quickly well done 👍

PR #1533, opened 2026-09-10
refactor: redesign profile settings page with card layout and inline editing

sven1103-agent:
## Summary

Final piece of the settings page refactor: the **Profile** settings page now matches the card-based `SettingsSection` style of the already-refactored *API Tokens* and *External Providers* pages.

## Changes

**Layout**
- `UserProfileMain` now extends `Main` and owns the `SettingsSection("Profile", …)` blueprint (consistent with `PersonalAccessTokenMain` / `ExternalProvidersMain`)
- `UserProfileComponent` renders two cards: **Profile Information** (avatar + full name, username, email) and **Linked Accounts** (reuses `AccountContentBox`)

**New reusable component: `InlineEditableField`** (`views/general/`)
- Focus-to-edit pattern: field is always visible and readable as an input; on focus it becomes editable with save/cancel controls
- Native Vaadin `setAutoselect(true)` for select-all-on-focus — no JS in Java
- Read-only display state: clean underline instead of Lumo's default dashed outline; content-sized width (`ch`) so long values like email addresses stay fully visible, with `title` tooltip as safety net
- `setEditable(false)` for fields that require access rights (email stays display-only)
- Validation errors via native TextField invalid state; control buttons stay centered regardless of error text
- Enter saves, Escape cancels, blur reverts to display mode; page reload on successful save refreshes the account overview header

**Cleanup**
- Removed orphaned legacy profile styles (`page-area.css`, `dialog.css`, `grid-templates.css`)

**Bug fix**
- Navigation to another settings section and back no longer crashes (whole section is removed on re-enter, not just the nested component)

## Requirements

Implements the Profile section of the settings hub refactor (progressive UI modernization; no new system capability beyond what the profile page already offered).


Steffengreiner (approved):
Works like a charm well done. 👍

PR #1545, opened 2026-09-14
feat: pinned-project quick access on the project overview (FEAT-PINNED-01)

sven1103-agent:
## Summary

Implements **pinned-project quick access**: a user can pin up to six of the projects they have access to, and those pins render as a personal shortlist at the top of the project overview — above the title and the search controls — so that returning to the projects one is actually working on no longer requires a search each time.

Closes the implementation side of `USER-R-04`, which was merged on its own in #1544 as required by `AGENTS.md` §0.

- **Requirement:** `USER-R-04` — Pinned-Project Quick Access
- **Feature:** `FEAT-PINNED-PROJECTS`
- **Story:** `FEAT-PINNED-01` — Pin Projects for Quick Access on the Project Overview
- **ADR:** [ADR-0008 — Pinned projects are user preferences, not project state](docs/adr/0008-pinned-projects-as-user-preferences.md)
- **Tasks:** [docs/plans/FEAT-PINNED-01-implementation-plan.md](docs/plans/FEAT-PINNED-01-implementation-plan.md)

## ⚠️ Human approval requested for two things

1. **Database schema change** — a new table `pinned_projects` (`AGENTS.md` §12). See §Schema below; `DATAMANAGEMENT_DB_DDL_AUTO` stays `none`, so nothing is created automatically.
2. **New ADR** — ADR-0008 plus its row in `docs/adr/README.md` (`AGENTS.md` §12).

## Behaviour

| Situation | What the user sees |
|---|---|
| Never pinned anything | The overview is byte-for-byte what it was before this PR — the row is not rendered at all |
| Pins exist | Up to six compact cards, newest pin first, above the title and search field; each opens the project. The row adapts: 1–3 cards in one row, 4 wraps to 2×2, 5 to 3+2, 6 to 3×2; a single card is capped at half the row width |
| Pin / unpin on a project card | Every card has a `⋮` (kebab) menu in the top-right. The menu offers **Pin project** or **Unpin project** depending on state. Pinned cards additionally show a small inline pin indicator left of the kebab. The kebab is a GitHub-style action button (transparent, padded hit area, background-only hover) and is a sibling of the card-body `RouterLink`, so clicking it cannot also navigate |
| Unpin from the pinned row | Each pinned card has the same `⋮` menu with **Unpin project** |
| Seventh pin attempt | Nothing is stored; toast: *"You can pin up to 6 projects. Unpin one to pin another."* No existing pin is ever evicted |
| Pinned project became unreadable | The card stays, marked **No access**, showing only the project code and title as they were when it was pinned, and it stays removable — otherwise it would hold one of the six places forever |
| Search / sort / page change | The row is unaffected and does not enter the list state in the URL |
| Pinned project also on the current result page | Shown twice, on purpose (see trade-off below) |
| Project code display | Codes render as a compact monospace badge (`ProjectCodeBadge`) distinct from the title, reused across overview cards and the pinned row |

## Design decisions

Two are architectural and recorded in **ADR-0008**; the rest are story-level.

- **A pin is an application-layer user preference, not `Project` aggregate state.** Storing it on the project would bump `Project#lastModified` on every pin — and `lastModified` is the default sort key of the very list this feature exists to speed up. It also leaks a private preference into shared project state.
- **The pin row carries a write-once snapshot of the project code and title**, used *only* when the live project is unreadable. Accessible pins always render live data from `project_overview`. The snapshot is never refreshed on read, so a page load stays a read.
- **Unpinning is keyed on (`userId`, `projectId`) and performs no project access check.** This is a deliberate, narrow exception to "all project reads go through the ACL-resolved accessible-id list"; without it a user could be locked at the cap by pins they cannot see or remove. It is documented in ADR-0008 and in the `PinnedProjectStore` javadoc.
- **The cap is counted over stored rows, not readable rows**, for the same reason.
- **A pinned project is not suppressed in the list below.** Suppressing it would change `countOverview` totals and page boundaries established in ADR-0007; a duplicated card is the cheaper cost.
- **The row never enters `ListState`/the URL**, so `ADR-0007` and `USER-R-03` semantics are untouched.
- **The service owns the ordering** ("most recently pinned first", enforced again by a stable `Instant`-precision sort after the store read, so the contract holds even if a deployed column stores coarser precision). The UI renders the returned order as-is — user-controlled reordering, if it ever ships, will be stored and served by the service rather than computed in the view.
- Requirement text and ADR were kept free of UI details on purpose (no "6", no cards, no placement) — those live in the story acceptance criteria.

## Schema

```sql
CREATE TABLE IF NOT EXISTS `pinned_projects` (
  `userId`               varchar(255) NOT NULL,   -- bare ref, no FK: follows personal_access_tokens
  `projectId`            varchar(255) NOT NULL,    -- FK → projects_datamanager ON DELETE CASCADE
  `pinnedAt`             datetime(6)  NOT NULL,
  `projectCodeSnapshot`  varchar(255)  DEFAULT NULL,
  `projectTitleSnapshot` varchar(255)  DEFAULT NULL,
  PRIMARY KEY (`userId`, `projectId`),
  KEY `idx_pinned_projects_user_pinned_at` (`userId`, `pinnedAt`),
  CONSTRAINT `fk_pinned_projects_project` FOREIGN KEY (`projectId`)
    REFERENCES `projects_datamanager` (`projectId`) ON DELETE CASCADE
);
```

Additive only — no existing table, column or view is altered. Idempotent migration
[`sql/migrations/create-pinned-projects.sql`](sql/migrations/create-pinned-projects.sql), operator entry
in [`docs/migrations/NEXT.md`](docs/migrations/NEXT.md), canonical DDL in `sql/complete-schema.sql`.
**Deployment order matters:** apply the migration before rolling out this build, otherwise the project
overview fails on the missing table. Rollback is `DROP TABLE pinned_projects` and loses only user
shortlists.

## Code layout

| Module | What was added |
|---|---|
| `project-management` | `application/PinnedProjectService` (cap, access resolution, label snapshot, newest-first ordering, `PinOutcome`), `application/pinned/{PinnedProject, PinnedProjectId, PinnedProjectView}`, port `application/api/PinnedProjectStore` |
| `project-management-infrastructure` | `PinnedProjectRepository` (explicit JPQL — composite embedded id makes derived queries brittle) and `PinnedProjectStoreImplementation` |
| `datamanager-app` | `PinnedProjectsComponent` (shortlist row), kebab-menu pin/unpin on the overview cards, `ProjectCodeBadge`, `page-area.css` (border-based cards, per-count grid, hover states, 1400px content cap on the overview), one toast message key |
| touched existing code | `ProjectInformationService` gained `findAccessibleProjectIds()` (public wrapper around the existing private ACL sid+authority resolution) so pin visibility and list visibility cannot drift apart; `ProjectCollectionComponent` constructor takes two more beans |

Expected failure paths are returned as `PinOutcome` values, never thrown, per `ExceptionHandling.md`.

## Tests

```
./run-tests.sh  # byte-buddy agent, see script header
./mvnw -Pdevelopment -pl project-management,project-management-infrastructure,datamanager-app -am test
→ BUILD SUCCESS
```

New specs: `PinnedProjectServiceSpec` (11) — live vs placeholder rendering, stored order, defensive newest-first re-sort, cap rejection,
pinning without access, pin idempotency, unpin without access, unpin of a non-existent pin, unauthenticated;
`PinnedProjectStoreImplementationSpec` (10) — duplicate guard, blank-user-id guards, delete reporting;
`PinnedProjectsComponentSpec` (5) — hidden empty row, one card per pin, placeholder has no `RouterLink`
but keeps its unpin control, re-render from supplier.

Not covered by unit tests and still to be checked by hand: the responsive reflow of the row, and the
revoked-pin path against a real database (needs a project whose ACL was revoked).

## Governance notes

- **Traceability deviation, deliberate:** this Feature and Story exist only in `docs/features.md`, with
  `—` in the `**GitHub**` fields, and their tasks live in the plan file. `AGENTS.md` §0/§11 were **not**
  amended; if docs-only tracking should become the norm for low-risk UI features, that deserves its own
  follow-up.
- `docs/requirements.md` is **not** touched in this PR (that was #1544).
- No new Maven dependency, no module added, no Spring Security or Vaadin version change, no Artemis or
  RO-Crate change. `personal_access_tokens`-style bare `userId` avoids a new cross-context dependency.
- The snapshot columns hold a project title outside `projects_datamanager`. They are read only for the
  owner's own placeholder rendering and must never be exported, indexed or used in RO-Crate output —
  stated in the entity javadoc, the migration header and ADR-0008.

## Commit structure

1. `docs: add FEAT-PINNED-PROJECTS feature, FEAT-PINNED-01 story and ADR-0008`
2. `feat(db): add pinned_projects table for user quick access`
3. `feat: pinned-project quick access on the project overview (FEAT-PINNED-01)`
4. `feat: refine pinned-project quick access UX (FEAT-PINNED-01)`

Deferred, tracked in the plan file §5: pinning from the project detail header, a global pinned-project
switcher in the navigation (reuses `PinnedProjectService` unchanged), a configurable cap, and a
"recently visited" derivation.


Steffengreiner (dismissed):
Well done 👍

PR #1542, opened 2026-09-14
feat: paginated project overview with responsive card layout (FEAT-PAG-LIST-01)

sven1103-agent:
## Summary

Implements **FEAT-PAG-LIST-01 — Paginated Project Overview with Responsive Card Layout** (#1537), the first story of the [FEAT-PAGINATED-LISTS](https://github.com/qbicsoftware/data-manager-app/issues/1536) feature.

Replaces the scroll-loaded (endless-scrolling) Vaadin Grid card container on the project overview with an explicitly paginated list:

- **Paginated card list (USER-R-01):** a reusable `PaginationBar` (prev/next, numbered window with ellipsis, "Page X of Y — N projects" location+total label, page-size selector 12/24/48/96) plus a compact top pager. Only the current page is rendered.
- **List state in the URL (USER-R-03):** `page`, `size`, `q`, `sort` query parameters are the single source of truth. Interactions write the URL via `pushState` (page/page-size/sort) and `replaceState` (search); `BeforeEnter` seeds state on load/reload/shared links; a history-state handler restores prior states on back/forward without re-initialising the view. Search and sort reset paging to page 1.
- **Responsive, native scrolling (USER-NFR-01):** cards render as one full-width card per row (preserving the previous list look) inside a plain DOM container — no embedded scroll container. The `grid-templates.css` height-floor stopgap for the overview is removed so the page scrolls natively.
- **Backend:** adds the missing count query for the pager total (`ProjectOverviewLookup.count`, `ProjectInformationService.countOverview`), reusing the existing filter + access-rights specification.
- **Ordering:** default remains `lastModified` descending (per stakeholder decision on #1537, until per-user list preferences exist); `projectCode asc` is applied only as an internal tie-break key for deterministic pagination.
- **ADR:** documents the architecture in [ADR-0007](docs/adr/0007-paginated-lists-and-list-state.md) (offset/limit + count, History-API URL state, custom pager, plain-DOM card container) — created via this PR per the ADR workflow.

## Requirement IDs

- `USER-R-01` — Paginated List Display
- `USER-R-03` — List View State in the URL
- `USER-NFR-01` — Responsive List Rendering

No new requirements or requirement changes are introduced.

## Tests

- 22 new/extended Spock specs for the pure list-state/pager logic (`ListStateCodecSpec`, `PageRangeSpec`, `PaginationBarSpec`, `ProjectOverviewSortOptionSpec`) — pass locally.
[...]

PR #1546, opened 2026-09-14
fix: project overview swallows all navigation out of the view (ADR-0007 regression)

sven1103-agent:
## Summary

Navigating away from the project overview does nothing: the browser URL changes, the view never
switches. Every `RouterLink` on the overview is affected — project cards, the connected-dataset
footer, and the pinned-project cards added in #1545.

**Regression introduced by #1542** (`feat: paginated project overview`, ADR-0007 / `USER-R-03`). It
is not caused by #1545; that PR only made it visible.

## Root cause

`ProjectOverviewMain.beforeEnter` registered a history state change handler and never removed it:

```java
UI.getCurrent().getPage().getHistory().setHistoryStateChangeHandler(this::onHistoryStateChange);
```

`Page.getHistory().setHistoryStateChangeHandler(...)` is **not** an additive subscription. Vaadin's
own javadoc:

> History state changes are triggered when a `popstate` event is fired in the browser **or when the
> user has navigated using a router link**. **There can only be one handler at a time.**

and `com.vaadin.flow.component.UI` uses that single slot for routing itself — on the first
navigation it installs

```java
getPage().setHistoryStateChangeHandler(e -> renderViewForRoute(e.getLocation(), e.getTrigger()));
```

`NavigationRpcHandler` then delivers *every* client navigation — link clicks included — to whichever
handler is installed. So the view replaced the router's handler, and its

```java
if (!ProjectRoutes.PROJECTS.equals(event.getLocation().getPath())) { return; }
```

discarded everything that was not the list route. The handler is page-scoped, so it also stayed
active after the view was gone.

## Fix

Same file, three parts:

1. `beforeEnter` **captures** the installed handler before replacing it, and installs a stable field
   reference (`listStateHistoryHandler`) instead of a new method reference per call, so re-entry can
   be detected instead of chaining the view onto itself.
2. `onHistoryStateChange` **delegates** non-list events to the captured handler, so the router performs
   the navigation the user asked for.
3. `beforeLeave` (new `BeforeLeaveObserver`) **restores** the captured handler, so the view never
   leaves a page-scoped handler behind.

`USER-R-03` behaviour is unchanged: entering `/projects` with `?page=&size=&filter=&sort=` and
browser back/forward still re-apply the list state; `ADR-0007` semantics hold.

## Verification

- **Manual (required, not yet performed by me):** from the overview, click a project card → the project
  opens; click the dataset footer → datasets open; then Back/Forward → page, page size, filter and
  sort are restored; then navigate to the overview from the side nav and leave again, to confirm the
  restore path.
- `./mvnw -Pdevelopment -pl datamanager-app -am compile` → success.
- No automated test. Asserting this needs a `UI` with a real `History` and Vaadin's own first-
  navigation handler installation, which unit tests cannot reproduce; the only cheap route is Karibu,
  i.e. a new third-party dependency (BOM entry + `sigcheck` PGP handling), which I did not want to
  bundle into a navigation fix. Happy to open that as a follow-up if the team wants it covered.

## ⚠️ This will be re-introduced three more times

`FEAT-PAG-LIST-02`, `-03` and `-04` port the same list-state mechanism to the sample, measurement and
raw-dataset lists. Any of them copying the original `setHistoryStateChangeHandler(this::x)` line
reproduces this defect on that view. The chain-and-restore rule belongs in `ADR-0007`, but ADRs are
append-only, so I did not edit it. Two options, both outside this PR:

- append a **correction/links note** to `ADR-0007` (a one-time exception to append-only), or
- add a short **ADR-0009**: "a view that claims the page's history state change handler must chain and
  restore it".

I'd take the second. Tell me which and I'll add it; I did not want to decide the ADR policy in a bug
fix.

## Governance

- Requirement touched: `USER-R-03` (list view state in the URL) — behaviour restored to what the
  requirement and `ADR-0007` specify, no requirement text change needed.
- No GitHub issue is referenced: none was created for this (deliberate; say the word and I'll file one).
- No schema, security, dependency or messaging change. One file, 42 insertions, 4 deletions.


sven1103 (approved):
It restores the navigation to project route targets while preserving the URL based search configuration

PR #1553, opened 2026-09-15
[Docs] Add sample batch removal requirements

### LINKED ISSUE #1330: Incorporate Batch definition during Sample Sheet upload/edit template
### What are you missing?

In the current upload process, the user has to do their sample updates and registration batchwise via the distinct batch component.

### Why do you need a solution?

**Multiple Reasons**

1.) Its needlessly cumbersome for other facilities that want to update multiple batches at the same time.
2.) For Facilities working with a singular batch, the distinction and focus on multiple batches is confusing
3.) Batch component takes up a lot of space in the sample grid making the visualization of samples within the view less ideal.
4.) By removing the batchwise focus during sample editing/registration, it would allow the dev team to reuse the generic filtergrid currently being defined within the measurement context. 

### Ideas

1.) Remove the batch component and move the batch definition as a (ideally mandatory) distinct column within the sample templates for:

A) Registration
B) Editing  

UI Prototype can be found [here](https://www.figma.com/design/G0wNwsfyuYMwlSfqLjvsod/Data-Manager-App-Design?node-id=1776-48182)

Please also ensure that for each sample the Modification and Registration Date is shown within the sample grid UI.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### LINKED ISSUE #1548: [Feature] Remove explicit sample batch from Data Manager
### Feature ID

`FEAT-SAMPLE-BATCH-REMOVAL`

### Requirement IDs

`SAMPLE-R-01`, `SAMPLE-R-02`, `SAMPLE-R-03`, `SAMPLE-R-04`

> These requirements were added to `docs/requirements.md` in PR #1553.

---

### Description

Remove the explicit **sample batch** (`sample_batches`) concept from Data Manager. Samples are registered directly in the project (still grouped under an experiment), and each sample carries a free-text `batch` label for grouping and backwards compatibility with existing data and downstream tooling that references the batch name.

Data Manager serves as the **hub** for project sample registration. The batch is no longer a first-class entity: there is no batch aggregate, no `sample_batches` table, and no batch-scoped grid or dialog. Instead, samples belong directly to a project and an experiment, and carry a mandatory `batch` label string.

---

### Scope / Boundaries

**In scope:**
- Registering samples directly in a project (grouped under an experiment) via an Excel spreadsheet, with the batch name as a mandatory column
- Adding a denormalised `project_id` association on the sample so project-wide sample queries avoid a transitive lookup through the experiment
- A mandatory free-text `batch` property on samples
- Moving batch creation/modification timestamps onto the sample (system-set, read-only)
- Removing the explicit batch entity, `sample_batches` table and `sample_batches_sampleid` join table
- Removing the batch-centric UI (batch grid, register/edit batch dialogs) in favour of a samples-only view with Excel-based edit and multi-select delete (mirroring the measurement workflow)
- Backwards-compatible data migration (stop-the-world: all nodes shut down at once, migrate, restart)

**Out of scope:**
- Changing how samples are registered into an experiment (registration still happens within an experiment)
- OpenBIS synchronisation behaviour
- Measurement assignment (covered by FEAT-MEASUREMENT-UPLOAD)
- Removing the batch label from external/API consumers that rely on it for backwards compatibility

---

### Notes & Context

- **Deployment model — stop-the-world migration:** The Data Manager runs on 3 nodes sharing a single database. It is acceptable to shut down all 3 nodes at once, apply the migration, and bring them back up together. No rolling transition window is required, so the migration is a single atomic step (backfill `sample_batches` → `batch` column and derive `project_id`, then drop the legacy tables). No dual-write to legacy tables is needed. See Story FEAT-SAMBAT-03.
- **Requirements** (added to `docs/requirements.md` in PR #1553):
  - `SAMPLE-R-01` — Direct sample registration within an experiment (batch mandatory)
  - `SAMPLE-R-02` — Sample–project and sample–experiment association (denormalised `project_id` on sample)
  - `SAMPLE-R-03` — Backwards-compatible `batch` property on samples and data preservation
  - `SAMPLE-R-04` — Removal of the explicit batch and samples-only management (Excel edit, multi-select delete)
- **Stakeholder request:** [Incorporate Batch definition during Sample Sheet upload/edit template #1330](https://github.com/qbicsoftware/data-manager-app/issues/1330).
- **Dependencies:** no other Feature blocks this; depends on the sample domain existing as-is today.

---

### Child Issues

| Story ID | Title | Status | GH Issue # |
|---|---|---|---|
| [FEAT-SAMBAT-01](https://github.com/qbicsoftware/data-manager-app/issues/1549) | Register samples directly in a project | 🟢 Done | #1549 |
| [FEAT-SAMBAT-02](https://github.com/qbicsoftware/data-manager-app/issues/1550) | Associate samples with project and experiment | 🟢 Done | #1550 |
| [FEAT-SAMBAT-03](https://github.com/qbicsoftware/data-manager-app/issues/1551) | Backwards-compatible batch property and data migration | 🟢 Done | #1551 |
| [FEAT-SAMBAT-04](https://github.com/qbicsoftware/data-manager-app/issues/1552) | Samples-only UI (remove batch grid and batch dialogs) | 🟢 Done | #1552 |

KochTobi-Agent:
Adds requirements for sample batch removal


docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+
+We want to remove the explicit batch as a first-class entity. Samples should be registered directly
+within an experiment of a project, and each sample should carry a free-text `batch` label purely for
+backwards compatibility with existing data and downstream tooling. This ADR pins the target data model;
+the strategy for migrating existing data is captured separately in [ADR-0009](0009-stop-the-world-migration-of-sample-batch-removal.md).
+
+## Decision Drivers
+
+* **SAMPLE-R-01:** samples are registered directly within an experiment; the batch is optional and not a
+  registration prerequisite.
```
Steffengreiner: The batch is not optional, A batch name should always be provided

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+* **SAMPLE-R-02:** every sample is directly associated with both its project and its experiment, so
+  project-wide sample queries avoid a transitive lookup through the experiment hierarchy.
+* **SAMPLE-R-03:** existing batch information is preserved — the batch name and the creation/modification
+  dates carry forward onto the sample. The pilot flag is not needed and is dropped.
```
Steffengreiner: Completely Right, However transferring the batch name to the sample could lead to migration issues if the batchname is identical. I'm unsure if thats an actual case or a constructed one. 🤔 

Image

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+* **SAMPLE-R-04:** the explicit batch is no longer exposed in the registration workflow; the UI presents
+  samples only, with batch as an optional property.
+* The batch currently provides no domain behaviour beyond grouping labels and a pilot marker (which is no
+  longer needed); removing the entity reduces schema and orchestration complexity.
+* Existing downstream tooling and historical data reference the batch name; a free-text `batch` property
+  preserves that without keeping the batch tables.
```
Steffengreiner: I'm unsure what is meant by downstream tooling and historical data here?

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Considered Options
+
+### Batch-to-sample mapping
+
+* [M1] Collapse the batch into a free-text `batch` property on the sample; move the batch creation and
+  modification dates onto the sample; drop the `Batch` aggregate, `sample_batches`, and
+  `sample_batches_sampleid`.
```
Steffengreiner: This is the intended outcome. The sample is the key entitiy our clients work with.

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+* [M2] Keep the `Batch` aggregate but relink it directly to project/experiment with real FKs; samples
+  keep referencing a batch by id.
+* [M3] Keep the `Batch` aggregate and tables unchanged; only stop exposing it in the UI.
+
```
Steffengreiner: No, if possible remove the outdated system 👍

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+### Sample–project association
+
+* [P1] Add a `project_id` column on the `sample` table with a real foreign key to `projects_datamanager`,
+  coexisting with the existing `experiment_id` (kept as a soft reference).
```
Steffengreiner: I agree this is the best solution since i'd expect down the line for an API call to have /project/samples since our clients are interested in getting the samples for their project instead of navigating down each experiment

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+* [P2] Add `project_id` without a foreign key constraint (matching the existing `experiment_id` /
+  `assigned_batch_id` soft-reference pattern).
+* [P3] Keep no project column on the sample; derive the project transitively via the experiment.
+
+## Decision Outcome
+
+Chosen: **M1 + P1.**
```
Steffengreiner: 👍

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
```diff
@@ -0,0 +1,148 @@
...
+
+* Good, because it matches the existing soft-reference columns.
+* Bad, because the association is not enforced and can drift, which is riskier for the primary
+  project-wide query path.
+
+### P3 — No project column
+
+* Good, because it changes the schema least.
+* Bad, because project-wide sample queries keep doing the transitive join through experiment, which is
+  exactly the inefficiency SAMPLE-R-02 exists to remove.
```
Steffengreiner: I think this section is nice for discussion but should not end up in the documentation

docs/adr/0009-stop-the-world-migration-of-sample-batch-removal.md
```diff
@@ -0,0 +1,100 @@
...
+* Bad, because it requires dual-write and dual-read logic in the application, a compatibility layer, and
+  a prolonged transition where the two representations can drift.
+* Bad, because it is significantly more complex than needed given the accepted stop-the-world window.
+
+### S3 — Keep legacy tables indefinitely
+
+* Good, because it avoids any destructive migration.
+* Bad, because it keeps redundant tables and dual-maintenance forever, and leaves the legacy batch
+  concept in the schema that SAMPLE-R-04 aims to remove.
+
```
Steffengreiner: As before, this is nice for making a decision but not necessary in the documentation itself no?

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
Steffengreiner: I think we can check the database if this issue is currently applicable and if so find a running number increment behind the batchname to keep the samples within the batches? (In this Case Test 2-1 and Test 2-2 etc.) Just a suggestion though

Steffengreiner (changes requested):
Heya @KochTobi-Agent and @KochTobi, Nicely Done! I have some suggestions and discussion points which should be addressed 👍

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
KochTobi-Agent: This just refers to the existing production data.

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
KochTobi-Agent: I think this is an important consideration for how the sample aggregate is designed. I added an architectural decision record. @sven1103 what do you think. Should we keep the ADR or get rid of it?

docs/adr/0009-stop-the-world-migration-of-sample-batch-removal.md
KochTobi-Agent: We need to document decisions made. This is why we introduced ADRs. @sven1103 would you categorize this decision as one we need to document?

docs/adr/0008-remove-explicit-sample-batch-from-data-model.md
KochTobi-Agent: Thank you @Steffengreiner for this excellent point. Although it is important to think of edge cases like this, I verified that no such case exists in the production data within an experiment.
```sql
SELECT s.experiment_id, sb.batchLabel,
       COUNT(DISTINCT sb.id) AS distinct_batches,
       COUNT(DISTINCT s.sample_id) AS sample_count
FROM sample_batches sb
JOIN sample_batches_sampleid sbs ON sbs.batch_id = sb.id
JOIN sample s ON s.sample_id = sbs.sample_id
GROUP BY s.experiment_id, sb.batchLabel
HAVING COUNT(DISTINCT sb.id) > 1
ORDER BY s.experiment_id, sb.batchLabel;
```

KochTobi-Agent: There are multiple batches with the same name on a global scale. But this is not an issue as far as I can tell. Names like "Batch 1" or "Batch1" or "first batch" occur multiple times. This rather indicates that no real consideration was taken in choosing the batch name or the batch name is used for batch ordering instead.

docs/adr/0009-stop-the-world-migration-of-sample-batch-removal.md
sven1103: no ADR needed, it is a PO decision, since this user story turned out to be not true and needs to be rewoked

KochTobi-Agent: Thank you for your feedback :)

KochTobi-Agent:
@Steffengreiner 
Thank you for your review. I addressed your points and am looking forward to reading your next review 😄

Steffengreiner (approved):
Well done! Thanks for checking the edge case and cleaning up the document 👍

